PR #4, opened 2022-09-26
Build textile to html

QuintinWillison:
We need the [Textile](https://textile-lang.com/) source files rendered for preview and consumption by client library developers.

For the time being I've made the decision not to convert _any_ of the textile files to a more modern and friendly format like [markdown](https://github.github.com/gfm/). This is mainly because it's useful to keep the rich Git commit history currently visible against the textile files ([as duplicated over from the `ably/docs` repository](https://github.com/ably/specification/blob/main/textile/notes-in-relation-to-move-from-ably-docs.md)).

I'm also trying to keep this build as simple and self-contained as possible, with minimal reliance on bigger frameworks or many other dependencies. Therefore, pulling this together with just a [basic Handlebars implementation](https://github.com/SmartBear/ruby-handlebars) fed with output from [Redcloth](https://redcloth.org/) feels appropriate and maintainable. Plus, [Tailwind CSS](https://tailwindcss.com/) for styling, which is most sensibly deployed natively using [Node.js](https://nodejs.org/) - so, using `npm` to run things, which makes sense (calling down to [Ruby](https://www.ruby-lang.org/) scripts when required).

Use the **View deployment** button on this pull request to see the rendered result.


QuintinWillison:
The failure of the 'Assemble' workflow is expected. I will run it again once we've release `v2` of the SDK Upload Action, which will include https://github.com/ably/sdk-upload-action/pull/46.

stmoreau (approved):
LGTM - nice one!

lmars (approved):

PR #88, opened 2022-10-03
Upgrade realtime protocol from version 1.2 to 2.0: 'no-connection-serial'

SimonWoolf:
includes:
- add channel reattaching and presence re-entering
- document connection recovery properly and explicitly
- remove RTN15f which was wrong (implied you should only republish if a resume was successful) and redundant to RTN19a

This is a port of https://github.com/ably/docs/pull/1526 , several discussions have already taken place on that PR.


SimonWoolf:
Two more changes, per discussion this morning and on slack:
- [Spec: Add RTP17h -- myMembers is keyed by clientId only](https://github.com/ably/specification/pull/88/commits/570aae74f861b934852217463e2c4e3c2c8e7f2a)
- [Spec: set channel.properties.channelSerial from ATTACHED as well](https://github.com/ably/specification/pull/88/commits/963f9a04d4dc8da3ace238746db2bec379b80008)

textile/features.textile
```diff
@@ -485,29 +483,37 @@ h3(#realtime-connection). Connection
...
+** @(RTN16j)@ When a library is instantiated with the @recover@ client option, for every channel/channelSerial pair in the @recoveryKey@, it should instantiate a corresponding channel and set its "RTL15b":#RTL15b @channelSerial@.
+** @(RTN16k)@ When the library first connects to Ably after being instantiated with a @recover@ client option, it should add an additional @recover@ querystring param to the websocket request, set from the @connectionKey@ component of the @recoveryKey@. Once the library has successfully connected to Ably, it should never again supply a @recover@ querystring param.
+** @(RTN16g)@ @Connection#getRecoveryKey@ is function that returns a string which incorporates the @connectionKey@, the current @msgSerial@, and a collection of pairs of channel @name@ and current @channelSerial@ for every currently attached channel. This must be serialized in a way which is able to encode any unicode channel name. The SDK may assume that the recovery key will only be consumed by the same type of SDK, so this spec does not specify any particular serialization; however, the format should be forward-compatible through the same major version of the SDK.
```
QuintinWillison, line 509: I don't think that `get` is a verb that we should be using as a prefix for any method names, especially when those methods are designed to do anything more complex than a 'simple property get'. This observation is made even more stark in this particular context, given it's replacing the previous field/property called `recoveryKey`.

My preference would be to rename this method to `createRecoveryKey`. I think this is more representative of the fact that a call to this method is, as I understand it, capturing a snapshot of state from this Realtime client instance... in that it's likely that a subsequent call to this method will return a different key, on the assumption it's being called while `CONNECTED`.

To further expand on why `get` is a difficult verb to use as a method name prefix for a method which is designed to be available across multiple platforms, in Java it's is idiomatic and common to expect that methods with this prefix are property getters - often returning the value of a `final`, `private` field (and Kotlin only goes on to further expand on that convention).

textile/features.textile
```diff
@@ -485,29 +483,37 @@ h3(#realtime-connection). Connection
...
+** @(RTN16h)@ @Connection#getRecoveryKey@ should return @Null@ when a the SDK is in the @CLOSED@, @CLOSING@, @FAILED@, or @SUSPENDED@ states, or when it does not have a @connectionKey@ (for example, it has not yet become connected). The @Connection#key@ and @Connection#id@ should also be @Null@ at these times.
```
QuintinWillison, line 510: `RTN16h` is a sibling of `RTN16g`, which is already defining the functionality of this method. I think this bullet should be `RTN16g1`, that is a child of `RTN16g`. That makes more sent for the way that we use spec points, and their hierarchy, in downstream conversations around compliance.

Related, I also think that `RTN16g` is too long, conflating many requirements together. Why not, also, break it into sub points? We have plenty of examples of where we have done that elsewhere in this spec..

QuintinWillison (changes requested):
A few observations, plus:

I would like to see this pull request built atop #105, if we can come to agreement that the approach I suggest there is agreeable to everyone, in particular @SimonWoolf who has already commented but not yet approved (at the time of me writing this, at least 😉).

textile/features.textile
SimonWoolf: :+1: refactored in https://github.com/ably/specification/pull/88/commits/cb9e6d33fa18a5b8494e761af73f1cd204ca1550

textile/features.textile
SimonWoolf: ok, https://github.com/ably/specification/pull/88/commits/ec268811dbf7054e318bf2c92f5e882247675195 changes to `create`

SimonWoolf: https://github.com/ably/ably-dotnet/pull/1185
https://github.com/ably/ably-ruby/pull/387/commits/19a7191a426e8071b95de3f7c40f9f3c1d591242
https://github.com/ably/ably-cocoa/pull/1496/commits/fa251382f46d9f96e34fd930bdf9192b8ff3e9e1
https://github.com/ably/ably-java/pull/842/commits/90dd27588b07f5f324b2e8ea8063f059b28c58f6
https://github.com/ably/ably-js/pull/1045/commits/9c0afb3e5078501c4e6e0c1ebcdbe1e6f0f78650

QuintinWillison (approved):
After some thought I have decided to [close my pull request, unmerged](https://github.com/ably/specification/pull/105#issuecomment-1305327659). It is not necessary to block this work on an as-yet-to-be-created change/release process, so I will strike out my [previous comment on this pull request](https://github.com/ably/specification/pull/88#pullrequestreview-1151950049). Thanks.

ikbalkaya (approved):
LGTM, just some minor typos

PR #89, opened 2022-10-03
Add mandatory version param for rest#request.

SimonWoolf:
Adds mandatory version param to rest#request for protocol version 2.0, per https://ably.atlassian.net/wiki/spaces/ENG/pages/2222456951/ADR61+Versioning+of+realtime+API+endpoints


textile/features.textile
```diff
@@ -128,8 +128,8 @@ h3(#restclient). RestClient
...
-** @(RSC19a)@ Method signature is @request(string method, string path, Dict<String, String> params?, JsonObject | JsonArray body?, Dict<String, String> headers?) -> HttpPaginatedResponse@ with arguments: @method@ is a valid "HTTP verb":https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html (must support @"GET"@, @"POST"@, and @"PUT"@, should support @"PATCH"@ and @"DELETE", may support others); @path@ is the path component of the URL such as @"/channels"@; @params@ and @headers@ are optional arguments containing pairs of key value strings (multi-valued headers are not supported) that will result in query params and HTTP headers being added respectively in the request (the argument types can be idiomatic for the language such as @Object@ in the case of JavaScript); @body@ is an optional @JsonObject@ or @JsonArray@ like object argument that can be easily serialized to MsgPack or JSON
-** @(RSC19b)@ All requests will unconditionally use the default authentication mechanism configured for the REST client i.e. basic or token authentication (see "Auth":#rest-auth)
```
QuintinWillison, line 132: If removing this spec point then we should replace it with a message similar to the one you've replaced `RSC19a` with.

textile/features.textile
```diff
@@ -128,8 +128,8 @@ h3(#restclient). RestClient
...
+** @(RSC19a)@ This clause has been removed as of the 2.0 spec and replaced with @RSC19f@.
+** @(RSC19f)@ Method signature is @request(string method, string path, string version, Dict<String, String> params?, JsonObject | JsonArray body?, Dict<String, String> headers?) -> HttpPaginatedResponse@ with arguments: @method@ is a valid "HTTP verb":https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html (must support @"GET"@, @"POST"@, and @"PUT"@, should support @"PATCH"@ and @"DELETE", may support others); @path@ is the path component of the URL such as @"/channels"@; @params@ and @headers@ are optional arguments containing pairs of key value strings (multi-valued headers are not supported) that will result in query params and HTTP headers being added respectively in the request (the argument types can be idiomatic for the language such as @Object@ in the case of JavaScript); @body@ is an optional @JsonObject@ or @JsonArray@ like object argument that can be easily serialized to MsgPack or JSON** @(RSC19b)@ All requests will unconditionally use the default authentication mechanism configured for the REST client i.e. basic or token authentication (see "Auth":#rest-auth)
```
QuintinWillison, line 132: I realise that some of this is inherited from `RSC19a`, however this is a new spec point so perhaps worth fixing now also... there's a mix of `string` versus `String` in the method signature. I think we should conform to one or the other, presumably `String` as that's what's in the IDL.

QuintinWillison (changes requested):

textile/features.textile
SimonWoolf: Did not intend to remove that spec item (and the [old docs pr](https://github.com/ably/docs/pull/1592/files) doesn't, so possibly a glitch when porting to the other repo), I'll fix.

textile/features.textile
SimonWoolf: done

QuintinWillison (approved):
Looks good to me, but I'll add a couple of other reviewers so we can perhaps get more than just my approval before we land this.

owenpearson (approved):

stmoreau (approved):

PR #92, opened 2022-10-03
Bump ruby version to latest v3 patch

SimonWoolf:
(3.0.2 for some reason is having trouble compiling for me, dunno why, 3.0.3 works fine. And is newer so eh why not)


QuintinWillison (commented):
Why not, indeed. 😁 

I'm slightly interested as to whether you would be able to get it working with the latest `3.0` Ruby `patch` which [appears to be `3.0.4` from April 2022](https://www.ruby-lang.org/en/news/2022/04/12/ruby-3-0-4-released/)... as that would perhaps be a more logical change to make.

SimonWoolf:
3.0.4 works fine, bumped to that

QuintinWillison (approved):

PR #97, opened 2022-10-07
Spec: add G4a, clarifying that protocol version should be treated literally

SimonWoolf:
Igor mentioned that ably-java defines has the protocol version as a float, which I'm not a fan of. This clarifies.


qsdigor:
Java SDK uses float internally which is then converted to String with one decimal point ("1.2" or "2.0"). I guess this is fine for now.

owenpearson (approved):

PR #105, opened 2022-10-20
More stringent versioning

QuintinWillison:
My motivation behind creating this pull request comes from an observation that came to me when reviewing #88.

I've been flying the flag for "don't mutate features spec points" for a while now, but had never written it down in so many (or is that few?) words.

In our version `1.2` spec, both `G4` and `RSC7a` had the spec version baked into them. This meant that these spec points needed to be either mutated each time the specification/protocol version was bumped (as @SimonWoolf had to do in the aforelinked PR), changing the semantic meaning of those spec points from an SDK's perspective.

The idea, going forwards, is that a spec point is all that is needed to define a particular feature or behaviour - staying constant from spec version to spec version. This means that spec points never need to carry any additional version information with them, as their meaning never changes.

This pull request aims to establish that footing more formally.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,9 +42,15 @@ by adding a local development HTTP server.
 When making changes to [the spec](textile/features.textile), please follow these guidelines:
 
 - **Ordering**: Spec items should generally appear in ID order, but priority should be placed on ordering them in a way that makes coherent sense, even if that results in them being numbered out-of-order. For example, if `XXX1`, `XXX2` and `XXX3` exist but it would make more sense for `XXX3` to follow `XXX1`, then just move the spec items accordingly without changing their IDs
-- **Addition**: When adding a new spec item, choose an ID that is greater than all others that exist in the given section, even if there is a gap in the currently assigned IDs. This is desirable so that client library references to spec items are still semantically valid even after they are removed from the spec rather than now having different semantics due to the ID being re-used. For example, if `XXX1a` and `XXX1c` exist but `XXX1b` doesn’t because it was removed in the past, then introduce `XXX1d` for the new spec item rather than re-using `XXX1b`
-- **Removal**: When removing a spec item, it must remain but replace all text with “This clause has been deleted.”. See [#1057](https://github.com/ably/docs/pull/1057) for an example of this in practice.
+- **Addition**: When adding a new spec item, choose an ID that is greater than all others that exist in the given section, even if there is a gap in the currently assigned IDs. 
+- **Modification**: Spec items should never be mutated, except to patch a mistake that doesn't change the semantics for SDK implementations. Follow the guidance outlined here in respect of _Replacement_ if the meaning or scope of a spec point needs to change.
+- **Removal**: When removing a spec item, it must remain but replace all text with `This clause has been deleted as of version Y.Z of this specification`. See [#1057](https://github.com/ably/docs/pull/1057) for an example of this in practice.
+- **Replacement**: When replacing a spec item, it must remain but replace all text with `This clause has been replaced by "X":#X as of version Y.Z of this specification`.
 - **Deprecation**: Our approach to deprecating features is yet to be fully evolved and documented, however we have a current standard in place whereby the text "(deprecated)" is inserted at the beginning of a specification point to declare that it will be removed in a future release. The likely outcome is that in the next major release of the spec/protocol we'll remove that spec item, per guidance above.
+
+Historically, before the above guidance was established - in particular around _Removal_ and _Replacement_ - there have been some cases where spec points were completely deleted.
+This left us open to the problem that client library references to spec items could end up semantically invalid if that spec point was re-used later.
+For example, if `XXX1a` and `XXX1c` exist but `XXX1b` doesn’t because it was removed in the past (prior to this guidance being established), then we should introduce `XXX1d` for the new spec item rather than re-using `XXX1b`.
 
 ## Release Process
 
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -54,8 +54,13 @@ h2(#test-guidelines). Test guidelines
 * @(G1)@ Every test should be executed using all supported protocols (i.e. JSON and "MessagePack":https://msgpack.org/ if supported).  This includes both sending & receiving data
 * @(G2)@ All tests by default are run against a special Ably sandbox environment.  This environment allows apps to be provisioned without any authentication that can then be used for client library testing. Bear in mind that all apps created in the sandbox environment are automatically deleted after 60 minutes and have low limits to prevent abuse. Apps are configured by sending a @POST@ request to @https://sandbox-rest.ably.io/apps@ with a JSON body that specifies the keys and their associated capabilities, channel namespace rules and any presence fixture data that is required; see "ably-common test-app-setup.json":https://github.com/ably/ably-common/blob/main/test-resources/test-app-setup.json. Presence fixture data is necessary for the REST library presence tests as there is no way to register presence on a channel in the REST library
 * @(G3)@ Testing statistics can be tricky due to timing issues and slow test suites as a result of sending requests to generate statistics.  As such, we provide a special stats endpoint in our sandbox environment that allows stats to be injected into our metrics system so that stats tests can make predictable assertions.  To create stats you must send an authenticated @POST@ request to the stats JSON to @https://sandbox-rest.ably.io/stats@ with the stats data you wish to create. See the "JavaScript stats fixture":https://github.com/ably/ably-js/blob/4e65d4e13eb8750a375b9511e4dd059092c0e481/spec/rest/stats.test.js#L8-L51 and "setup helper":https://github.com/ably/ably-js/blob/4e65d4e13eb8750a375b9511e4dd059092c0e481/spec/common/modules/testapp_manager.js#L158-L182 as an example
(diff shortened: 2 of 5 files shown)


textile/features.textile
```diff
@@ -54,8 +54,13 @@ h2(#test-guidelines). Test guidelines
...
-** @(G4a)@ When sending the API version to Ably, the SDK must send the exact string specified in @G4@. Therefore, it is recommended that the SDK treat the version opaquely, as a string, not a float.
+* @(G4)@ This clause has been replaced by "@G5@":#G5 as of version 2.0 of this specification
+** @(G4a)@ This clause has been replaced by "@G5c@":#G5c as of version 2.0 of this specification
+* @(G5)@ A client library must identify the version of this specification it uses in all requests and connections to the Ably service, using the mechanisms described in "@RSC7e@":#RSC7e and "@RTN2h@":#RTN2h.
+** @(G5a)@ When sending the API version to Ably, the SDK must send an exact string in the form @MAJOR.MINOR@ where the @MAJOR@ and @MINOR@ components are defined according to "SemVer":https://semver.org/ in the "the @ably/specification@ repository":https://github.com/ably/specification at the commit containg the specification source to which the SDK is conforming.
```
SimonWoolf, line 60: It's not clear from this what it means for the components to be defined "in the spec repo .. at the commit". Where in the repo? (It's clarified by the "Versioning" paragraph in the readme (which says: in package.json), but that's a bit hidden, someone reading the spec won't necessarily know to look there)

SimonWoolf (commented):
Honestly, I'm not sure I'm a big fan of this aesthetically. It effectively annexes package.json to be part of the spec -- you're implementing an sdk and need to know what string to set `x-ably-version` to implement the spec, but to find that info you need to look in package.json.

Which also means it's no longer sufficient to look at https://sdk.ably.com/builds/ably/specification/main/features/ , because package.json isn't exposed there, you need to go to the git repo to get that info. That seems a bit messier to me.

>  This meant that these spec points needed to be ... mutated each time the specification/protocol version was bumped

Is that actually bad thing though? Like, `RSC7a`, in spec v1.2, required you to send the header "X-Ably-Version: 1.2". In spec v2.0, it now requires you to send the header "X-Ably-Version: 2.0". That's a behavioural change, that needs to be implemented. Given your modification/replacement guidelines, isn't it right that it follows the 'replacement' guideline and gets a new spec version, rather than avoiding that through indirection?

(But to be clear: that's just my two cents, the decision is yours to make if you disagree)

QuintinWillison:
Thanks, @SimonWoolf. As you can see, I've created an issue ☝️ to address your opening observation.

In respect of:

> Is that actually bad thing though? Like, `RSC7a`, in spec v1.2, required you to send the header "X-Ably-Version: 1.2". In spec v2.0, it now requires you to send the header "X-Ably-Version: 2.0". That's a behavioural change, that needs to be implemented. Given your modification/replacement guidelines, isn't it right that it follows the 'replacement' guideline and gets a new spec version, rather than avoiding that through indirection?

I agree it's a "behavioural change", but these things are layered and that phrasing will depend from what context you consider it from.

What remains static behaviour is that these headers are always sent, which is what I'm leaving strictly specified - the core part of the protocol, if you will.

What for me constitutes conflation is the inclusion of meta-data in the data (i.e. the version values being included in the spec, and varying from spec version to spec version). The version number should only be defined in a single place, at as high a level as to be appropriate for the source files that it's versioning - in this case, the entire repository contents.

So, while I've considered this observation, my feeling is that the approach I'm proposing is the correct one going forwards. I will be interested to hear what others think.

textile/features.textile
QuintinWillison: I'll admit I had several runs at getting this sentence "right", yet clearly didn't... I'll have a re-think on how to articulate this. Though, perhaps, my [other response](https://github.com/ably/specification/pull/105#issuecomment-1286029865) might have helped you to be able to assist me in re-writing this...?

QuintinWillison: Re-reading this conversation in the light of a new day, I'm not sure I can think of how to re-write this any more clearly. The guiding principle for this work is [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), hence why I remain of the belief that once a reader of this spec point navigates to the source repository that they will find the "Versioning" header+paragraph added in https://github.com/ably/specification/pull/105/commits/860575c43af92bbe884c3cdfafa524b0e1c86755, so this detail does not need repeating elsewhere. I don't think that needs stronger clarification. And, if in future, we decide to move the version definition from `package.json` to some other location then there's only one place we need to update instructions in this repository (the root readme). Also, [as I've indicated elsewhere](https://github.com/ably/specification/pull/105#issuecomment-1286029865), I don't feel the version number is "data" (i.e. spec point worthy), it's "meta data" relating to the source code base as a whole.

lmars: I think it is simpler if the spec explicitly states the version number it represents, rather than stating the commit it is from with the reader then needing to lookup what version that commit corresponds to.

Perhaps we should be considering templating this file, so in the source it has something like:

```
This spec defines API version {{ VERSION }}.
```

and then we render it so that the published spec says:

```
This spec defines API version 2.0.
```

CONTRIBUTING.md
```diff
@@ -42,10 +42,16 @@ by adding a local development HTTP server.
...
-- **Removal**: When removing a spec item, it must remain but replace all text with “This clause has been deleted.”. See [#1057](https://github.com/ably/docs/pull/1057) for an example of this in practice.
+- **Addition**: When adding a new spec item, choose an ID that is greater than all others that exist in the given section, even if there is a gap in the currently assigned IDs. 
+- **Modification**: Spec items should never be mutated, except to patch a mistake that doesn't change the semantics for SDK implementations. Follow the guidance outlined here in respect of _Replacement_ if the meaning or scope of a spec point needs to change.
+- **Removal**: When removing a spec item, it must remain but replace all text with `This clause has been deleted as of version Y.Z of this specification`. See [#1057](https://github.com/ably/docs/pull/1057) for an example of this in practice.
```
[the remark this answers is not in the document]
QuintinWillison: That is very true. 😅 Currently we seem to always have a clear vision of where we're headed, so this is the standard we've informally adopted (it's already in the features spec, including [in the v2 integration branch](https://github.com/ably/specification/pull/108), hence I was just writing it down)... however, now you mention this, this does need more thought. 🤔

QuintinWillison:
I've moved this pull request back to _Draft_ state while I work on refactoring to address comments received so far.

textile/features.textile
```diff
@@ -423,8 +429,9 @@ h3(#realtime-connection). Connection
 ** @(RTN2b)@ @echo@ should be @true@ by default; @false@ will prevent messages published by the client being echoed back
 ** @(RTN2d)@ @clientId@ contains the provided @clientId@ option of @ClientOptions@, unless @clientId@ is @null@
 ** @(RTN2e)@ Depending on the authentication scheme, either @accessToken@ contains the token string, or @key@ contains the API key
-** @(RTN2f)@ API version param @v@ should be the API version per "G4":#G4
+** @(RTN2h)@ Protocol version param @v@ must be the populated, where the value to be sent is defined in "@G5a@":#G5a
 ** @(RTN2g)@ The library and version (in the form described in "RSC7b":#RSC7b) should be included as the value of a @lib@ querystring param. For example, the 1.0.0 version of the JavaScript library would use the param @lib=js-1.0.0@
+** @(RTN2f)@ This clause has been replaced by "@RTN2h@":#RTN2h as of version 2.0 of this specification
```
lawrence-forooghian, line 434: How come RTN2f has been moved from its previous position? Would it not be best for it to be alongside the point it's replaced by?

textile/features.textile
lawrence-forooghian: I think that if @lmars's suggested approach works, then that would preferable.

textile/features.textile
QuintinWillison: Because of my interpretation of the 'Ordering' bullet here: https://github.com/ably/specification/blob/main/CONTRIBUTING.md#features-spec-points

I may be interpreting that wrong. Perhaps @SimonWoolf can add thoughts around this, as this convention pre-dates my time at Ably.

textile/features.textile
```diff
@@ -93,6 +98,7 @@ h3(#restclient). RestClient
 *** @(RSC7d6)@ Libraries may offer a @ClientOptions#agents@ property, for use only by other Ably-authored SDKs, on a need-to-have basis.
 **** @(RSC7d6a)@ The product/version key-value pairs supplied to that property should be injected into all @Agent@ library identifiers emitted by connections made as a result of REST or Realtime instances created using those @ClientOptions@.
 **** @(RSC7d6b)@ An API commentary must be provided for this property. This commentary must make it clear that this interface is only to be used by Ably-authored SDKs.
+** @(RSC7a)@ This clause has been replaced by "@RSC7e@":#RSC7e as of version 2.0 of this specification
```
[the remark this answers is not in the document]
QuintinWillison: My [other answer](https://github.com/ably/specification/pull/105#discussion_r1011485563) also applies here.

textile/features.textile
lawrence-forooghian: Sorry, but I don't think I understand what you mean. It says "priority should be placed on ordering them in a way that makes coherent sense". For me, that would mean that a replaced spec point should be alongside the thing it's been replaced by. Please could you explain your interpretation of the spec point?

PR #112, opened 2022-11-14
Require agents specified by `ClientOptions#agents` to have a version

lawrence-forooghian:
This comes from [a conversation I had with Quintin](https://github.com/ably/ably-cocoa/issues/1525#issuecomment-1311449934) about trying to figure out the best way to represent “this agent has no version” in Objective-C, where a dictionary cannot contain a nil value.

He suggested that it would be best to simply remove this scenario by requiring a version, since the `#agents` property is only for use by other Ably SDKs (as already made explicit by [RSC7d6b](https://sdk.ably.com/builds/ably/specification/main/features/#RSC7d6b)), which will always have a version.


owenpearson (commented):
It doesn't logically follow from [RSC7d6b](https://sdk.ably.com/builds/ably/specification/main/features/#RSC7d6b) that the Ably authored SDK won't ever have a requirement to append an unversioned agent. I think your point in the PR description assumes that the SDKs using this option will only ever use it to represent themselves which I don't think is a safe assumption (even if it's probably true for all SDKs currently).

If objective-c doesn't allow dictionaries with nil values is there not something like the rust `Option` type that we can use instead?

lawrence-forooghian:
> I think your point in the PR description assumes that the SDKs using this option will only ever use it to represent themselves which I don't think is a safe assumption

That's a very good point.

> If objective-c doesn't allow dictionaries with nil values is there not something like…

Yep, there are definitely ways that we can get around it, I just went with the spec change because it was what Quintin was leaning towards, but I think given your point above it makes more sense for me to address this at the ably-cocoa level. Gonna put this back into draft and remove reviewers whilst I think about it. Thanks for your input!

lawrence-forooghian:
Closing this one.

PR #113, opened 2022-11-14
Define API for accessing Agent identifier and its components

### LINKED ISSUE #111: Specify API for Ably wrapper SDKs to obtain 'inner' agents components
As observed in https://github.com/ably/ably-asset-tracking-common/pull/66#issuecomment-1309993388.

It would be useful for Ably wrapper SDKs to be able to obtain a 'rendered', aggregate agents string, matching what the SDK sends to the Ably service.

This needs specifying. It _might_ fit as a new client options API, however - even if static in nature (i.e. a class, not instance, method) - we have nothing similar currently in that API.

lawrence-forooghian:
Closes #111. See commit messages for more details.

I'm open to names other than `ClientLibraryInfo`, given that in fact it returns an combination of information about the library and information about the platform it's running on.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -35,6 +35,7 @@ jump_to:
     - Data types#types
     - Options#options
     - Push notifications#types-push
+    - Client library introspection#introspection
   Interface Definition:
     - Complete API IDL#idl
   Previous version:
@@ -1655,6 +1656,13 @@ h4. DevicePushDetails
 * @(PCP3)@ recipient - a map of string key/value pairs containing details of the push transport and address
 * @(PCP4)@ state - the state of the push registration, one of  @Active@, @Failing@, @Failed@
 
+h3(#introspection). Client library introspection
+
+h4. ClientInformation
+* @(CR1)@ Provides information about the client library and the environment in which it’s running.
+* @(CR2)@ @agents@ static property:
+** @(CR2a)@ Returns a @Dict<String, String?>@ that lists the default key-value entries that the library uses to populate the @Agent@ library identifier, as described by "@RSC7d1@":#RSC7d1. An example would be @{ "ably-java": "1.2.1", "android": "24" }@.
+
 h3(#defaults). Client Library defaults
 
 The following default values are configured for the client library:
@@ -2348,6 +2356,9 @@ class DeltaExtras // DE*
 class ReferenceExtras: // REX*
   timeserial: String // REX2a
   type: String //REX2b
+
+class ClientInformation: // CR*
+  +agents: Dict<String, String?> // CR2
 </pre>
 
 h2(#old-specs). Old specs
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1662,6 +1662,11 @@ h4. ClientInformation
 * @(CR1)@ Provides information about the client library and the environment in which it’s running.
 * @(CR2)@ @agents@ static property:
 ** @(CR2a)@ Returns a @Dict<String, String?>@ that lists the default key-value entries that the library uses to populate the @Agent@ library identifier, as described by "@RSC7d1@":#RSC7d1. An example would be @{ "ably-java": "1.2.1", "android": "24" }@.
+* @(CR3)@ @agentIdentifier@ static method:
+** @(CR3a)@ The client library must only offer this method if it offers the @ClientOptions#agents@ property described by "@RSC7d6@":#RSC7d6.
+** @(CR3b)@ Returns the @Agent@ library identifier described by "@RSC7d@":#RSC7d.
+** @(CR3c)@ Accepts an @additionalAgents@ parameter of type @Dict<String, String?>?@, whose value is to be interpreted with the same semantics as the @ClientOptions#agents@ property. The @agentIdentifier@ method will inject these agents into the @Agent@ library identifier that it returns.
+** @(CR3d)@ An API commentary must be provided for this method. This commentary must make it clear that this interface is only to be used by Ably-authored SDKs.
 
 h3(#defaults). Client Library defaults
 
@@ -2359,6 +2364,7 @@ class ReferenceExtras: // REX*
 
 class ClientInformation: // CR*
   +agents: Dict<String, String?> // CR2
+  +agentIdentifier(additionalAgents: Dict<String, String?>?) => String // CR3; interface only offered by some libraries
 </pre>
 
 h2(#old-specs). Old specs


owenpearson (changes requested):

textile/features.textile
```diff
@@ -1655,6 +1656,17 @@ h4. DevicePushDetails
...
 
+h3(#introspection). Client library introspection
+
+h4. ClientLibraryInfo
+* @(CLI1)@ Provides information about the client library.
+* @(CLI2)@ @agents@ static property:
+** @(CLI2a)@ Returns a @Dict<String, String?>@ that lists the default key-value entries that the library uses to populate the @Agent@ library identifier, as described by "@RSC7d@":#RSC7d.
+* @(CLI3)@ @agentIdentifier@ static method:
+** @(CLI3a)@ Returns the @Agent@ library identifier described by "@RSC7d@":#RSC7d.
+** @(CLI3b)@ If the library offers the @ClientOptions#agents@ property described by "@RSC7d6@":#RSC7d6, then the @agentIdentifier@ method must accept an @additionalAgents@ parameter of type @Dict<String, String?>?@, whose value is to be interpreted with the same semantics as the @ClientOptions#agents@ property. The @agentIdentifier@ method will inject these agents into the @Agent@ library identifier that it returns.
```
maratal, line 1667: `agents` being presented in two places looks like we couldn't decide the right one tbh.

lawrence-forooghian: Hey @maratal, I'm not sure I understand what you mean here. Would you be able to elaborate a bit please?

textile/features.textile
```diff
@@ -2349,6 +2361,9 @@ class ReferenceExtras: // REX*
   timeserial: String // REX2a
   type: String //REX2b
 
+class ClientLibraryInfo: // CLI*
```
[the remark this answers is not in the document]
lawrence-forooghian: Do you think that it needs to link to other SDK classes? It's a class which only has static methods (i.e. you can use its functionality without needing an instance of it), so I thought people wanting to use it would just access the class from the global namespace.

textile/features.textile
```diff
@@ -1655,6 +1656,17 @@ h4. DevicePushDetails
...
+h3(#introspection). Client library introspection
+
+h4. ClientLibraryInfo
+* @(CLI1)@ Provides information about the client library.
+* @(CLI2)@ @agents@ static property:
+** @(CLI2a)@ Returns a @Dict<String, String?>@ that lists the default key-value entries that the library uses to populate the @Agent@ library identifier, as described by "@RSC7d@":#RSC7d.
+* @(CLI3)@ @agentIdentifier@ static method:
+** @(CLI3a)@ Returns the @Agent@ library identifier described by "@RSC7d@":#RSC7d.
+** @(CLI3b)@ If the library offers the @ClientOptions#agents@ property described by "@RSC7d6@":#RSC7d6, then the @agentIdentifier@ method must accept an @additionalAgents@ parameter of type @Dict<String, String?>?@, whose value is to be interpreted with the same semantics as the @ClientOptions#agents@ property. The @agentIdentifier@ method will inject these agents into the @Agent@ library identifier that it returns.
+** @(CLI3c)@ An API commentary must be provided for the @additionalAgents@ parameter, if present. This commentary must make it clear that only Ably-authored SDKs should pass a non-null value for this parameter.
```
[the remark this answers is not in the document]
lawrence-forooghian: I was doing it to be consistent with [RSC7d6b](https://sdk.ably.com/builds/ably/specification/main/features/#RSC7d6b), which states that the concept of specifying additional agents is exclusively for Ably-authored SDKs.

Do you think it would be better for me to change CLI3c to say that this whole method should be exclusively for Ably-authored SDKs?

textile/features.textile
maratal: I mean `ClientLibraryInfo.agents` and 'ClientOptions.agents'

textile/features.textile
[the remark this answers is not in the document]
owenpearson: OK, I may have misread this a bit the first time around, but yeah I think it would be better if the method itself was labelled as Ably-only 👍

textile/features.textile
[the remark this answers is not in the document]
owenpearson: Oh interesting, I hadn't thought of that. I'm not sure how that would work when libraries need to gather information after initialisation to populate the agent string (eg ably-js needs to run `process.version()` to get the NodeJS version... maybe we could run that as an import side-effect but I would prefer not tbh).

Also, is the expectation that any agents configured via the `ClientOptions.agent` option would not be present in this list?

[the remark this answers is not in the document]
lawrence-forooghian: > I'm not sure how that would work when libraries need to gather information after initialisation to populate the agent string (eg ably-js needs to run `process.version()` to get the NodeJS version

Perhaps I've not understood properly here, but is your concern to do with the fact that `ClientLibraryInfo.agents` is specified as a property here and hence you're worried that we wouldn't be able to run code to derive its value? If you were to think of it as a class method instead of a property would that solve this issue?

> Also, is the expectation that any agents configured via the `ClientOptions.agent` option would not be present in this list?

Correct. The `ClientOptions#agents` property is used to configure a specific _instance_ of `RestClient` or `RealtimeClient`. The intention is that the `ClientLibraryInfo` interface provides information about the library without needing to instantiate a client.

textile/features.textile
lawrence-forooghian: They do different things though – `ClientLibraryInfo.agents` is for querying the library’s version, and `ClientOptions.agents` is for telling the library which additional agents to use.

textile/features.textile
```diff
@@ -2349,6 +2361,9 @@ class ReferenceExtras: // REX*
   timeserial: String // REX2a
   type: String //REX2b
 
+class ClientLibraryInfo: // CLI*
+  +agents: Dict<String, String?> // CLI2
+  +agentIdentifier(additionalAgents: Dict<String, String?>?) => String // CLI3; support for the additionalAgents parameter is optional
```
maratal, line 2366: How is it supposed to provide these `additionalAgents` to `ClientOptions`, which now only accepts dictionary as an input?

textile/features.textile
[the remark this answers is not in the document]
owenpearson: Thanks for the explanation. Yeah, a method would work 👍 Although now that you've explained your intention of the word "default", would that also exclude stuff like environment/runtime versions? If so then we could probably get away with it just being a property.

textile/features.textile
```diff
@@ -1655,6 +1656,17 @@ h4. DevicePushDetails
 * @(PCP3)@ recipient - a map of string key/value pairs containing details of the push transport and address
 * @(PCP4)@ state - the state of the push registration, one of  @Active@, @Failing@, @Failed@
 
+h3(#introspection). Client library introspection
+
+h4. ClientLibraryInfo
```
QuintinWillison, line 1661: I am not a fan of abbreviations in API names (I know we already have some like `Params`, but I would prefer not to add more going forwards), so would prefer expansion to `ClientLibraryInformation` - but that is starting to get a little unwieldy. Perhaps `Library` does not need to be stated, given context of use? And, also, even the word `Information` feels like it's grappling at straws in terms of adding value - these are static methods/properties, so is it not obvious that this is a 'utility' class? 🤔 

So, I wonder whether a shorter name, perhaps a little more aligned with existing `RestClient` and `RealtimeClient` names might be `ClientRuntime`?

textile/features.textile
QuintinWillison, line 1667: This spec point is hard to prove downstream in SDKs, due to variability of the interface shape. Also, when an SDK then goes on to implement `RSC7d6` after having only previously implemented `CLI3b` then we're going to have to make a breaking change just to refactor this method. That feels like planned maintenance headache to me.

Might it not just be clearer to state that this method should only be implemented if `RSC7d6` has _also_ been implemented? Then it will always have the same signature.

QuintinWillison (changes requested):

textile/features.textile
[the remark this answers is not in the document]
lawrence-forooghian: It does include environment/runtime versions, yep. I've added an example to CR2a to make this clear.

textile/features.textile
lawrence-forooghian: Naming is horrible, isn't it?! I have made your suggested change, and am happy to hear others’ opinions too.

textile/features.textile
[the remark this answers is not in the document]
owenpearson: awesome, that example looks good thanks

textile/features.textile
lawrence-forooghian: I've changed it to say that "The client library must offer this method if and only if it offers the ClientOptions#agents property described by RSC7d6".

textile/features.textile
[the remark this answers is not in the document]
lawrence-forooghian: I've added that as CR3d.

textile/features.textile
lawrence-forooghian: I'm not sure what you mean here. There’s no interaction between `ClientLibraryInfo` and `ClientOptions`… would you be able to explain in a bit more detail please?

textile/features.textile
owenpearson: I definitely agree that we don't need `Library` there. I'm not sure how I feel about "Runtime" though, I think someone looking through an SDK API will probably be more confused by that than if we just call it `ClientInformation`, @QuintinWillison wdyt?

textile/features.textile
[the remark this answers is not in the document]
lawrence-forooghian: I've added a bit about the choice to use a property in the commit message: https://github.com/ably/specification/pull/113/commits/4654b93e4237c70ac5457d4294f29ef291739c19

textile/features.textile
QuintinWillison: Fair point around specificity. I would also be happy with `ClientInformation`.

textile/features.textile
```diff
@@ -1655,6 +1656,18 @@ h4. DevicePushDetails
...
 * @(PCP4)@ state - the state of the push registration, one of  @Active@, @Failing@, @Failed@
 
+h3(#introspection). Client library introspection
+
+h4. ClientRuntime
+* @(CR1)@ Provides information about the client library and the environment in which it’s running.
+* @(CR2)@ @agents@ static property:
+** @(CR2a)@ Returns a @Dict<String, String?>@ that lists the default key-value entries that the library uses to populate the @Agent@ library identifier, as described by "@RSC7d1@":#RSC7d1. An example would be @{ "ably-java": "1.2.1", "android": "24" }@.
+* @(CR3)@ @agentIdentifier@ static method:
+** @(CR3a)@ The client library must offer this method if and only if it offers the @ClientOptions#agents@ property described by "@RSC7d6@":#RSC7d6.
```
QuintinWillison, line 1666: I'm worried that we're inferring that the client library must implement this method if it's implemented `RSC7d6`. Perhaps clearer like this?:

```suggestion
** @(CR3a)@ The client library must only offer this method if it also offers the @ClientOptions#agents@ property described by "@RSC7d6@":#RSC7d6.
```

QuintinWillison (changes requested):

textile/features.textile
QuintinWillison: Thanks, though I think it needs further refinement. Hence: https://github.com/ably/specification/pull/113#pullrequestreview-1182941317

textile/features.textile
lawrence-forooghian: Ah yes, that was bad wording on my part. I've changed it to your suggestion 👍

textile/features.textile
lawrence-forooghian: Changed it to `ClientInformation`.

QuintinWillison (approved):

owenpearson (approved):
LGTM

textile/features.textile
maratal: resolved in DM

maratal (approved):
LGTM

textile/features.textile
maratal: How about `AgentInformation`, since `ClientInformation` sounds _a bit_ fishy @QuintinWillison @owenpearson

lawrence-forooghian:
Hi @QuintinWillison, I really appreciate your consideration in addressing the merge conflicts introduced into this PR after https://github.com/ably/specification/pull/116 was merged. However, I prefer not to have merge commits on my feature branches (for reasons that you probably have a good idea of by now based on our previous conversations 😁). I've removed your change and rebased onto `main` instead.

textile/features.textile
lawrence-forooghian: I'm going to leave this one for the three of you to battle out haha. @QuintinWillison @owenpearson did you have any opinions on @maratal’s question? I'm keen to get this PR merged pretty soon.

QuintinWillison: I understand where @maratal is coming from, however our context is an Ably SDK and this means that 'client' should be well understood as having come from 'client library'. I also think that 'Agent' start to be too specific again, perhaps inferring that the APIs here-presented relate only to the agents headers sent to the Ably service on connection. I think we should stick with `ClientInformation` and get this PR landed!

QuintinWillison:
[Point](https://github.com/ably/specification/pull/113#issuecomment-1319255890) taken, @lawrence-forooghian, as I do understand you do prefer rebase over merge (per [allowance in guidance](https://github.com/ably/engineering/blob/main/sdk/pull-requests.md#flexible-regarding-merge-from-or-rebase-atop-target)). 😉

I have, however, added https://github.com/ably/specification/pull/113/commits/0542408673da9575c455b1cd8ecc5dabc7d2fe68 via the GitHub UI in order to verify that checks pass before I (now) merge to `main` (target) branch. Our branch protection rule for the `main` (default) branch has "Require branches to be up to date before merging" checked, which I do feel is important.

PR #116, opened 2022-11-17
Format IDL by presenting it within an HTML `pre` element

### LINKED ISSUE #11: Format code-fenced blocks in Textile sources
I've created this issue to offset solving this problem until later, rather than trying to solve it during the lifespan of https://github.com/ably/specification/pull/4.

There's presumably something in the docs toolchain that is doing some additional magic that's beyond RedCloth's basic capabilities, or I've missed something in terms of how I've configured RedCloth, or perhaps how my Ruby code is treating the textile content. Either way, this does not need to be a blocker for the aforementioned pull request to land.

To see what I'm talking about, take a look at [my current rendering of the IDL section of the features spec.](https://sdk.ably.com/builds/ably/specification/pull/4/features/#idl). The code-fence is rendering to HTML literally:

Screenshot 2022-10-02 at 18 50 27

If I look at [our current docs site rendering](https://docs.ably.com/client-lib-development-guide/features/#idl) then it renders a `code` element within a `pre` element (which, in itself seems a bit curious) producing the desired result:

Screenshot 2022-10-02 at 18 46 35

QuintinWillison:
We only had one location in the textile documents [moved over](https://github.com/ably/specification/blob/main/textile/notes-in-relation-to-move-from-ably-docs.md) from the original `ably/docs` home that this needed fixing in, being the IDL within our features spec.. There was no need for '[Block code](https://textile-lang.com/doc/block-code)' (`code` elements _within_ outer `pre` element) so I've kept it simple and replaced with just [Pre-formatted text](https://textile-lang.com/doc/pre-formatted-text).

When `features.textile` was rendered by the tooling in the docs repository then this backtick-fenced block was passing through [this filter](https://github.com/ably/docs/blob/65da30b98f3175e6d07d9f320d7f49ce758d57ef/lib/ably_pre_textile_filter.rb#L119), which is very clever but completely over the top for what we need in this repository context. Let's, instead, just use basic textile functionality - which is what the change in this pull request does.

Closes #11.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1687,8 +1687,7 @@ Please note the following conventions:
 
 Each type, method, and attribute is labelled with the name of one or more clauses from the preceding feature spec. An asterisk is used as a wildcard to mean all clauses with a given prefix.
 
-```
-class RestClient: // RSC*
+pre.. class RestClient: // RSC*
   constructor(keyOrTokenStr: String) // RSC1
   constructor(ClientOptions) // RSC1
   auth: Auth // RSC5
@@ -2348,8 +2347,6 @@ class DeltaExtras // DE*
 class ReferenceExtras: // REX*
   timeserial: String // REX2a
   type: String //REX2b
-
-```
 
 h2(#old-specs). Old specs
 


Peter-Maguire (approved):

owenpearson (changes requested):
Seems like tailwind is setting `margin: 0` on all `pre` elements, can we override that? the default for chrome is `1em 0px` which looks much nicer IMO

QuintinWillison:
Thanks, @owenpearson - while it's in `rem` rather than `em` (see [tailwind docs](https://tailwindcss.com/docs/margin)), does https://github.com/ably/specification/pull/116/commits/0157482b344b2833ce4ad9e6c43218f68eb108a9 improve this for you?

owenpearson:
@QuintinWillison I was actually thinking more about vertical margin, I think it looks quite compact with no spaces between the classes (although I think the horizontal margin is an improvement too). An alternative approach is to just put the whole IDL in one `pre` element since they preserve whitespace.

QuintinWillison:
@owenpearson, between https://github.com/ably/specification/pull/116/commits/227bd5dc9664edf7591b788702531d6ce1d4379d and https://github.com/ably/specification/pull/116/commits/3f9b1a4c66bec678e7507e90f5681d4e77153ecf I've hopefully addressed your observation around spacing plus further improved the formatting.

QuintinWillison:
ok, I've now made it look rubbish elsewhere - e.g. [here](https://sdk.ably.com/builds/ably/specification/pull/116/protocol/#other-message-structs). Am going to roll back to Draft again so I can fix that up. Sorry for the noise.

QuintinWillison:
This isn't quite the small and quick pull request I had hoped it would be this morning, but on the flip side I think we're tidying up presentation across the board so it's a good improvement. I've removed other 'Block code' instances in https://github.com/ably/specification/pull/116/commits/b552ae1ddb8d0744f06c23d885a1d52699125989 - looking at how GitHub markup code blocks, they only use `pre` elements too, so this feels like the appropriate approach going forwards (also reading MDN docs on `pre` vs `code`).

owenpearson (approved):
Looks great now 😄

PR #119, opened 2022-11-25
Spec 2.0: make api-compatible with 1.2

SimonWoolf:
Revert the mandatory version param for rest#request, and re-add recoveryKey, per decision that spec v2 (no-connection-serials) is going to be sdk api compatible with 1.2, and breaking changes are being moved to a spec 3.0.


SimonWoolf:
@QuintinWillison dunno how you want to deal with the Gx 'test guidelines' section, I imagine you want to rewrite it, but for now I've thrown in a commit that tweaks the wording of G4 per your guidence about separating the protocol version and the spec version https://github.com/ably/specification/pull/119/commits/ec00d1710a83f6a903fd279e8edc9452f4a5fc96

QuintinWillison (approved):
Thanks. I think the full impact of this change will be much clearer to see once this pull request has been merged into the `integration/protocol-2.0` target branch.

The "guidance about separating the protocol version and the spec version" which @SimonWoolf refers to relates to work that is still only on my local workstation but which I was able to preview with Simon when we had a call last week to discuss things. I will be pushing up a pull request for that today, hopefully.

lmars:
@SimonWoolf 

> Revert the mandatory version param for rest#request ... per decision that spec v2 (no-connection-serials) is going to be sdk api compatible with 1.2

This came up in an internal conversation, is there any further background to this decision? I believe we have been starting to assume this would be part of the v2 API (e.g. in documentation for channel enumeration changes).

lmars:
Thanks for the clarification 👍 seems like we need to reintroduce the changes from this PR and publish as v3?

PR #108, opened 2022-10-21
Integration: Protocol Version `2`

QuintinWillison:
All of the changes included in this pull request have already been through a round of code review under their original pull requests, so this pull request represents the aggregate change set for protocol version `2` and specification version `2.0.0`.

**Reviewers**: It's up to you the depth of review you give this pull request but I expect that it should be a once-over / sanity-check.

Once merged to `main` (default) branch then, per [release process](https://github.com/ably/specification/pull/125), we'll push the following Git tags:

- `v2.0.0` - static and immutable
- `v2.0` - mutable
- `v2` - mutable

Which should ([fingers crossed](https://github.com/ably/specification/pull/126)) publish tagged builds.

---

_The integration branch should never be rebased (force pushed) - we have a branch protection rule configured for all integration branches in this repository that disallows that._

PR #123, opened 2022-12-07
Fallback retry conditions: split up RSC15d into individual items, add a new one

SimonWoolf:
per https://ably.atlassian.net/browse/INC-230


textile/features.textile
```diff
@@ -115,9 +115,9 @@ h3(#restclient). RestClient
 *** @(RSC15b1)@ @ClientOptions#restHost@ has not been set to an explicit value (see "RSC11a":#RSC11a) and @ClientOptions#port@ is not set and @ClientOptions#tlsPort@ is not set
 *** @(RSC15b2)@ An array of @ClientOptions#fallbackHosts@ is provided
 *** @(RSC15b3)@ the deprecated @ClientOptions#fallbackHostsUseDefault@ option is set to @true@
-** @(RSC15k)@ When none of the statements in "RSC15b":#RSC15b are true then host fallback does not apply, and failing HTTP requests that would have "qualified for a retry against a fallback host (see RSC15d)":#RSC15d will instead result in an error immediately
+** @(RSC15k)@ When none of the statements in "RSC15b":#RSC15b are true then host fallback does not apply, and failing HTTP requests that would have "qualified for a retry against a fallback host (see RSC15l)":#RSC15l will instead result in an error immediately
```
QuintinWillison, line 118: This is an interesting case of spec point mutation that I hadn't considered when working on https://github.com/ably/specification/pull/120. However, I still think it's too much of a change to this spec point to permit modification like this as it does imply change required for SDK implementations relating directly to this spec point when the change they need to act upon is only the replacement of `RSC15d` with `RSC15l`.

Therefore I think the change here could be much simpler - being no change at all. Because we keep spec points in place in the case of _Replacement_ and the replacement text includes a pointer to the replacement spec point, then I would prefer to see `RSC15d` to stay here (and for `RSC15a` too). That makes this diff smaller and, I argue, the change easier to reason about and understand when looking back in future.

QuintinWillison (changes requested):

owenpearson (approved):
I don't know under which conditions you would get a 400+ error from cloudfront so can't comment on that, but in terms of SDK implementation changes, all LGTM

textile/features.textile
SimonWoolf: :+1: reverted that change

textile/features.textile
```diff
@@ -126,8 +126,13 @@ h3(#restclient). RestClient
...
+** @(RSC15d)@ This clause has been replaced by "@RSC15l@":#RSC15l. It was valid up to and including specification version @1.2@.
+** @(RSC15l)@ Errors that necessitate use of an alternative host include any of the following conditions. (Resending requests that have failed for other failure conditions will not fix the problem and will simply increase the load on other datacenters unnecessarily).
+*** @(RSC15l1)@ host unresolvable or unreachable
+*** @(RSC15l2)@ request timeout
+*** @(RSC15l3)@ a response with an HTTP status code in the range @500 <= code <= 504@
+*** @(RSC15l4)@ 🆕 a response with a `Server: CloudFront` header and an HTTP status code @>= 400@
```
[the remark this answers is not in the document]
SimonWoolf: nah I just chucked that in for funsies

textile/features.textile
```diff
@@ -126,7 +126,12 @@ h3(#restclient). RestClient
...
 ** @(RSC15j)@ Requests to fallback hosts must use a matching Host header as this is necessary when fallbacks are proxied through a CDN. For example, if a request to @rest.ably.io@ fails and will be retried to @c.ably-realtime.com@, the Host header must be set to @c.ably-realtime.com@ in the retried request
-** @(RSC15d)@ Errors that necessitate use of an alternative host include: host unresolvable or unreachable, request timeout, or a response but with an applicable HTTP status code in the range @500 <= code <= 504@. Resending requests that have failed for other failure conditions will not fix the problem and will simply increase the load on other datacenters unnecessarily
+** @(RSC15d)@ This clause has been replaced by "@RSC15l@":#RSC15l. It was valid up to and including specification version @1.2@.
```
QuintinWillison, line 129:
```suggestion
** @(RSC15d)@ This clause has been replaced by "@RSC15l@":#RSC15l.
```

textile/features.textile
```diff
@@ -517,7 +522,9 @@ h3(#realtime-connection). Connection
...
-** @(RTN17d)@ Errors that necessitate use of an alternative host include: host unresolvable or unreachable, connection timeout, or a @DISCONNECTED@ response with an @error.statusCode@ in the range @500 <= code <= 504@ or HTTP response status code in the range @500 <= code <= 504@. Attempting to reconnect to a fallback host for other failure conditions will not fix the problem and will simply increase the load on other data-centers unnecessarily
+** @(RTN17d)@ This clause has been replaced by "@RTN17f@":#RTN17f. It was valid up to and including specification version @1.2@.
```
QuintinWillison, line 525:
```suggestion
** @(RTN17d)@ This clause has been replaced by "@RTN17f@":#RTN17f.
```

QuintinWillison (changes requested):
Noticing that this pull request is targeting the `main` (default) branch, the 'bleeding edge' of the current specification version `1.2`, rather than the `integration/protocol-2.0` branch (#108)... I think we should simplify these statements, as the new form guidance is only introduced from specification version `2.0.0` (#120).

(or, alternatively, if you prefer then you could rebase this PR atop the v2 integration branch)

SimonWoolf:
Given we're making it on ably-go which does not confirm to 2.0 spec, I think it probably makes sense to keep this pr against `main`. Made those changes

QuintinWillison (approved):

PR #137, opened 2023-04-05
Spec: add RTP2b1a to clarify edge-case in newness comparison

SimonWoolf:
Context: https://github.com/ably/realtime/pull/4779


owenpearson (approved):
LGTM

PR #139, opened 2023-04-28
New batch specs

### LINKED ISSUE #140: Add SDK APIs for new batch REST API format
The new spec should cover:
- Batch publishing
- Batch presence
- Token revocation

These specs should use the `newBatchResponse` query param to enable the new response format

For further information, see:
- [The original batch publish/presence PR](https://github.com/ably/docs/pull/1421/files)
- [The original token revocation PR](https://github.com/ably/docs/pull/1426)
- [The RFC to change the REST batch response format](https://github.com/ably/realtime/issues/4186)

owenpearson:
Resolves #140 

Doing this now so that we can publish this as spec v2.1 before working on spec v3

Adds new specs for batch REST methods, using the new batch response format via the `newBatchResponse` query param.

For further information, see:
- [The original batch publish/presence PR](https://github.com/ably/docs/pull/1421/files)
- [The original token revocation PR](https://github.com/ably/docs/pull/1426)
- [The RFC to change the REST batch response format](https://github.com/ably/realtime/issues/4186)

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -256,6 +256,13 @@ h3(#rest-auth). Auth
 ** @(RSA16b)@ If the library is provided with a token without the corresponding @TokenDetails@, then this holds a @TokenDetails@ instance in which only the @token@ attribute is populated with that token string
 ** @(RSA16c)@ Is set with the current token (if applicable) on instantiation and each time it is replaced, whether the result of an explicit @Auth#authorize@ operation, or a library-initiated renewal resulting from expiry or a token error response
 ** @(RSA16d)@ Is @null@ if there is no current token, including after a previous token has been determined to be invalid or expired, or if the library is using basic auth
+* @(RSA17)@ @Auth#revokeTokens@ function:
+** @(RSA17a)@ Takes a @TokenRevocationTargetSpecifier@ or an array of @TokenRevocationTargetSpecifier@s and sends them in a POST request to /keys/{API_KEY_NAME}/revokeTokens, with the newBatchResponse query param set to "true", where @API_KEY_NAME@ is the API key name obtained by reading @AuthOptions#key@ up until the first @:@ character.
+** @(RSA17b)@ The @TokenRevocationTargetSpecifier@s should be mapped to strings by joining the @type@ and @value@ with a ":" character and sent in the @targets@ field of the request body
+** @(RSA17c)@ Returns an array of @BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>@s. Optionally, in languages where this is idiomatic, an overload may be implemented whereby the method can be called with a single @TokenRevocationTargetSpecifier@ and return a single @BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>@. This is not a feature of the REST API, whose response will still be an array, so if implementing this overload, the SDK will have to extract the element from the array.
+** @(RSA17d)@ If called from a client using token authentication, should raise an @ErrorInfo@ with a @40162@ error code and @401@ status code
+** @(RSA17e)@ Accepts an optional @issuedBefore@ timestamp, represented as milliseconds since the epoch, or a `Time` object if idiomatic to the language
+** @(RSA17f)@ If an @allowReauthMargin@ boolean is supplied, it should be included in the @allowReauthMargin@ field of the request body
 
 h3(#rest-channels). Channels
 
@@ -1555,6 +1562,28 @@ h4. BatchPresenceFailureResult
 ** @(BGF2a)@ @channel@ string - the name of the channel
 ** @(BGF2b)@ @error@ @ErrorInfo@ - @ErrorInfo@ indicating the reason the presence request failed for the given channel
 
+h4. TokenRevocationTargetSpecifier
+
+* @(TRT1)@ Describes which tokens should be affected by a token revocation request
+* @(TRT2)@ The attributes of @TokenRevocationTargetSpecifier@ consist of:
+** @(TRT2a)@ @type@ string - the type of token revocation target specifier (eg. "clientId", "revocationKey", "channel")
+** @(TRT2b)@ @value@ string - the value of the token revocation target specifier
+
+h4. TokenRevocationSuccessResult
+
+* @(TRS1)@ Contains information about the result of a successful token revocation request for a single target specifier
+* @(TRS2)@ The attributes of @TokenRevocationSuccessResult@ consist of:
+** @(TRS2a)@ @target@ string - the target specifier
+** @(TRS2b)@ @appliesAt@ Time - a timestamp at which the token revocation will take effect
+** @(TRS2c)@ @issuedBefore@ Time - a timestamp for which tokens previously issued will be revoked
+
+h4. TokenRevocationFailureResult
+
+* @(TRF1)@ Contains information about the result of an unsuccessful token revocation request for a single target specifier
+* @(TRF2)@ The attributes of @TokenRevocationFailureResult@ consist of:
+** @(TRF2a)@ @target@ string - the target specifier
+** @(TRF2b)@ @error@ @ErrorInfo@ - an @ErrorInfo@ indicating the reason the token revocation request failed for the given specified
+
 h4. MessageFilter
 * @(MFI1)@ Supplies filter options to subscribe as defined in #RTL22
 * @(MFI2)@ Contains the following attributes:
@@ -1872,6 +1901,8 @@ class Auth: // RSA*
   createTokenRequest(TokenParams?, AuthOptions?) => io TokenRequest // RSA9
   requestToken(TokenParams?, AuthOptions?) => io TokenDetails // RSA8
   tokenDetails: TokenDetails? // RSA16
+  revokeTokens(TokenRevocationTargetSpecifier, issuedBefore Time?, allowReauthMargin boolean?) => io BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult> // RSA17
+  revokeTokens(TokenRevocationTargetSpecifier[], issuedBefore Time?, allowReauthMargin boolean?) => io BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>[] // RSA17
 
 class TokenDetails: // TD*
   +fromJson(String | JsonObject) -> TokenDetails // TD7
@@ -2424,6 +2455,19 @@ class BatchPresenceSuccessResult:
 class BatchPresenceFailureResult
   channel: string // BGF2a
   error: ErrorInfo // BGF2b
+
+class TokenRevocationTargetSpecifier:
+  type: string // TRT2a
+  value: string // TRT2b
+
+class TokenRevocationSuccessResult:
+  target: string // TRS2a
+  appliesAt: Time // TRS2b
+  issuedBefore: Time // TRS2c
+
+class TokenRevocationFailureResult:
+  target: string // TRF2a
+  error: ErrorInfo // TRF2b
 </pre>
 
 h2(#old-specs). Old specs
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -162,6 +162,7 @@ h3(#restclient). RestClient
 * @(RSC22)@ @RestClient#batchPublish@ function:
 ** @(RSC22a)@ Takes a @BatchPublishSpec@ or an array of @BatchPublishSpec@s and sends then in a POST request to @/messages@ with the @newBatchResponse@ query param set to "true".
 ** @(RSC22b)@ Returns an array of @BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>@s. Optionally, in languages where this is idiomatic, an overload may be implemented whereby the method can be called with a single @BatchPublishSpec@ and return a single @BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>@. This is not a feature of the REST API, whose response will still be an array, so if implementing this overload, the SDK will have to extract the element from the array.
+* @(RSC23)@ @RestClient#batchPresence@ function takes an array of channel name strings and sends them as a comma separated string in the @channels@ query parameter in a GET request to @/presence@ with the @newBatchResponse@ query param set to "true", returning a @BatchResult<BatchPresenceSuccessResult | BatchPresenceFailureResult>@ object.
 
 h3(#rest-auth). Auth
 
@@ -1540,6 +1541,20 @@ h4. BatchPublishFailureResult
 ** @(BPF2a)@ @channel@ string - the name of the channel
 ** @(BPF2b)@ @error@ @ErrorInfo@ - an @ErrorInfo@ indicating the reason the message(s) failed to publish
 
+h4. BatchPresenceSuccessResult
+
+* @(BGR1)@ Contains information about the result of a successful batch presence request for a single channel
+* @(BGR2)@ The attributes of @BatchPresenceSuccessResult@ consist of:
+** @(BGR2a)@ @channel@ string - the name of the channel
+** @(BGR2b)@ @presence@ @PresenceMessage[]@ - an array containing all members present on the channel
+
+h4. BatchPresenceFailureResult
+
+* @(BGF1)@ Contains information about the result of an unsuccessful batch presence request for a single channel
+* @(BGF2)@ The attributes of @BatchPresenceFailureResult@ consist of:
+** @(BGF2a)@ @channel@ string - the name of the channel
+** @(BGF2b)@ @error@ @ErrorInfo@ - @ErrorInfo@ indicating the reason the presence request failed for the given channel
+
 h4. MessageFilter
 * @(MFI1)@ Supplies filter options to subscribe as defined in #RTL22
 * @(MFI2)@ Contains the following attributes:
@@ -1762,6 +1777,7 @@ class RestClient: // RSC*
   time() => io Time // RSC16
   batchPublish(BatchPublishSpec) => io BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult> // RSC22
   batchPublish(BatchPublishSpec[]) => io BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>[] // RSC22
+  batchPresence(string[]) => io BatchResult<BatchPresenceSuccessResult | BatchPresenceFailureResult>[] // RSC23
 
 class RealtimeClient: // RTC*
   constructor(keyOrTokenStr: String) // RTC12
@@ -1791,6 +1807,7 @@ class RealtimeClient: // RTC*
   time() => io Time // RTC6
   batchPublish(BatchPublishSpec) => io BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult> // RSC22
   batchPublish(BatchPublishSpec[]) => io BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>[] // RSC22
+  batchPresence(string[]) => io BatchResult<BatchPresenceSuccessResult | BatchPresenceFailureResult>[] // RSC23
 
 class ClientOptions: // TO*
   embeds AuthOptions // This is not currently documented in the spec and needs to be – see https://github.com/ably/docs/issues/1476
@@ -2399,6 +2416,14 @@ class BatchPublishSuccessResult:
 class BatchPublishFailureResult:
   channel: string // BPF2a
   error: ErrorInfo // BPF2c
+
+class BatchPresenceSuccessResult:
+  channel: string // BGR2a
+  presence: [PresenceMessage] // BGR2b
+
+class BatchPresenceFailureResult
+  channel: string // BGF2a
+  error: ErrorInfo // BGF2b
 </pre>
 
 h2(#old-specs). Old specs
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -159,6 +159,9 @@ h3(#restclient). RestClient
 ** @(RSC20d)@ All personally identifiable information, as much as is practicable, must be redacted or stripped completely before being submitted to Ably. Our intent is only to capture necessary information to debug issues in our own code
 ** @(RSC20e)@ Failures to log exceptions to the @errors.ably.io@ endpoint must be handled gracefully. This includes for example DNS failures, TCP/HTTP requests rejected, slow requests and internal failure errors. Additionally, as specified in @RSC20b2@, a failure to log an exception is logged with log level @info@ i.e. an exception reporting failure is not consider a client library @error@ or @warning@
 ** @(RSC20f)@ Any errors emitted by the library as a result of an internal failure must contain a status code @500@, an error code in the range @51000@ to @51999@ and a suitable error message. The error code must match one of "our common error codes":https://github.com/ably/ably-common/blob/main/protocol/errors.json
+* @(RSC22)@ @RestClient#batchPublish@ function:
+** @(RSC22a)@ Takes a @BatchPublishSpec@ or an array of @BatchPublishSpec@s and sends then in a POST request to @/messages@ with the @newBatchResponse@ query param set to "true".
+** @(RSC22b)@ Returns an array of @BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>@s. Optionally, in languages where this is idiomatic, an overload may be implemented whereby the method can be called with a single @BatchPublishSpec@ and return a single @BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>@. This is not a feature of the REST API, whose response will still be an array, so if implementing this overload, the SDK will have to extract the element from the array.
 
 h3(#rest-auth). Auth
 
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -1532,6 +1539,22 @@ h4. BatchPresenceResult
 ** @(BGR2b)@ @presence@ @PresenceMessage[]@ - an optional array containing all members present on the channel if the request succeeded
 ** @(BGR2c)@ @error@ @ErrorInfo@ - an optional @ErrorInfo@ indicating the reason the presence request failed for the given channel, if unsuccessful
 
+h4. TokenRevocationTarget
```
lawrence-forooghian, line 1542: Might `TokenRevocationTargetSpecifier` be more appropriate, given this is the language used in the [REST API documentation](https://ably.com/docs/core-features/authentication#target-specifiers)? Or are we using the words "target" and "target specifier" interchangeably?

textile/features.textile
```diff
@@ -1532,6 +1539,22 @@ h4. BatchPresenceResult
...
+
+* @(TRT1)@ Contains information used to specify a token revocation target
```
lawrence-forooghian, line 1544:
```suggestion
* @(TRT1)@ Describes which tokens should be affected by a token revocation request
```

lawrence-forooghian (commented):
As well as the comments below, I also wanted to mention that we need to write documentation comments for these new APIs and open a PR into https://github.com/ably/sdk-api-reference. [The process described by EdX DR 5.4](https://ably.atlassian.net/wiki/spaces/ENG/pages/2289500233/DR+5.4+The+future+of+the+Canonical+API+Ref+table+and+process#The-process-to-keep-the-canonical-table-up-to-date) isn't quite clear about when this should happen, but I’d suggest that it would be best to do it alongside the spec PR (i.e. now), so that reviewers of the spec PR can also understand how the new functionality is meant to be used by users.

textile/features.textile
```diff
@@ -159,6 +159,7 @@ h3(#restclient). RestClient
...
 ** @(RSC20e)@ Failures to log exceptions to the @errors.ably.io@ endpoint must be handled gracefully. This includes for example DNS failures, TCP/HTTP requests rejected, slow requests and internal failure errors. Additionally, as specified in @RSC20b2@, a failure to log an exception is logged with log level @info@ i.e. an exception reporting failure is not consider a client library @error@ or @warning@
 ** @(RSC20f)@ Any errors emitted by the library as a result of an internal failure must contain a status code @500@, an error code in the range @51000@ to @51999@ and a suitable error message. The error code must match one of "our common error codes":https://github.com/ably/ably-common/blob/main/protocol/errors.json
+* @(RSC22)@ @RestClient#batchPublish@ function takes a @BatchSpec@ or an array of @BatchSpec@s and sends then in a POST request to @/messages@ with the @newBatchResponse@ query param set to "true". When called with a single @BatchSpec@ it should return a single @BatchResult<BatchPublishResult>@, otherwise it returns an array of @BatchResult<BatchPublishResults>@.
```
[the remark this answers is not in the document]
owenpearson: > I just wanted to check whether the batch publish REST API can accept both of these request body types, or whether the SDK needs to wrap a single BatchSpec in an array?

The REST API does accept a single batch spec or an array of them

> Does this statement describe the behaviour of the REST API, or a behaviour that needs to be implemented by the SDK?

Nope, this behaviour does need to be implemented by the SDK, I'll update the specs

textile/features.textile
```diff
@@ -1499,6 +1500,29 @@ h4. ChannelMetrics
...
 
+h4. BatchResult
+
+* @(BAR1)@ Contains information about the results of a batch operation
+** @(BAR2)@ The attributes of @BatchResult@ consist of:
+** @(BAR2a)@ @successCount@ number - the number of successful operations
+** @(BAR2b)@ @failureCount@ number - the number of unsuccessful operations
+** @(BAR2c)@ @results@ array - an array of results for the batch operation
+
+h4. BatchSpec
```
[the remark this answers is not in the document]
owenpearson: Good point, done

textile/features.textile
```diff
@@ -1499,6 +1500,29 @@ h4. ChannelMetrics
...
+
+* @(BSP1)@ Describes the messages that should be published by a batch publish operation, and the channels to which they should be published
+* @(BSP2)@ The attributes of @BatchSpec@ consist of:
+** @(BSP2a)@ @channels@ an array of strings - the names of the channels to which all of the messages contained in the @messages@ attribute should be published
+** @(BSP2b)@ @messages@ an array of @Message@ objects - the names of the channels to which all of the messages contained in the @messages@ attribute should be published
+
+h4. BatchPublishResult
+
+* @(BPR1)@ Contains information about the result of publishes requested by a single @BatchSpec@ for a single channel
```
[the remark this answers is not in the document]
owenpearson: Yeah you're right - it's not supposed to be for a single channel, that was my bad. It is indeed possible to correlate messages and channels since the returned `BatchPublishResult` array is in the same order as the provided `BatchSpec` array

textile/features.textile
```diff
@@ -1499,6 +1500,29 @@ h4. ChannelMetrics
...
+* @(BPR2)@ The attributes of @BatchPublishChannelResult@ consist of:
```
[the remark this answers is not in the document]
owenpearson: I've fixed it now - it was supposed to be `BatchPublishResult`

textile/features.textile
```diff
@@ -1499,6 +1500,29 @@ h4. ChannelMetrics
...
+** @(BPR2a)@ @channel@ string - the name of the channel
+** @(BPR2b)@ @messageId@ string - an optional string containing the @messageId@ prefix for the published message(s), if successful
```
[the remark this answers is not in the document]
owenpearson: If you publish multiple messages to the same channel in a batch, the messages will all start with the same prefix,  IIRC the message IDs look like `<some_prefix>:n`, where `n` is a unique non-negative integer. The REST API returns these in order for clients to identify which messages were published as part of a batch request

textile/features.textile
owenpearson: Good point - I've changed it to `TokenRevocationTargetSpecifier`

textile/features.textile
owenpearson: thanks, done

textile/features.textile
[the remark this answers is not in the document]
owenpearson: OK I've updated the specs for `batchPublish` and `revokeTokens` to mention that the response array should be unwrapped when called with a single spec

textile/features.textile
```diff
@@ -254,6 +254,13 @@ h3(#rest-auth). Auth
 ** @(RSA16b)@ If the library is provided with a token without the corresponding @TokenDetails@, then this holds a @TokenDetails@ instance in which only the @token@ attribute is populated with that token string
 ** @(RSA16c)@ Is set with the current token (if applicable) on instantiation and each time it is replaced, whether the result of an explicit @Auth#authorize@ operation, or a library-initiated renewal resulting from expiry or a token error response
 ** @(RSA16d)@ Is @null@ if there is no current token, including after a previous token has been determined to be invalid or expired, or if the library is using basic auth
+* @(RSA17)@ @Auth#revokeTokens@ function:
+** @(RSA17a)@ Takes a @TokenRevocationTarget@ or an array of @TokenRevocationTarget@s and sends them in a POST request to /keys/{API_KEY_NAME}/revokeTokens, with the newBatchResponse query param set to "true", where @API_KEY_NAME@ is the API key obtained from @AuthOptions#key@
```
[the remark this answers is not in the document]
owenpearson: Yep, this is a behaviour of the REST API

textile/features.textile
```diff
@@ -254,6 +254,13 @@ h3(#rest-auth). Auth
...
+** @(RSA17b)@ The @TokenRevocationTarget@s should be mapped to strings by joining the @type@ and @value@ with a ":" character and sent in the @targets@ field of the request body
+** @(RSA17c)@ When called with a single @TokenRevocationTarget@, returns a @BatchResult<TokenRevocationResult>@, otherwise returns an array of @BatchResult<TokenRevocationResult>@s
```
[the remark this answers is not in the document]
owenpearson: Nope, this isn't a behaviour of the REST API, I've updated the spec to reflect this

textile/features.textile
```diff
@@ -161,6 +159,8 @@ h3(#restclient). RestClient
...
 ** @(RSC20f)@ Any errors emitted by the library as a result of an internal failure must contain a status code @500@, an error code in the range @51000@ to @51999@ and a suitable error message. The error code must match one of "our common error codes":https://github.com/ably/ably-common/blob/main/protocol/errors.json
+* @(RSC22)@ @RestClient#batchPublish@ function takes a @BatchPublishSpec@ or an array of @BatchPublishSpec@s and sends then in a POST request to @/messages@ with the @newBatchResponse@ query param set to "true". When called with an array of @BatchPublishSpec@s it should return an array of @BatchResult<BatchPublishResult>@s. When called with a single @BatchPublishSpec@, the REST response will be an array containing a single @BatchResult<BatchPublishResult>@ which should be returned without the enclosing array.
```
SimonWoolf, line 162: > When called with a single @BatchPublishSpec@, the REST response will be an array containing a single @BatchResult<BatchPublishResult>@ which should be returned without the enclosing array.

Should we consider making this optional, giving some sdks the option of not having an overload for the single-batchspec case? It's easy enough in things like typescript, but in eg golang there's no nice way of doing that, you'd have to just have a separate method, which seems not worth the faff

textile/features.textile
```diff
@@ -254,6 +254,13 @@ h3(#rest-auth). Auth
 ** @(RSA16b)@ If the library is provided with a token without the corresponding @TokenDetails@, then this holds a @TokenDetails@ instance in which only the @token@ attribute is populated with that token string
 ** @(RSA16c)@ Is set with the current token (if applicable) on instantiation and each time it is replaced, whether the result of an explicit @Auth#authorize@ operation, or a library-initiated renewal resulting from expiry or a token error response
 ** @(RSA16d)@ Is @null@ if there is no current token, including after a previous token has been determined to be invalid or expired, or if the library is using basic auth
+* @(RSA17)@ @Auth#revokeTokens@ function:
+** @(RSA17a)@ Takes a @TokenRevocationTargetSpecifier@ or an array of @TokenRevocationTargetSpecifier@s and sends them in a POST request to /keys/{API_KEY_NAME}/revokeTokens, with the newBatchResponse query param set to "true", where @API_KEY_NAME@ is the API key obtained from @AuthOptions#key@
```
SimonWoolf, line 258: > `@API_KEY_NAME@ is the API key obtained from @AuthOptions#key@`

it's the key name only, not the full api key. (sdks already know what a key name is because the requestToken endpoint also uses the key name, and they need it to create a signed token request too)

textile/features.textile
```diff
@@ -254,6 +254,13 @@ h3(#rest-auth). Auth
...
+** @(RSA17b)@ The @TokenRevocationTargetSpecifier@s should be mapped to strings by joining the @type@ and @value@ with a ":" character and sent in the @targets@ field of the request body
+** @(RSA17c)@ When called with an array of @TokenRevocationTargetSpecifier@s, returns an array of @BatchResult<TokenRevocationTargetResult>@s. When called with a single @TokenRevocationTargetSpecifier@, the REST response will be an array containing a single @BatchResult<TokenRevocationTargetResult>@ which should be returned without the enclosing array.
```
SimonWoolf, line 260: similar comment to batch publshing

textile/features.textile
```diff
@@ -1510,43 +1508,52 @@ h4. ChannelMetrics
...
+* @(BPR1)@ Contains information about the result of publishes requested by a single @BatchPublishSpec@
+* @(BPR2)@ The attributes of @BatchPublishResult@ consist of:
+** @(BPR2a)@ @channel@ string - the name of the channel
+** @(BPR2b)@ @messageId@ string - an optional string containing the @messageId@ prefix for the published message(s), if successful
+** @(BPR2c)@ @error@ @ErrorInfo@ - an optional @ErrorInfo@ indicating the reason the message(s) failed to publish, if unsuccessful
+
+h4. BatchPresenceResult
+
+* @(BGR1)@ Contains information about the result of a batch presence request for a single channel
+* @(BGR2)@ The attributes of @BatchPublishChannelResult@ consist of:
```
SimonWoolf, line 1537: `BatchPresenceResult`

textile/features.textile
```diff
@@ -1510,43 +1508,52 @@ h4. ChannelMetrics
...
+** @(BGR2a)@ @channel@ string - the name of the channel
+** @(BGR2b)@ @presence@ @PresenceMessage[]@ - an optional array containing all members present on the channel if the request succeeded
+** @(BGR2c)@ @error@ @ErrorInfo@ - an optional @ErrorInfo@ indicating the reason the presence request failed for the given channel, if unsuccessful
+
+h4. TokenRevocationSpecifierTarget
+
+* @(TRT1)@ Describes which tokens should be affected by a token revocation request
+* @(TRT2)@ The attributes of @TokenRevocationTargetSpecifier@ consist of:
+** @(TRT2a)@ @type@ string - either "clientId" or "revocationKey"
```
SimonWoolf, line 1546: or "channel". (or some other type that ably might add support for in future, so sdks probably shouldn't be trying to do any clientside validation on the type, other than that it's a nonempty string)

textile/features.textile
```diff
@@ -1510,43 +1508,52 @@ h4. ChannelMetrics
...
+** @(TRT2b)@ @value@ string - the value of the token revocation target specifier
+
+h4. TokenRevocationResult
+
+* @(TRR1)@ Contains information about the result of a token revocation request for a single target specifier
+* @(TRR2)@ The attributes of @TokenRevocationTargetResult@ consist of:
+** @(TRR2a)@ @target@ string - the target specifier
+** @(TRR2b)@ @appliesAt@ number - an optional timestamp at which the token revocation will take effect, if successful
+** @(TRR2c)@ @issuedBefore@ number - an optional timestamp for which tokens previously issued will be revoked, if successful
+** @(TRR2d)@ @error@ @ErrorInfo@ - an optional @ErrorInfo@ indicating the reason the token revocation request failed for the given specified, if unsuccessful
```
SimonWoolf, line 1556: wonder if we could be noting a slightly stronger type here? Like, the server guarantees that a targetresult will be `{ target, issuedBefore, appliesAt } | { target, error }`, would be nice to encode that guarantee in the spec rather than just marking every field other than `target` as optional, for sdks that can represent that in the type system. Not sure what the best way of representing that here would be though

SimonWoolf (approved):
Looks good, couple comments

textile/features.textile
owenpearson: Updated. I've still kept some instructions on how to obtain the key name for clarity, since this doesn't exist elsewhere in the spec

textile/features.textile
owenpearson: I've changed it so that each 'result' type has a success and a failure variant so that this constraint is expressed correctly.

SimonWoolf: nice :+1:

textile/features.textile
```diff
@@ -2405,6 +2408,45 @@ class ReferenceExtras: // REX*
...
+
+class TokenRevocationSuccessResult:
+  target: string // TRS2a
+  appliesAt: number // TRS2b
+  issuedBefore: number // TRS2c
+
+class TokenRevocationFailureResult:
+  target: string // TRF2a
+  error: ErrorInfo // TRF2b
+>>>>>>> 75f5279 (feat: add `Auth#revokeTokens` specs)
```
lawrence-forooghian, line 2449: Just noticed there's a merge conflict marker here 🙂

textile/features.textile
```diff
@@ -1546,6 +1553,28 @@ h4. BatchPresenceFailureResult
 ** @(BGF2a)@ @channel@ string - the name of the channel
 ** @(BGF2b)@ @error@ @ErrorInfo@ - @ErrorInfo@ indicating the reason the presence request failed for the given channel, if unsuccessful
 
+h4. TokenRevocationSpecifierTarget
```
lawrence-forooghian, line 1556:
```suggestion
h4. TokenRevocationTargetSpecifier
```

textile/features.textile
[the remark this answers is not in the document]
lawrence-forooghian: I remain a bit confused about the structure of the result of `batchPublish(BatchPublishSpec)`.

Would you mind confirming whether this is correct?

> that is, given a `BatchSpec` for publishing $m$ messages to $n$ channels, the `BatchResult.results` array will contain $mn$ `BatchPublishResult` objects

If so, I'm not sure I understand this comment of yours:

> the returned `BatchPublishResult` array is in the same order as the provided `BatchSpec` array

`BatchSpec` is a 2-dimensional array but `BatchPublishResult` is 1-dimensional. So to say that they have the "same order" doesn’t make sense to me — there must be some sort of flattening going on?

textile/features.textile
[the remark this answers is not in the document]
lawrence-forooghian: A couple of questions, in that case:

- isn't the attribute name `messageId` a bit misleading, if it's actually just a prefix of the message’s ID?
- if I've understood the structure of the result of `batchPublish(BatchPublishSpec)` correctly (which I may not have done — see my question elsewhere in this PR), then we have one `BatchPublishSuccessResult` for each message that resulted from the batch operation, so couldn't we actually include the full message ID instead of just the prefix?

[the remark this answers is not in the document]
SimonWoolf: When you do a normal realtime/rest publish of an array of Messages, they're handled atomically. the client library bundles them up into a single ProtocolMessage, that either succeeds or fails. That ProtocolMessage is the thing that actually has the real messageId. But on the subscribe side, because the user doesn't actually interact with the ProtocolMessage only the Messages but we still want to expose that information somehow, the receiving sdk takes the protocolmessage message id (eg `abc123:0`), appends an index for each of the Messages inside, and sets it as the Message.id (eg `abc123:0:0` through `abc123:0:3`). See [TM2a](https://sdk.ably.com/builds/ably/specification/main/features/#TM2a).

In basic pubsub this is fairly transparent to the user; the realtime system only knows about the 'real' messageId, the user only knows the fake ones. But the distinction is kinda important to the guarantees of what the system does (eg the whole ProtocolMessage will succeed or be rejected for a rate limit atomically), and so it is documented. Eg if the user wants to do idempotent publishing with customer-specified ids, [they will need to learn about this](https://faqs.ably.com/client-specified-message-id-restrictions-for-multiple-messages-published-atomically), because the idempotency applies to the ProtocolMessage, not the individual Message. And in the batch api, this abstraction is exposed too -- all messages in a single batchspec are published atomically (for each channel in that batchspec), so by specifying multiple batchspecs, the user has control over which groups of messages are published atomically. The return type reflects this by having one BatchPublishResult for each group of messages atomically published to the system.

It's a bit of a balance between making it easy for newbies and allowing advanced users more insight into and control over what's actually happening. Ofc we may have gotten that balance wrong here. There's still time to propose a change to the v3 batch response type if you want to do so, no customers are using it yet. (eg plausibly realtime could anticipate the numbering that the subscribing sdks will give the individual messages and return `messageIds: ["foo:0:0", "foo:0:1"]` instead of `messageId: "foo"`)

textile/features.textile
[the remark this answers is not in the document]
SimonWoolf: > that is, given a BatchSpec for publishing _m_ messages to _n_ channels, the BatchResult.results array will contain _mn_ BatchPublishResult objects

(right now it will have _n_ since all messages for one channel are bundled together and sent atomically, cf reply to the messageId thread)

textile/features.textile
owenpearson: Nice, thanks! updated

textile/features.textile
owenpearson: thanks! fixed

textile/features.textile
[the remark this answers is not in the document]
owenpearson: > BatchPublishSpec is a 2-dimensional structure (has messages and channels) but BatchResult.results is 1-dimensional. So to say that they have the "same order" doesn’t automatically make sense to me — there must be some sort of flattening going on?

What I meant was: If you post an array of `BatchSpecs`, you will get an array of `BatchResult`s in the same order as the corresponding array of `BatchSpec`s. For each individual `BatchSpec` you can match on channel name within the `BatchResult.results` array to determine which publishes were successful/unsuccessful. The code example [here](https://github.com/ably/docs/pull/1450#issuecomment-1161500432) may be helpful.

textile/features.textile
```diff
@@ -254,6 +256,13 @@ h3(#rest-auth). Auth
...
+** @(RSA17b)@ The @TokenRevocationTargetSpecifier@s should be mapped to strings by joining the @type@ and @value@ with a ":" character and sent in the @targets@ field of the request body
+** @(RSA17c)@ Returns an array of @BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>@s. Optionally, in languages where this is idiomatic, an overload may be implemented whereby the method can be called with a single @TokenRevocationTargetSpecifier@ and return a single @BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>@. This is not a feature of the REST API, whose response will still be an array, so if implementing this overload, the SDK will have to extract the element from the array.
+** @(RSA17d)@ If called from a client using token authentication, should raise an @ErrorInfo@ with a @40162@ error code
+** @(RSA17e)@ If an @issuedBefore@ timestamp is supplied, it should be included in the @issuedBefore@ field of the request body
```
lawrence-forooghian, line 264: Suggestion to be clearer about the meaning of the timestamp number and more idiomatic options:

```suggestion
** @(RSA17e)@ Accepts an optional @issuedBefore@ timestamp, represented as milliseconds since the epoch, or a `Time` object if idiomatic to the language
```

textile/features.textile
```diff
@@ -1856,6 +1887,8 @@ class Auth: // RSA*
   createTokenRequest(TokenParams?, AuthOptions?) => io TokenRequest // RSA9
   requestToken(TokenParams?, AuthOptions?) => io TokenDetails // RSA8
   tokenDetails: TokenDetails? // RSA16
+  revokeTokens(TokenRevocationTargetSpecifier, issuedBefore number?, allowReauthMargin boolean?) => io BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult> // RSA17
```
lawrence-forooghian, line 1890:
```suggestion
  revokeTokens(TokenRevocationTargetSpecifier, issuedBefore Time?, allowReauthMargin boolean?) => io BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult> // RSA17
```

textile/features.textile
```diff
@@ -1856,6 +1887,8 @@ class Auth: // RSA*
...
+  revokeTokens(TokenRevocationTargetSpecifier[], issuedBefore number?, allowReauthMargin boolean?) => io BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>[] // RSA17
```
lawrence-forooghian, line 1891:
```suggestion
  revokeTokens(TokenRevocationTargetSpecifier[], issuedBefore Time?, allowReauthMargin boolean?) => io BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>[] // RSA17
```

textile/features.textile
```diff
@@ -2405,6 +2411,44 @@ class ReferenceExtras: // REX*
...
+  channel: string // BGF2a
+  error?: ErrorInfo // BGF2b
+
+class TokenRevocationTargetSpecifier:
+  type: string // TRT2a
+  value: string // TRT2b
+
+class TokenRevocationSuccessResult:
+  target: string // TRS2a
+  appliesAt: number // TRS2b
```
lawrence-forooghian, line 2446:
```suggestion
  appliesAt: Time // TRS2b
```

textile/features.textile
owenpearson: Good idea 👍

textile/features.textile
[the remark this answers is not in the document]
SimonWoolf: Eh, maybe.

Part of my reluctance is that `BatchPublishSuccessResult` isn't actually anything specific to the batch api. It's literally just the response body you get from a normal REST publish (try it). It'd be wrong from an api coherence point of view to changing the name of the property for the batch publish api but not the normal publish api.

So the question then becomes, should we change it in both for protocol v3? And... maybe, but I'm not quite convinced it's worth the faff of a breaking change, even if yeah it would be more correct (given that, afaik, no customers have ever complained about this. almost no-one ever even looks at the message id of a publish afaik, most people only care if it succeeded or failed)

textile/features.textile
```diff
@@ -2405,6 +2411,44 @@ class ReferenceExtras: // REX*
...
+  channel: string // BPR2a
+  messageId: string // BPR2b
+
+class BatchPublishFailureResult:
+  channel: string // BPF2a
+  error: ErrorInfo // BPF2c
+
+class BatchPresenceSuccessResult:
+  channel: string // BGR2a
+  presence?: [PresenceMessage] // BGR2b
```
lawrence-forooghian, line 2434: @owenpearson @SimonWoolf is `presence` intentionally optional here? Ditto for whether the "if the request succeeded" in BGR2b is redundant.

textile/features.textile
```diff
@@ -2405,6 +2411,44 @@ class ReferenceExtras: // REX*
...
+
+class BatchPresenceFailureResult
+  channel: string // BGF2a
+  error?: ErrorInfo // BGF2b
```
lawrence-forooghian, line 2438: @owenpearson @SimonWoolf is `error` intentionally optional here? Ditto for whether the "if unsuccessful" in BGF2b is redundant.

textile/features.textile
SimonWoolf: yeah you're right --  I guess it was originally written as BatchPublishResult with `channel & presence? & error?` and then later split into `{channel & presence} | {channel and error}` but we forgot to remove the optional mark. similarly in other places

textile/features.textile
```diff
@@ -1510,43 +1510,70 @@ h4. ChannelMetrics
...
+* @(TRT2)@ The attributes of @TokenRevocationTargetSpecifier@ consist of:
+** @(TRT2a)@ @type@ string - the type of token revocation target specifier (eg. "clientId", "revocationKey", "channel")
+** @(TRT2b)@ @value@ string - the value of the token revocation target specifier
+ 
+h4. TokenRevocationSuccessResult
+
+* @(TRS1)@ Contains information about the result of a successful token revocation request for a single target specifier
+* @(TRS2)@ The attributes of @TokenRevocationSuccessResult@ consist of:
+** @(TRS2a)@ @target@ string - the target specifier
+** @(TRS2b)@ @appliesAt@ number - a timestamp at which the token revocation will take effect
```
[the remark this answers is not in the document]
SimonWoolf: realtime returns both as numbers in milliseconds since the epoch (same as all its time apis). We might want sdks to expose / take a Time object where that's idiomatic, I'l lleave that to Owen to decide (if so it should probs be done for both consistently, and maybe consistently with whatever other apis that take times do, eg history start/end)

textile/features.textile
```diff
@@ -254,6 +256,13 @@ h3(#rest-auth). Auth
...
+** @(RSA17a)@ Takes a @TokenRevocationTargetSpecifier@ or an array of @TokenRevocationTargetSpecifier@s and sends them in a POST request to /keys/{API_KEY_NAME}/revokeTokens, with the newBatchResponse query param set to "true", where @API_KEY_NAME@ is the API key name obtained by reading @AuthOptions#key@ up until the first @:@ character.
+** @(RSA17b)@ The @TokenRevocationTargetSpecifier@s should be mapped to strings by joining the @type@ and @value@ with a ":" character and sent in the @targets@ field of the request body
+** @(RSA17c)@ Returns an array of @BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>@s. Optionally, in languages where this is idiomatic, an overload may be implemented whereby the method can be called with a single @TokenRevocationTargetSpecifier@ and return a single @BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>@. This is not a feature of the REST API, whose response will still be an array, so if implementing this overload, the SDK will have to extract the element from the array.
+** @(RSA17d)@ If called from a client using token authentication, should raise an @ErrorInfo@ with a @40162@ error code
```
lawrence-forooghian, line 263: @SimonWoolf @owenpearson this should probably state the `statusCode` too — is a 401 appropriate ("you're not authorised to revoke tokens if you're using token auth") or something else?

SimonWoolf: yeah. 401xx codes almost always correspond to 401 statuscodes unless there's some particular reason to do otherwise, similarly 500xx ->500, 403xx, 400xx etc

PR #142, opened 2023-05-01
Bump specification version to 3.0.0

lawrence-forooghian:
In preparation for adding breaking API changes on the `integration/3.0.0` branch, which this PR targets.


lawrence-forooghian:
I found out from `CONTRIBUTING.md` that we're actually supposed to bump the spec version alongside the content change that necessitates it. So I'm closing this in favour of #143 — sorry about the noise.

PR #146, opened 2023-05-02
fix(RTN20): DISCONNECTING -> DISCONNECTED

owenpearson:
This appears to have been a typo - `DISCONNECTING` is not a valid connection state


SimonWoolf:
looks like the DISCONNECTED case is handled by RTN20b below? (and "When DISCONNECTED ... the client library should immediately transition the state to DISCONNECTED" wouldn't make sense)

textile/features.textile
```diff
@@ -520,7 +520,7 @@ h3(#realtime-connection). Connection
...
 * @(RTN20)@ When the client library can subscribe to the Operating System events for network/internet connectivity changes:
-** @(RTN20a)@ When @CONNECTED@, @CONNECTING@ or @DISCONNECTING@, if the operating system indicates that the underlying internet connection is no longer available, then the client library should immediately transition the state to @DISCONNECTED@ with emit a state change with an appropriate @reason@. This state change will automatically trigger the client library to attempt to reconnect, see @RTN15@ above
+** @(RTN20a)@ When @CONNECTED@, @CONNECTING@ or @DISCONNECTED@, if the operating system indicates that the underlying internet connection is no longer available, then the client library should immediately transition the state to @DISCONNECTED@ with emit a state change with an appropriate @reason@. This state change will automatically trigger the client library to attempt to reconnect, see @RTN15@ above
```
maratal, line 523: I believe that transition from `DISCONNECTED` to `DISCONNECTED` shouldn't be performed, probably this state needs to be removed from this list.

owenpearson:
Ah, my bad, thanks both. I've changed it to just remove the mention of `DISCONNECTING` entirely.

maratal (approved):

Issue #134, opened 2023-03-21
Make queueMessages=false suppress the behaviour of resending pending messages on a new transport?


SimonWoolf:
RTN19a the requirement (after a transport disconnection the client lib must resend any pending (awaiting ack/nack) messages on the new transport) does not currently make an exception for when queueMessages=false.

Arguably it should, because the intention of queueMessages=false is 'never send stale messages', and with this behaviour, we might sometimes be doing that

So, proposal: add RTN19a3 (or RTN7d? it'd fit in either place) to clarify that when queueMessages=false, immediately on transport disconnection, the all messages awaiting ack/nack will have their callback called with an appropriate error, and should not be resent on any new transport

cc @paddybyers

paddybyers:
If we're connected, in what situation will we be queueing messages? The only case I can think of would be presence messages that are waiting pending attachment.

What about the case of a deploy where there is a momentary disconnection - don't we want to try to make that fully transparent? Is there an argument for saying that messages are nacked if the first immediate reconnect attempt fails?

sacOO7:
Do we have more update on this @SimonWoolf ?

SimonWoolf:
https://github.com/ably/specification/pull/150

PR #150, opened 2023-05-23
Add exception to resent-messages-on-new-transport if queueMessages=false

SimonWoolf:
And trivial reword to some related spec items

Implements https://github.com/ably/specification/issues/134

I struggled a bit with how this is split across RTN7 (ack/nack) and RTN19a (transport state change side effects), it's messy. Considered merging everything into RTN7 and getting rid of RTN19a, but gave up - ultimately added it as RTN7d because it was most similar to RTN7c (and it's a connection state change side effect not a transport one), but made sure the two both mention each other.


owenpearson (approved):
LGTM 👍

owenpearson:
@SimonWoolf I reckon `main` unless you have any particular reasons not to? I don't see anyone already using `queueMessages` not being happy with this change in a non-major release

PR #152, opened 2023-06-28
Add `ChannelStateChange.hasBacklog` and return `ChannelStateChange` from attach/subscribe

owenpearson:
See ably/ably-js#1347 for discussion + example implementation.

The motivation for this change is to allow users, when using `rewind=1`, to wait for rewind messages to have been received (this wasn't possible before since there was no way to determine whether a rewind message is to be expected)

Returning `ChannelStateChange` from attach/subscribe is an API convenience which makes sense given that both methods return a promise/future/whatever resolving when the channel is attached and, by convention, information about the attachment is exposed via `ChannelStateChange` objects.

I don't expect there to be a demand for either of these features outside of ably-js so I've described both as optional.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -609,6 +609,7 @@ h3(#realtime-channel). RealtimeChannel
 *** @(RTL4c1)@ The @ATTACH@ ProtocolMessage @channelSerial@ field must be set to the @channelSerial@ of the most recent message/presence ProtocolMessage received on that channel (which will have been stored in the channel per @RTL15b@). If no messages have been received on the channel, the field may be set to @null@ or omitted.
 ** @(RTL4f)@ Once an @ATTACH@ @ProtocolMessage@ is sent, if an @ATTACHED@ @ProtocolMessage@ is not received within the "default realtime request timeout":#defaults, the attach request should be treated as though it has failed and the channel should transition to the @SUSPENDED@ state. The channel will then be subsequently automatically re-attached as described in "RTL13":#RTL13
 ** @(RTL4d)@ A callback (or other language-idiomatic equivalent) can be provided that is called when the channel next moves to one of @ATTACHED@, @DETACHED@, @SUSPENDED@, or @FAILED@ states. In the case of @ATTACHED@ the callback is called with no argument. In all other cases it is called with an @ErrorInfo@ corresponding to the @ChannelStateChange.reason@ of the state change (or a fallback if there is no @reason@) to indicate that the attach has failed. (Note: when combined with RTL4f, this means that if the connection is @CONNECTED@, the callback is guaranteed to be called within @realtimeRequestTimeout@ of the @attach()@ call)
+*** @(RTL4d1)@ Optionally, upon success, the callback may be invoked with the @ChannelStateChange@ object once the channel is attached. If the channel is already attached, it should be invoked with @null@.
 ** @(RTL4e)@ If the user does not have sufficient permissions to attach to the channel, the channel will transition to @FAILED@ and set the @RealtimeChannel#errorReason@
 ** @(RTL4j)@ If the attach is not a clean attach (defined in @RTL4j1@), for example an automatic reattach triggered by "@RTN15c3@":#RTN15c3 or "@RTL13a@":#RTL13a (non-exhaustive), the library should set the "@ATTACH_RESUME@":#TR3f flag in the @ATTACH@ message
 *** @(RTL4j1)@ A 'clean attach' is an attach attempt where the channel has either not previously been attached or has been explicitly detached since the last time it was attached. Note that this is not purely a function of the immediate previous channel state. An example implementation would be to set the flag from an @attachResume@ private boolean variable on the channel, that starts out set to @false@, is set to @true@ when the channel moves to the @ATTACHED@ state, and set to @false@ when the channel moves to the @DETACHING@ or @FAILED@ states.
@@ -1925,7 +1926,7 @@ class RealtimeChannel: // RTL*
   push: PushChannel // RSH7
   modes: readonly [ChannelMode] // RTL4m
   params: readonly Dict<String, String> // RTL4k1
-  attach() => io // RTL4
+  attach() => io ChannelStateChange // RTL4
   detach() => io // RTL5
   history(
     start: Time, // RTL10a
@@ -1937,9 +1938,9 @@ class RealtimeChannel: // RTL*
   publish(Message) => io // RTL6, RTL6i
   publish([Message]) => io // RTL6, RTL6i
   publish(name: String?, data: Data?) => io // RTL6, RTL6i
-  subscribe((Message) ->) => io // RTL7, RTL7a
-  subscribe(String, (Message) ->) => io // RTL7, RTL7b
-  subscribe(MessageFilter, (Message) ->) // RTL22
+  subscribe((Message) ->) => io ChannelStateChange // RTL7, RTL7a
+  subscribe(String, (Message) ->) => io ChannelStateChange // RTL7, RTL7b
+  subscribe(MessageFilter, (Message) ->) io ChannelStateChange // RTL22
   unsubscribe() // RTL8, RTL8c
   unsubscribe((Message) ->) // RTL8, RTL8a
   unsubscribe(String, (Message) ->) // RTL8, RTL8b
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -589,6 +589,7 @@ h3(#realtime-channel). RealtimeChannel
...
 ** @(RTL2h)@ Optionally, for backwards compatibility with 0.8 libraries, the @RealtimeChannel@ @EventEmitter@ can provide an overloaded method that supports @on(ChannelState)@, but must issue a deprecation warning
+** @(RTL2i)@ Optionally, when a channel @ATTACHED@ @ProtocolMessage@ is received, the @ProtocolMessage@ may contain a @HAS_BACKLOG@ bit flag indicating that the channel should expect a backlog of messages from a rewind or resume. The corresponding @ChannelStateChange@ may contain a @hasBacklog@ boolean attribute with value @true@ if the bit flag @HAS_BACKLOG@ was included.
```
lawrence-forooghian, line 592: A couple of thoughts:

- Do we really need to make this optional? Ultimately, it's at the discretion of SDKs whether they implement _any_ of the spec points. It's a very minor thing to implement and seems like we should just say that SDKs should do it.
- The word "may" is a bit overloaded here — it's being used to describe a possible behaviour of the server and also an optional API choice from a client library. Makes it a bit harder to read than I think it needs to be.

textile/features.textile
```diff
@@ -616,6 +617,7 @@ h3(#realtime-channel). RealtimeChannel
...
 ** @(RTL4l)@ If the user has specified a @modes@ array in the @ChannelOptions@ ("@TB2d@":#TB2d), it must be encoded as a bitfield per "@TR3@":#TR3 and set as the @flags@ field of the @ATTACH@ @ProtocolMessage@. (For the avoidance of doubt, when multiple different spec items require flags to be set in the @ATTACH@, the final @flags@ field should be the bitwise OR of them all)
 ** @(RTL4m)@ On receipt of an @ATTACHED@, the client library should decode the @flags@ into an array of @ChannelMode@ s (that is, the same format as @ChannelOptions.modes@) and expose it as a read-only @modes@ field of the @RealtimeChannel@ (or a @getModes()@ method where that is more idiomatic). This should only contain @ChannelMode@ s: it should not contain flags which are not modes (see "@TB2d@":#TB2d)
+** @(RTL4n)@ Optionally, upon success, may return the @ChannelStateChange@ object once the channel is attached. If the channel is already attached, a synthetic @ChannelStateChange@ from @ATTACHED@ to @ATTACHED@ may be returned.
```
lawrence-forooghian, line 620: `RTL2g` says that "The library must never emit a `ChannelState` `ChannelEvent` for a state equal to the previous state". I think our options are to either emit an `UPDATE` `ChannelEvent` or to return `null`.

owenpearson: I don't think RTL2g applies here, it's meant for the `RealtimeChannel` `EventEmitter` and makes sense there because it's helpful to be able to assume that the listener in `channel.on('attached', listener)` only fires when the channel actually becomes attached. In this case i don't think the same reasoning applies. Also, the `UPDATE` channel event is just the name of the event which comes with a `ChannelStateChange`. Since we're not emitting an event here, that's not an option.

Having thought a bit about this I will change it to return `null` though, mainly just to not mislead users into thinking that the channel state has changed just because they have a `ChannelStateChange` object

textile/features.textile
owenpearson: We use discretionary language like "optional" or "may" in plenty of other spec points so I don't see the harm tbh, especially for something like this where we have no plans to implement it in any SDK other than ably-js.

As for your second point, I've tweaked the wording slightly for readability but imo it's still necessary to highlight that both the bit flag and the api are optional. Let me know if you have any specific suggestions to improve it further

lawrence-forooghian: OK, my opinions on making it optional aren't particularly strong.

As for wording, how about this (I'm not sure whether it's better or not):

> `ChannelStateChange` may optionally expose a boolean `hasBacklog` property. This property should be set to `true` if and only if the state change corresponds to an `ATTACHED` `ProtocolMessage` containing a `HAS_BACKLOG` bit flag.

textile/features.textile
```diff
@@ -666,6 +668,7 @@ h3(#realtime-channel). RealtimeChannel
 ** @(RTL7d)@ Messages delivered are automatically decoded based on the @encoding@ attribute; see @RestChannel@ encoding features. Tests should exist to publish and subscribe to encoded messages using the "AES 128":https://github.com/ably/ably-common/blob/main/test-resources/crypto-data-128.json and "AES 256":https://github.com/ably/ably-common/blob/main/test-resources/crypto-data-256.json fixture test data
 ** @(RTL7e)@ If a message cannot be decoded or decrypted successfully, it should be delivered to the listener with the @encoding@ attribute set indicating the residual encoding state, and an error should be logged
 ** @(RTL7f)@ A test should exist ensuring published messages are not echoed back to the subscriber when @echoMessages@ is set to false in the @RealtimeClient@ library constructor
+** @(RTL7g)@ Optionally, upon success, may return the @ChannelStateChange@ object once the channel is attached. If the channel is already attached, a synthetic @ChannelStateChange@ from @ATTACHED@ to @ATTACHED@ may be returned.
```
lawrence-forooghian, line 671: This should probably be written with reference to the "optional callback" instead of the return value since that's the language used in RTL7c. Also I think it needs updating to match the changed wording for the corresponding `#attach` spec point.

textile/features.textile
```diff
@@ -1460,6 +1463,7 @@ h4. ChannelStateChange
 * @(TH5)@ The @ConnectionStateChange@ object contains the @event@ that generated the channel state change
 * @(TH3)@ If the channel state change includes error information, then the @reason@ attribute will contain an @ErrorInfo@ object describing the reason for the error
 * @(TH4)@ The @ChannelStateChange@ object contains an attribute @resumed@ which in combination with an @ATTACHED@ state, indicates whether the channel attach successfully resumed its state following the connection being resumed or recovered. If @resumed@ is true, then the attribute indicates that the attach within Ably successfully recovered the state for the channel, and as such there is no loss of message continuity. In all other cases, @resumed@ is false, and may be accompanied with a "channel state change error reason":#TH3
+* @(TH6)@ The @ChannelStateChange@ object may contain an attribute @hasBacklog@ which, upon transitioning to @ATTACHED@, indicates whether the channel should expect a backlog of messages from a resume or rewind.
```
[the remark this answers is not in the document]
owenpearson: done

textile/features.textile
owenpearson: cool, i've changed it to that

textile/features.textile
```diff
@@ -589,6 +589,7 @@ h3(#realtime-channel). RealtimeChannel
...
 ** @(RTL2h)@ Optionally, for backwards compatibility with 0.8 libraries, the @RealtimeChannel@ @EventEmitter@ can provide an overloaded method that supports @on(ChannelState)@, but must issue a deprecation warning
+** @(RTL2i)@ @ChannelStateChange@ may optionall expose a boolean @hasBacklog@ property. This property should be set to @true@ if and only if the state change corresponds to an @ATTACHED@ @ProtocolMessage@ containing a @HAS_BACKLOG@ bit flag.
```
lawrence-forooghian, line 592:
```suggestion
** @(RTL2i)@ @ChannelStateChange@ may optionally expose a boolean @hasBacklog@ property. This property should be set to @true@ if and only if the state change corresponds to an @ATTACHED@ @ProtocolMessage@ containing a @HAS_BACKLOG@ bit flag.
```

textile/features.textile
```diff
@@ -608,6 +609,8 @@ h3(#realtime-channel). RealtimeChannel
...
 ** @(RTL4d)@ A callback (or other language-idiomatic equivalent) can be provided that is called when the channel next moves to one of @ATTACHED@, @DETACHED@, @SUSPENDED@, or @FAILED@ states. In the case of @ATTACHED@ the callback is called with no argument. In all other cases it is called with an @ErrorInfo@ corresponding to the @ChannelStateChange.reason@ of the state change (or a fallback if there is no @reason@) to indicate that the attach has failed. (Note: when combined with RTL4f, this means that if the connection is @CONNECTED@, the callback is guaranteed to be called within @realtimeRequestTimeout@ of the @attach()@ call)
+*** @(RTL4d1)@ Optionally, upon success, the callback may be invoked with the @ChannelStateChange@ object once the channel is attached. If the channel is already attached, it should be invoked with @null@.
+** @(RTL4n)@ Optionally, upon success, may return the @ChannelStateChange@ object once the channel is attached. If the channel is already attached, it should return @null@.
```
lawrence-forooghian, line 613: Am I missing something or is this a differently-worded duplicate of `RTL4d1`?

owenpearson: copy and paste error 🙃

SimonWoolf (approved):

lawrence-forooghian (approved):

PR #154, opened 2023-07-04
refactor(RSC19f1): mention that protocol version is specified by CSV2

owenpearson:


textile/features.textile
```diff
@@ -150,7 +150,7 @@ h3(#restclient). RestClient
...
-*** @(RSC19f1)@ The @version@ parameter specifies the protocol version that the Ably service should use when interpreting the request. The library must include an @X-Ably-Version@ header in the request that it sends, with value given by this argument.
+*** @(RSC19f1)@ The @version@ parameter, as specified by "@CSV2@":#CSV2 determines the protocol version that the Ably service should use when interpreting the request. The library must include an @X-Ably-Version@ header in the request that it sends, with value given by this argument.
```
lawrence-forooghian, line 153: The phrase "as specified by CSV2" reads a bit odd here — the `version` parameter is surely being specified by the current spec point.

I'm not really sure that we have a good spec point to link to here — CSV2 doesn't really define the _concept_ of protocol version, it's more about saying what protocol version this spec is compatible with. I think if we want to link to that part of the spec, the best thing might be to explain how we deviate from it here? 🤷‍♂️

```suggestion
*** @(RSC19f1)@ The @version@ parameter specifies the protocol version that the Ably service should use when interpreting the request. The library must include an @X-Ably-Version@ header in the request that it sends, with value given by this argument. (As such, @request()@ does not use the "@CSV2b@":#CSV2b version number.)
```

lawrence-forooghian (approved):

PR #160, opened 2023-07-19
Added RSH3e2c spec point

### LINKED ISSUE #159: RSH3e2c proposal
In case of successful update, if it was triggered by event other than `CalledActivate`, `updatedCallback:` with no error should be called.

Currently only `updateFailedCallback:` called in case of update error. But now on iOS it's needed to inform a user about successful location device token update.

Specs [RSH3e2](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3e2) and [RSH3e3](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3e3)

```
RSH3e2 On event RegistrationSynced:
   RSH3e2b If the machine is in state WaitingForRegistrationSync as a result of a CalledActivate event, make Push#activate return or call its callback with no error.
   RSH3e2a Transitions to WaitingForNewPushDeviceDetails.

RSH3e3 On event SyncRegistrationFailed:
   RSH3e3c If the machine is in state WaitingForRegistrationSync as a result of a CalledActivate event, make Push#activate return or call its callback with the error.
   RSH3e3a Otherwise, calls the updateFailedCallback provided to Push#activate with the error.
   RSH3e3b Transitions to AfterRegistrationSyncFailed.
```

maratal:
Closes #159

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1184,10 +1184,11 @@ h3(#activation-state-machine). Activation State Machine
 **** @(RSH3e1b)@ Transitions to @WaitingForRegistrationSync@.
 *** @(RSH3e2)@ On event @RegistrationSynced@:
 **** @(RSH3e2b)@ If the machine is in state @WaitingForRegistrationSync@ as a result of a @CalledActivate@ event, make @Push#activate@ return or call its callback with no error.
+**** @(RSH3e2c)@ Otherwise, calls the @updatedCallback@ provided to @Push#activate@ with no error.
 **** @(RSH3e2a)@ Transitions to @WaitingForNewPushDeviceDetails@.
 *** @(RSH3e3)@ On event @SyncRegistrationFailed@:
 **** @(RSH3e3c)@ If the machine is in state @WaitingForRegistrationSync@ as a result of a @CalledActivate@ event, make @Push#activate@ return or call its callback with the error.
-**** @(RSH3e3a)@ Otherwise, calls the @updateFailedCallback@ provided to @Push#activate@ with the error.
+**** @(RSH3e3a)@ Otherwise, calls the @updatedCallback@ provided to @Push#activate@ with the error.
 **** @(RSH3e3b)@ Transitions to @AfterRegistrationSyncFailed@.
 ** @(RSH3f)@ State @AfterRegistrationSyncFailed@:
 *** @(RSH3f1)@ On events @CalledActivate@ or @GotPushDeviceDetails@:
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -2324,8 +2324,8 @@ class Push: RSH1, RSH2
     registerCallback: ((ErrorInfo?, DeviceDetails?) -> io String)?,
     // Only on platforms that, after first set, can update later its push
     // device details:
-    updatedCallback: ((ErrorInfo) ->)?,
-    updateFailedCallback: ((ErrorInfo) ->) // Deprecated, see RSH3e3a and RSH3e3d
+    updateFailedCallback: ((ErrorInfo) ->), // Deprecated, see RSH3e3a and RSH3e3d
+    updatedCallback: ((ErrorInfo?) ->)?
   ) => io ErrorInfo? // RSH2a
   deactivate(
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -1184,10 +1184,11 @@ h3(#activation-state-machine). Activation State Machine
 **** @(RSH3e1b)@ Transitions to @WaitingForRegistrationSync@.
 *** @(RSH3e2)@ On event @RegistrationSynced@:
 **** @(RSH3e2b)@ If the machine is in state @WaitingForRegistrationSync@ as a result of a @CalledActivate@ event, make @Push#activate@ return or call its callback with no error.
+**** @(RSH3e2c)@ Otherwise, calls the @updatedCallback@ provided to @Push#activate@ with no error.
 **** @(RSH3e2a)@ Transitions to @WaitingForNewPushDeviceDetails@.
 *** @(RSH3e3)@ On event @SyncRegistrationFailed@:
 **** @(RSH3e3c)@ If the machine is in state @WaitingForRegistrationSync@ as a result of a @CalledActivate@ event, make @Push#activate@ return or call its callback with the error.
-**** @(RSH3e3a)@ Otherwise, calls the @updateFailedCallback@ provided to @Push#activate@ with the error.
```
lawrence-forooghian, line 1190: Please could you take a look at the documentation about the [specification version](https://github.com/ably/specification#specification-version) and [modifying spec points](https://github.com/ably/specification/blob/main/CONTRIBUTING.md#features-spec-points)?

In particular, assuming that the intention is that we replace `updateFailedCallback` with `updatedCallback` (I don't know since you haven't updated the IDL) then the change proposed here is a breaking change — you're proposing changing the type of `updateFailedCallback` and also the circumstances in which it is called. If you want to introduce a breaking change, it requires a major version increase, and hence needs to go into the `integration/3.0.0` branch. But are you sure that you want to make a breaking change? Could we instead preserve (and deprecate) `updateFailedCallback` and add another optional `updatedCallback`? That, in fact, seems to be what you have done in [your ably-cocoa implementation of location pushes](https://github.com/ably/ably-cocoa/pull/1771/files#diff-007a192d0e318a7b962b4d5816211576f6f86bbdcf6f018373d587ead25d77e3).

maratal: https://github.com/ably/specification/pull/160/commits/08294c01078b268cf8235af27addb8ffb038e7cc

textile/features.textile
```diff
@@ -1184,10 +1184,11 @@ h3(#activation-state-machine). Activation State Machine
 **** @(RSH3e1b)@ Transitions to @WaitingForRegistrationSync@.
 *** @(RSH3e2)@ On event @RegistrationSynced@:
 **** @(RSH3e2b)@ If the machine is in state @WaitingForRegistrationSync@ as a result of a @CalledActivate@ event, make @Push#activate@ return or call its callback with no error.
+**** @(RSH3e2c)@ Otherwise, calls the @updatedCallback@ provided to @Push#activate@ with no error.
```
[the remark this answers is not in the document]
maratal: https://github.com/ably/specification/pull/160/commits/08294c01078b268cf8235af27addb8ffb038e7cc

textile/features.textile
```diff
@@ -2322,7 +2324,8 @@ class Push: RSH1, RSH2
     registerCallback: ((ErrorInfo?, DeviceDetails?) -> io String)?,
     // Only on platforms that, after first set, can update later its push
     // device details:
-    updateFailedCallback: ((ErrorInfo) ->)
+    updatedCallback: ((ErrorInfo) ->)?,
```
lawrence-forooghian, line 2327: I think it's more readable with the optional argument at the end. Also, shouldn't it be

```suggestion
    updatedCallback: ((ErrorInfo?) ->)?,
```

maratal: https://github.com/ably/specification/pull/160/commits/0699da3656e120c524af9c0f76b08957f7da3525

lawrence-forooghian (approved):
@maratal please could you open a corresponding PR in https://github.com/ably/sdk-api-reference? Doesn't need to be right now

PR #161, opened 2023-07-25
Fix return type of `revokeTokens`

lawrence-forooghian:
This was a mistake in 97186e7; the API response is a single `BatchResult`. The single-specifier variant is now redundant so I’ve removed that too.


owenpearson (approved):
LGTM

PR #162, opened 2023-07-26
Fix return type of `batchPresence`

lawrence-forooghian:
This was a mistake in 038b50e; the API response is a single `BatchResult`. RSC23 was already correct, just the IDL needs changing.


owenpearson (approved):
ah my bad, thanks for the fix 👍

PR #166, opened 2023-09-07
Refactored spec definition for new sync proto message sent by server

sacOO7:


textile/features.textile
```diff
@@ -723,7 +723,7 @@ h3(#realtime-presence). RealtimePresence
...
+** @(RTP18a)@ The client library determines that a sync has started whenever a @SYNC @ProtocolMessage@ is received with a @non-empty@ @channelSerial@ attribute. The @channelSerial@ is a two-part identifier @<sync sequence id>:<cursor value>@. If there is an ongoing sync with different @sync sequence id@, it should be ended while discarding related changes and a new sync should be started.
 ** @(RTP18b)@ The sync operation for that sequence identifier has completed once the cursor is empty; that is, when the @channelSerial@ looks like @<sync sequence id>:@
 ** @(RTP18c)@ a @SYNC@ may also be sent with no @channelSerial@ attribute. In this case, the sync data is entirely contained within that @ProtocolMessage@
```
sacOO7, line 728: @SimonWoolf I have not encountered this particular case [RTP18c] yet. Is this still valid or do we need to remove it? 
As of now, I mostly get `channelSerial` in the format `sdfsdf:` with all presence data. So, first two cases `RTP18a` and `RTP18b` are covered. But third one is not covered : (

textile/features.textile
```diff
@@ -743,9 +743,9 @@ h3(#realtime-presence). RealtimePresence
...
+** @(RTP5f)@ If the channel enters the @SUSPENDED@ state then all queued presence messages will fail immediately, and the @PresenceMap@ and "internal PresenceMap (see RTP17)":#RTP17 is maintained. This ensures that if the channel later becomes @ATTACHED@, it will only emit presence events for the changes in the @PresenceMap@ that have occurred (during sync) whilst the channel was @SUSPENDED@. A test should exist for a channel that is in the @SUSPENDED@ state containing presence members to transition to the @ATTACHED@ state, and following the @SYNC@ process after attaching, any members present before and after the sync should not emit presence events, all other changes should be reflected in the @PresenceMap@ and should emit presence events on the channel
```
sacOO7, line 746: Do we also need to add a test that says, it should enter internal presence map members to the channel after attached?
I think it probably should get covered as a part of RTP17f

textile/features.textile
```diff
@@ -723,7 +723,7 @@ h3(#realtime-presence). RealtimePresence
...
 * @(RTP18)@ The realtime system reserves the right to initiate a sync of the presence members at any point once a channel is attached. A server initiated sync provides Ably with a means to send a complete list of members present on the channel at any point
-** @(RTP18a)@ The client library determines that a new sync has started whenever a @SYNC@ @ProtocolMessage@ is received with a @channel@ attribute and a new sync sequence identifier in the @channelSerial@ attribute. The @channelSerial@ is used as the sync cursor and is a two-part identifier @<sync sequence id>:<cursor value>@. If a new sequence identifier is sent from Ably, then the client library must consider that to be the start of a new sync sequence and any previous in-flight sync should be discarded
+** @(RTP18a)@ The client library determines that a sync has started whenever a @SYNC@ @ProtocolMessage@ is received with a @non-empty@ @channelSerial@ attribute. The @channelSerial@ is a two-part identifier @<sync sequence id>:<cursor value>@. If there is an ongoing sync with different @sync sequence id@, it should be ended while discarding related changes and a new sync should be started.
```
SimonWoolf, line 726: > discarding related changes

This makes it sound like the sdk needs to remember what the presence set was like before the previous sync and revert back to it, which is not necessary

textile/features.textile
SimonWoolf: sure feel free to remove RTP18c

textile/features.textile
SimonWoolf: > `the changes in the @PresenceMap@ that have occurred (during sync)`

This addition I think makes it more confusing. The sdk finds out about the changes during the sync, but they might have happened at any point while the channel was suspended

textile/features.textile
```diff
@@ -743,9 +743,9 @@ h3(#realtime-presence). RealtimePresence
...
+** @(RTP5f)@ If the channel enters the @SUSPENDED@ state then all queued presence messages will fail immediately, and the @PresenceMap@ and "internal PresenceMap (see RTP17)":#RTP17 is maintained. This ensures that if the channel later becomes @ATTACHED@, it will only emit presence events for the changes in the @PresenceMap@ that have occurred (during sync) whilst the channel was @SUSPENDED@. A test should exist for a channel that is in the @SUSPENDED@ state containing presence members to transition to the @ATTACHED@ state, and following the @SYNC@ process after attaching, any members present before and after the sync should not emit presence events, all other changes should be reflected in the @PresenceMap@ and should emit presence events on the channel
 ** @(RTP5b)@ If a channel enters the @ATTACHED@ state then all queued presence messages will be sent immediately. A presence @SYNC@ may be initiated per "@RTP1@":#RTP1
-* @(RTP16)@ Connection state conditions:
+* @(RTP16)@ Channel/Connection state conditions side effects while "enter, update and leave presence message (see RTP8, RTP9, RTP10 and RTP14)":#RTP8
```
SimonWoolf, line 748: this addition is quite unclear. "conditions side effects", "while enter presence message"

textile/features.textile
SimonWoolf: > Do we also need to add a test that says, it should enter internal presence map members to the channel after attached? I think it probably should get covered as a part of RTP17f

Just my opinion, but personally, I don't like the "A test should exist that..." spec items. All the spec items specify behaviour, most behaviour should be tested as normal practice. The spec shouldn't need to explicitly require that.

textile/features.textile
sacOO7: I mean it's related to enter, update and leave presence message. Problem is, all specs mentioned right above RTP16 are supposed to be applied while processing presence message. So, this one gets confusing, unless referred to the working code.

sacOO7:
About this https://github.com/ably/specification/pull/166#discussion_r1320075332, yeah it's confusing either way, because you have to consider server `presencemap` and not local `presencemap` while referring to the statement : (

textile/features.textile
sacOO7: So, what is supposed to be done instead? That's the only thing I can think of, reverting presencemap state to previous sync state

sacOO7: Or do we need to wait for current sync to complete and apply the next immediate sync 🤔

textile/features.textile
sacOO7: > > `the changes in the @PresenceMap@ that have occurred (during sync)`
> 
> This addition I think makes it more confusing. The sdk finds out about the changes during the sync, but they might have happened at any point while the channel was suspended

yeah it's confusing either way, because you have to consider server presencemap and not local presencemap while referring to the statement : (

sacOO7: I was confused at first, because for all spec statements, `presencemap` refers to the local one, except this statement.

textile/features.textile
SimonWoolf: > So, what is supposed to be done instead? That's the only thing I can think of, reverting presencemap state to previous sync state

eg clearing the state that needs to be tracked for RTP19, removing any members that you don't see in the sync.

textile/features.textile
sacOO7: yeah, this makes more sense

textile/features.textile
sacOO7: Are we sure about it? This means we will remove check at
https://github.com/ably/ably-dotnet/blob/b359101412f57844c09b8ae276d49008a7adfbd9/src/IO.Ably.Shared/Realtime/Presence.cs#L609
So instead of 
```
if (syncChannelSerial == null || syncCursor.Length <= 1)
```
it will become 
```
if (syncCursor.IsEmpty())
```

sacOO7: Just that I don't want presence to break, if we receive presence data with no channelSerial.
I will be doing this change across dotnet, go and java codebase :  )

PR #167, opened 2023-09-10
connection resume/recover

sacOO7:


textile/features.textile
```diff
@@ -554,7 +555,7 @@ h3(#realtime-connection). Connection
...
 *** @(RTN19a1)@ One possible implementation of this requirement would be to add any in-flight messages to the @RTL6c2@ connection-wide queue of messages that will be sent once the connection next becomes @CONNECTED@
-*** @(RTN19a2)@ In the case of an @RTN15c6@ successful resume, the @msgSerial@ of the reattempted @ProtocolMessage@s should remain the same as for the original attempt. In the case of an @RTN15c7@ failed resume, the message must be assigned a new @msgSerial@ from the SDK's internal counter.
+*** @(RTN19a2)@ In the case of an @RTN15c6@ successful resume, the @msgSerial@ of the reattempted @ProtocolMessage@s should remain the same as for the original attempt. In the case of an @RTN15c7@ failed resume, the message must be assigned a new @msgSerial@ from the SDK's internal counter ( need clarification).
```
sacOO7, line 558: @SimonWoolf not sure what this means, I thought we reset `msgSerial` on resume failure.

textile/features.textile
```diff
@@ -502,16 +502,17 @@ h3(#realtime-connection). Connection
...
+** @(RTN15c)@ The system's response to a resume request (with a non-empty connectionKey) will be one of the following:
+**** @(RTN15c6)@ A @CONNECTED@ @ProtocolMessage@ with the same @connectionId@ as the current client (and no @error@ property). This indicates that the resume attempt was valid.
+**** @(RTN15c7)@ @CONNECTED@ @ProtocolMessage@ with a new @connectionId@ and an @ErrorInfo@ in the @error@ field. In this case, the resume was invalid, and the error indicates the cause. The @error@ should be set as the @reason@ in the @CONNECTED@ event, and as the @Connection#errorReason@. The internal @msgSerial@ counter should be reset so that the first message published to Ably will contain a @msgSerial@ of @0@.
+**** @(RTN15c8)@ Irrespective of success/failure for a resume request, the client library should move all channels that were in the @ATTACHING@, @ATTACHED@, or @SUSPENDED@ states to the @ATTACHING@ state, and initiate an @RTL4c@ attach sequence for each. The connection should also process any messages queued per @RTL6c2@ (there is no need to wait for the attaches to finish before processing queued messages).
```
SimonWoolf, line 515: But this item is a child of a spec item saying "The system's response to a resume request (with a non-empty connectionKey) will be one of the following:". Every other child is one of the four things that can happen:

```
(RTN15c) The system’s response to a resume request will be one of the following:
↳ CONNECTED, same connId
↳ CONNECTED, new connId
↳ token ERROR
↳ other ERROR
```

Adding an item in the middle that's actually is a shared spec item for the first two of the items doesn't make sense

textile/features.textile
```diff
@@ -552,9 +553,9 @@ h3(#realtime-connection). Connection
...
+** @(RTN19a)@ Any @ProtocolMessage@ that is awaiting an @ACK@/@NACK@ on the old transport will not receive the @ACK@/@NACK@ on the new transport (failed resume as per @RTN15c7@). The client library must therefore resend any @ProtocolMessage@ that is awaiting a @ACK@/@NACK@ to Ably in order to receive the expected @ACK@/@NACK@ for that message (subject to @RTN7c@/@RTN7d@). The Ably service is responsible for keeping track of messages, ignoring duplicates and responding with suitable @ACK@/@NACK@ messages
+*** @(RTN19a1)@ One possible implementation of this requirement would be to add those @ACK@/@NACK@ messages to the @RTL6c2@ connection-wide queue of messages that will be sent once the connection becomes @CONNECTED@.
+*** @(RTN19a2)@ In the case of an @RTN15c6@ successful resume, the @msgSerial@ of the reattempted @ProtocolMessage@s remains unchanged. In the case of an @RTN15c7@ failed resume, each message must be assigned a new incremental @msgSerial@ from the Connection#MessageSerial.
```
SimonWoolf, line 558: > remains unchanged

why the change to the present tense? that sentence is saying "don't change it in this case (only in the other case)"

textile/features.textile
```diff
@@ -552,9 +553,9 @@ h3(#realtime-connection). Connection
...
-*** @(RTN19a2)@ In the case of an @RTN15c6@ successful resume, the @msgSerial@ of the reattempted @ProtocolMessage@s should remain the same as for the original attempt. In the case of an @RTN15c7@ failed resume, the message must be assigned a new @msgSerial@ from the SDK's internal counter.
+** @(RTN19a)@ Any @ProtocolMessage@ that is awaiting an @ACK@/@NACK@ on the old transport will not receive the @ACK@/@NACK@ on the new transport (failed resume as per @RTN15c7@). The client library must therefore resend any @ProtocolMessage@ that is awaiting a @ACK@/@NACK@ to Ably in order to receive the expected @ACK@/@NACK@ for that message (subject to @RTN7c@/@RTN7d@). The Ably service is responsible for keeping track of messages, ignoring duplicates and responding with suitable @ACK@/@NACK@ messages
+*** @(RTN19a1)@ One possible implementation of this requirement would be to add those @ACK@/@NACK@ messages to the @RTL6c2@ connection-wide queue of messages that will be sent once the connection becomes @CONNECTED@.
```
SimonWoolf, line 557: No, we're not adding ack/nacks to the queue, the sdk never generates or sends an ack or nack

textile/features.textile
```diff
@@ -552,9 +553,9 @@ h3(#realtime-connection). Connection
...
-*** @(RTN19a1)@ One possible implementation of this requirement would be to add any in-flight messages to the @RTL6c2@ connection-wide queue of messages that will be sent once the connection next becomes @CONNECTED@
-*** @(RTN19a2)@ In the case of an @RTN15c6@ successful resume, the @msgSerial@ of the reattempted @ProtocolMessage@s should remain the same as for the original attempt. In the case of an @RTN15c7@ failed resume, the message must be assigned a new @msgSerial@ from the SDK's internal counter.
+** @(RTN19a)@ Any @ProtocolMessage@ that is awaiting an @ACK@/@NACK@ on the old transport will not receive the @ACK@/@NACK@ on the new transport (failed resume as per @RTN15c7@). The client library must therefore resend any @ProtocolMessage@ that is awaiting a @ACK@/@NACK@ to Ably in order to receive the expected @ACK@/@NACK@ for that message (subject to @RTN7c@/@RTN7d@). The Ably service is responsible for keeping track of messages, ignoring duplicates and responding with suitable @ACK@/@NACK@ messages
```
SimonWoolf, line 556: No, not just failed resume, rtn19a applies to successful resumes as well

SimonWoolf (changes requested):

textile/features.textile
sacOO7: Actually I meant sending a `non-empty connectionKey` in a request

textile/features.textile
sacOO7: I meant `Ack.Nack` messages, not `ack` itself

textile/features.textile
sacOO7: I am not getting. It says new transport, that means it will be a new connection id right? We only get new `connectionId` for a failed resume right? Okay, there can be a case where it's not a resume attempt as such 🤔.
I think spec should be modified so we have explicit understanding that, new transport means `new connectionId`. If it's same connectionId, then we don't need to add it to the queue, and it doesn't have anything to do with resume as such

textile/features.textile
sacOO7: I was not able to understand in-flight messages at all : (
It should explicitly mention type of mesages we should be adding to the queue.

textile/features.textile
SimonWoolf: > I think spec should be modified so we have explicit understanding that, new transport means new connectionId.

It doesn't mean that. A transport is not an Ably Connection, it's the level below, eg a websocket (effectively tcp) connection. This isn't our own jargon, it's OSI terminology

Like -- RTN19a2 explains that you have to do something slightly different if the connectionId changes vs if it stays the same. If the whole of RTN19a only applied if the connectionId changed, that wouldn't make any sense.

> If it's same connectionId, then we don't need to add it to the queue, and it doesn't have anything to do with resume as such

You might think that. But then you'd read RTN19a and discover that you were wrong, the message needs to be resent on any new transport, not just a new connectionId. That's why RTN19a exists 🙂

sacOO7: So, what happens if you have a `new connectionId` and `no error` in errorField. 
1. Successful resume means `same connectionid` and `no error`. 
2. Failed resume means `new connectionid` and `error`
For both of above cases, request should have `connectionkey` sent.
For new transport, there can be a case where no `connectionkey` is sent and new `connectionId` is received and there can a error or no error. So as per new transport ( new connectionId received ), we should append to the queue. So, we can ignore all resume related checks I feel

textile/features.textile
sacOO7: Or we should change it to 
``` @(RTN19a2)@ Only in case of RTN15c7@ `new connection Id`, each message must be assigned a new incremental @msgSerial@ from the Connection#MessageSerial.```

File: textile/features.textile (hunk 5 above)
Comment (sacOO7) [reply textile/features.textile:556]: I am confused here. I thought new transport clearly meant `new connctionId`? Because that's how ably maintains connectionState on the server right? Otherwise, there can always be a new raw transport when new connection is made and without `connectionId` we will not be able to detect if it's coming from same old device.

Comment (sacOO7) [reply textile/features.textile:556]: What if you have a new raw transport and same connectionid? This means connection is resumed yet we won't be able to receive ack for sent messages 🤔. Clearly contradicts with `RTN19b`

File: textile/features.textile (hunk 3 above)
Comment (sacOO7) [reply textile/features.textile:558]: Okay, it seems new transport has nothing to do with maintaining connection state, or `connectionState` doesn't maintain pending ack/nack on the server.

File: textile/features.textile (hunk 5 above)
Comment (sacOO7) [reply textile/features.textile:556]: Okay, it seems new transport has nothing to do with maintaining connection state, or `connectionState` doesn't maintain pending ack/nack on the server. It just maintains pending messages.

File: textile/features.textile [hunk 6]
```diff
@@ -553,9 +554,9 @@ h3(#realtime-connection). Connection
 ** @(RTN17e)@ If the realtime client is connected to a fallback host endpoint, then for the duration that the transport is connected to that host, all HTTP requests, such as history or token requests, should be first attempted to the same datacenter the realtime connection is established with i.e. the same fallback host must be used as the default HTTP request host. If however the HTTP request against that fallback host fails, then the normal fallback host behavior should be followed attempting the request against another fallback host as described in "RSC15":#RSC15
 * @(RTN19)@ Transport state side effects - when a transport is disconnected for any reason:
 ** @(RTN19a)@ Any @ProtocolMessage@ that is awaiting an @ACK@/@NACK@ on the old transport will not receive the @ACK@/@NACK@ on the new transport. The client library must therefore resend any @ProtocolMessage@ that is awaiting a @ACK@/@NACK@ to Ably in order to receive the expected @ACK@/@NACK@ for that message (subject to @RTN7c@/@RTN7d@). The Ably service is responsible for keeping track of messages, ignoring duplicates and responding with suitable @ACK@/@NACK@ messages
```
sacOO7, line 556: Imagine, connection was in `suspendedState`, after several retries, it's connected to the server ( with new transport ) 
As per `RTN8c` and `RTN9c` we are clearing `connectionKey` and `connectionId`. This means a new connection attempt is not a resume attempt. So, there's neither resume failure nor resume success as a part of a new connection. 

In this case,  what should happen to ack/nack messages? Are we supposed to treat it as a resume failure and send ack/nack messages with incremental message-serial ? Same we do it for channel attach in case of `RTN15g3`

sacOO7: Also, it feels we are attaching channels (in attaching, attached or suspended state ) for any type of reconnection 🤔

textile/features.textile
```diff
@@ -502,21 +502,22 @@ h3(#realtime-connection). Connection
...
-**** @(RTN15c7)@ @CONNECTED@ @ProtocolMessage@ with a new @connectionId@ and an @ErrorInfo@ in the @error@ field. In this case, the resume was invalid, and the error indicates the cause. The @error@ should be set as the @reason@ in the @CONNECTED@ event, and as the @Connection#errorReason@. The internal @msgSerial@ counter should be reset so that the first message published to Ably will contain a @msgSerial@ of @0@. The rest of the process is the same as for @RTN16c6@: The client library should move all channels that were in the @ATTACHING@, @ATTACHED@, or @SUSPENDED@ states to the @ATTACHING@ state, and initiate an @RTL4c@ attach sequence for each. The connection should also process any messages queued per @RTL6c2@.
```
sacOO7, line 514: what if there was no resume attempt as a part of suspended retry, are we supposed to treat is as resume failed? That means we need to reset `msgSerial` in this case too 🤔

sacOO7: We will surely get a `new connectionId` but there won't be an error in the errorField. I think this third case is coming in almost every other spec that mentions only dual behavior for resume success/failure.

sacOO7: I think resume failure should be `new connectionId` != `prev connectionId`. We shouldn't have extra checks for errors as such. So, if there is no resume attempt, still it will be a resume failure since connectionId's will not match.

textile/features.textile
```diff
@@ -553,9 +554,9 @@ h3(#realtime-connection). Connection
...
-*** @(RTN19a2)@ In the case of an @RTN15c6@ successful resume, the @msgSerial@ of the reattempted @ProtocolMessage@s should remain the same as for the original attempt. In the case of an @RTN15c7@ failed resume, the message must be assigned a new @msgSerial@ from the SDK's internal counter.
-** @(RTN19b)@ If there are any pending channels i.e. in the @ATTACHING@ or @DETACHING@ state, the respective @ATTACH@ or @DETACH@ message should be resent to Ably
+*** @(RTN19a1)@ One possible implementation of this requirement would be to add those @ACK@/@NACK@ messages to the @RTL6c2@ connection-wide queue of messages that will be sent once the connection becomes @CONNECTED@.
+*** @(RTN19a2)@ In the case of an @RTN15c6@ successful resume, the @msgSerial@ of the reattempted @ProtocolMessage@s remains unchanged. In the case of an @RTN15c7@ failed resume, each message must be assigned a new incremental @msgSerial@ from the Connection#MessageSerial.
+** @(RTN19b)@ If there are any pending channels i.e. in the @ATTACHING@ or @DETACHING@ state, the respective @ATTACH@ or @DETACH@ message should be resent to Ably once reconnected.
```
sacOO7, line 559: Can we make this spec more clear?
Does this mean, we tried to `attach` on old transport, but before receiving `attached`/`detached`, connection was disconnected. So, on connection resume/non-resume, we should resend these messages?

textile/features.textile
sacOO7, line 556: Statement says `The Ably service is responsible for keeping track of messages, ignoring duplicate`. Since, messages published via realtimechannel doesn't have ID, and it's assigned by server, when ack/nacks are republished, we receive duplicate messages for the same. This means, as of now, any realtimechannel that doesn't receive ack/nack on same transport will publish duplicate messages on new transport : (
I have tested this as a part of https://github.com/ably/ably-dotnet/pull/1199/commits/ce4a0730c334c4fbb0681041c58480bcf1ca2e62

sacOO7: Update - 
1. When resume is successful, resent ack/nack doesn't have duplicates -> https://github.com/ably/ably-dotnet/blob/201feb14f3934b3a1c6778dcddbf14deb9c17027/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandboxTransportSideEffectsSpecs.cs#L199

2. When resume is not successful, reset ack/nack have duplicates -> 
https://github.com/ably/ably-dotnet/blob/201feb14f3934b3a1c6778dcddbf14deb9c17027/src/IO.Ably.Tests.Shared/Realtime/ConnectionSandboxTransportSideEffectsSpecs.cs#L190C19-L190C19
 
Not sure how server actually checks for duplicates : (
My only guess is by checking message id's which are not present in both cases. Though for resume success, we have same connection id.

textile/features.textile
```diff
@@ -502,16 +502,17 @@ h3(#realtime-connection). Connection
...
-*** @(RTN15g3)@ When a connection attempt succeeds after the connection state has been cleared in this way, channels that were previously @ATTACHED@, @ATTACHING@, or @SUSPENDED@ must be automatically reattached, just as if the connection was a resume attempt which failed per "RTN15c7":#RTN15c7
+*** @(RTN15g2)@ Another consequence of that is that the measure of whether the client been disconnected for too long (for the purpose of this check) cannot just be whether the client left the @CONNECTED@ state more than @connectionStateTtl@ ago. Instead, it should be whether the difference between the current time and the last activity time (recorded timestamp for last received protocolMessage) is greater than the sum of the @connectionStateTtl@ and the @maxIdleInterval@, where the last activity time is the time of the last known actual sign of activity from Ably per "RTN23a":#RTN23a
```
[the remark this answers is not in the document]
sacOO7: Thanks, will revert this change 👍

textile/features.textile
sacOO7: Okay, though the first 2 do exhibit a common behavior for attaching channels and processing pending messages.

sacOO7: I can remove this if not needed. Though I am not sure if resume failure applies here too as per https://github.com/ably/specification/pull/167#discussion_r1328095814

textile/features.textile
sacOO7: Update - I checked the code, we clear `ack/nack` messages on `suspended` state, so we don't really have to worry about sending them. But can there be a case where a re-connection attempt is not a resume attempt in `disconnected` state?

sacOO7: Update - for the second case, since we are adding ack/nack to pending messages, is it a desired behavior?

textile/features.textile
sacOO7: Update - getting state to `detached` will work on new transport only if `detach` message is explicitly sent to old transport and connection is resumed.
I think it doesn't matter if connection is resumed/not-resumed for attached, for any of the new transport, `attach` will work.

textile/features.textile
```diff
@@ -526,16 +527,16 @@ h3(#realtime-connection). Connection
...
 * @(RTN16)@ @Connection@ recovery:
 ** @(RTN16i)@ Connection recovery is similar to connection resumption (see "RTN15c":#RTN15c), except that instead of the library resuming from a time at which that library instance was previously connected, it is doing so from external state provided in the client options, "(TO3i)":#TO3i. Since the library has no state at the time of connection, the channels must be explicitly attached by the user; continuity preservation is achieved by the @channelSerial@s for each channel being stored in the recovery key.
-** @(RTN16f)@ When a library is instantiated with the @recover@ client option, it should initialize its internal @msgSerial@ counter to the @msgSerial@ component of the @recoveryKey@. (If the recover fails, the counter should be reset to 0 per "RTN15c7":#RTN15c7 )
+** @(RTN16f)@ When a library is instantiated with the @recover@ client option, it should initialize its internal @connectionId@ and @msgSerial@ counter to the @connectionId@ and @msgSerial@ component of the @recoveryKey@. (If the recover fails, the counter should be reset to 0 per "RTN15c7":#RTN15c7 )
```
sacOO7, line 530: If we don't set `connectionId`, `connectionId` will not match for `new AblyRealtime` instance, means resume is not successful as per `RTN15c6`.

textile/features.textile
```diff
@@ -526,16 +527,16 @@ h3(#realtime-connection). Connection
 ** @(RTN20c)@ When @CONNECTING@, if the operating system indicates that the underlying internet connection is now available, the client should restart the pending connection attempt
 * @(RTN16)@ @Connection@ recovery:
 ** @(RTN16i)@ Connection recovery is similar to connection resumption (see "RTN15c":#RTN15c), except that instead of the library resuming from a time at which that library instance was previously connected, it is doing so from external state provided in the client options, "(TO3i)":#TO3i. Since the library has no state at the time of connection, the channels must be explicitly attached by the user; continuity preservation is achieved by the @channelSerial@s for each channel being stored in the recovery key.
```
sacOO7, line 529: Small doubt - 
Not sure why user need to manually attach channels in recover. Seems they need to iterate over each channel and attach them explicitly once recover is successful. 
In case of resume, it will automatically attach the channels.

textile/features.textile
```diff
@@ -526,16 +527,16 @@ h3(#realtime-connection). Connection
 ** @(RTN20c)@ When @CONNECTING@, if the operating system indicates that the underlying internet connection is now available, the client should restart the pending connection attempt
 * @(RTN16)@ @Connection@ recovery:
```
sacOO7, line 528: We also need to add a case for recovery failure ->
Currently, we cover following cases 
1. setting messageserial to zero as a part of `RTN15c7`. 
2. We never set `connectionKey` directly, it will be updated automatically for new connection.

We don't cover 
```
reverting a change made as a part of `RTN16j` -> initialized channels with their serials 
```

textile/features.textile
SimonWoolf: this spec item is basically obselete: the 'send an attach for any attaching' part of this item is no longer needed, redundant to rtn15c6/7. and of course there's no need to send a DETACH for detaching channels any more. So it can just say "If there were any channels in the DETACHING state they should be moved to the DETACHED state"

textile/features.textile
SimonWoolf: > Only in case of new connection Id

I'd really prefer it refer back to the spec item that defines the case. The relevant principle is DRY (Don't Repeat Yourself): if you define the same requirement in multiple places, with slightly different wording each time, the same logic gets implemented multiple times, and then if that needs changing it's way too easy to miss a place it needs to be changed in.

> Okay, Correct me if I am wrong, server connectionState doesn't maintain pending ack/nack on the server. It just maintains pending messages.

yes

textile/features.textile
SimonWoolf: > I meant Ack.Nack contained messages

I don't understand what you mean by this

The types of messages that would be requeued are ones that the library is waiting to hear an ACK or NACK from the server for, which is MESSAGE and PRESENCE protocolmessages per RTN7a

textile/features.textile
SimonWoolf: > this means, as of now, any realtimechannel that doesn't receive ack/nack on same transport will publish duplicate messages on new transport

This is what RTN19a2 is for, which requires the re-send attempt (for a successful resume) to use the same msgSerial.

The message id for a realtime connection is implicitly `connectionId:msgSerial`. If your resume is successful, and you resend with the same msgId, you will be correctly deduplicated.

In the case of a resume failure, you're out of luck, so yes you can get duplicates.

SimonWoolf: > In this case, what should happen to ack/nack messages?

they'll have been cleared already per RTN7c

> can there be a case where a re-connection attempt is not a resume attempt in disconnected state?

yes: before you're connected for the first time

textile/features.textile
SimonWoolf: You could store the connectionId if you like, yes. But it's not strictly necessary, because you can just use whether there was an `error` property in the CONNECTED to decide whether the recover succeeded.

It's true that the spec is unclear on this. Maybe a sentence like "In the case of recover, in the absence of a connectionId to compare, the library can use whether there is an `@error@` in the `CONNECTED` message to determine whether the recover succeeded"?

textile/features.textile
SimonWoolf: think about it. In the case of recover the subscribe listeners don't exist yet, the user needs to re-subscribe. If we automatically attached the channels, the sdk might receive and process all the resume message backlog before anyone's attached any subscribe listeners yet, so the messages would be discarded

textile/features.textile
SimonWoolf: it doesn't need to be reverted. the channelSerials are fine to keep. (similarly we don't clear any channel state in the case of an RTN15c7 resume failure)

textile/features.textile
SimonWoolf: > what if there was no resume attempt as a part of suspended retry, are we supposed to treat is as resume failed? That means we need to reset msgSerial in this case too 🤔

RTN15g: "_If a client has been disconnected for longer than the @connectionStateTtl@, it should not attempt to resume. Instead, it should clear the local connection state, and any connection attempts should be made as for a fresh connection_" -- the msgSerial is part of the connection state, so it will have been reset in this case. (Perhaps the spec should explicitly list the properties that should be cleared, for clarity?)

(see also RTN15g1 which explains why RTN15g is strictly stronger than just doing this on going into SUSPENDED)

> Also, we will get a new connectionId but there won't be an error in the errorField. I think this third case is coming for almost every other spec that mentions just a dual behavior for resume success/failure.

There's two cases. Either connection state (including connectionKey and msgSerial) is still there, in which case the library will try and do a resume, and the possible cases are listed in RTN15c. Or the connection state has been cleared, including the msgSerial, in which case the library is just doing a clean connection. The sdk never stops trying to resume but keeps its msgSerial.

textile/features.textile
sacOO7: Then we should mention sending MESSAGE and PRESENCE type messages associated with acks/nacks.

textile/features.textile
sacOO7: They should be explicitly moved to detached state or those channels will automatically receive detached message from the server on new transport?

sacOO7: Btw, this is still not clear, `Only if `detach` is sent on old transport, new transport will be detached`

textile/features.textile
sacOO7: fair 👍

textile/features.textile
sacOO7: In the `ably-js`, I can find `messageSerial` being cleared as a part of `clearing connection state`. 

https://github.com/ably/ably-js/blob/d735e23d41b558925b02fab0420832f80a77491a/src/common/lib/transport/connectionmanager.ts#L1087-L1092

But, I couldn't find any references to clearing messageSerial as a part of https://sdk.ably.com/builds/ably/specification/main/features/#RTN8c and https://sdk.ably.com/builds/ably/specification/main/features/#RTN9c

textile/features.textile
```diff
@@ -679,7 +680,7 @@ h3(#realtime-channel). RealtimeChannel
...
+* @(RTL12)@ An attached channel may receive an additional @ATTACHED@ @ProtocolMessage@ from Ably at any point. (This is typically triggered following a transport being resumed to indicate a partial loss of message continuity on that channel, in which case the @ProtocolMessage@ will have a @resumed@ flag set to false). If the @resumed@ flag is false, this should result in the channel emitting an @UPDATE@ event with a @ChannelStateChange@ object, otherwise don't set the state or emit attached event. The @ChannelStateChange@ object should have both @previous@ and @current@ attributes set to @attached@, the @reason@ attribute set to to the @error@ member of the @ATTACHED@ @ProtocolMessage@ (if any), and the @resumed@ attribute set per the @RESUMED@ bitflag of the @ATTACHED@ @ProtocolMessage@. (Note that @UPDATE@ should be the only event emitted: in particular, the library must not emit an @ATTACHED@ event if the channel was already attached, see @RTL2g@).
```
sacOO7, line 683: This change has been raised as per https://github.com/ably/ably-go/pull/621/files#r1412495177

maratal: _If the @resumed@ flag is false ... otherwise don't set the state or emit attached event._

Why not to begin sentence with "_If the @resumed@ flag is true..."? This "not not" gives that confusion that we try to avoid of.

sacOO7: I also had a question, where client sends explicit attach on already attached channel. I could see `ATTACHED` message received for duplicate `ATTACH` sent from client side.

textile/features.textile
```diff
@@ -502,21 +502,22 @@ h3(#realtime-connection). Connection
...
+** @(RTN15c)@ The system's response to a resume request (with a non-empty connectionKey) will be one of the following:
+**** @(RTN15c6)@ A @CONNECTED@ @ProtocolMessage@ with the same @connectionId@ as the current client (and no @error@ property). This indicates that the resume attempt was valid.
+**** @(RTN15c7)@ @CONNECTED@ @ProtocolMessage@ with a new @connectionId@ and an @ErrorInfo@ in the @error@ field. In this case, the resume was invalid, and the error indicates the cause. The @error@ should be set as the @reason@ in the @CONNECTED@ event, and as the @Connection#errorReason@. The internal @msgSerial@ counter should be reset so that the first message published to Ably will contain a @msgSerial@ of @0@.
```
sacOO7, line 514: Spec doesn't mention about value of channel `Flag.resumed` after channel attach @SimonWoolf ?

PR #170, opened 2023-10-09
Improve wording in RTN15a

lawrence-forooghian:
- Replace "recovery key" (which misleadingly suggests RTN16 recovery) with "private connection key" (as used in RTN15b).
- Add a link to RTN15b so it’s clear what is meant by "restore the connection state".


maratal (approved):
LGTM

textile/features.textile
```diff
@@ -501,7 +501,7 @@ h3(#realtime-connection). Connection
...
-** @(RTN15a)@ If a @Connection@ transport is disconnected unexpectedly or because a token has expired, then the @Connection@ manager will immediately attempt to reconnect and restore the connection state. Connection state recovery is provided by the Ably service and ensures that whilst the client is disconnected, all events are queued and channel state is retained on the Ably servers. When a new connection is made with the correct connection recovery key, the client is able to catch up by receiving the queued @ProtocolMessages@ from Ably.
+** @(RTN15a)@ If a @Connection@ transport is disconnected unexpectedly or because a token has expired, then the @Connection@ manager will immediately attempt to reconnect and restore the connection state, by following the process described in "RTN15b":#RTN15b. Connection state recovery is provided by the Ably service and ensures that whilst the client is disconnected, all events are queued and channel state is retained on the Ably servers. When a new connection is made with the correct private connection key, the client is able to catch up by receiving the queued @ProtocolMessages@ from Ably.
```
SimonWoolf, line 504: honestly, what even is the point of this spec item? it specifies that an unexpected disconnection should result in an immediate retry, sure, but then goes off on a tangent and starts vaguely waffling on about the point of connection recovery...?

tempted to say this spec item should just read: `** @(RTN15a)@ If a @Connection@ transport is disconnected unexpectedly or because a token has expired, then the SDK must transition to the @CONNECTING@ state and make an immediate  "RTN15b":#RTN15b reconnection attempt.` and that's it

lawrence-forooghian:
This was fixed in https://github.com/ably/specification/pull/204, closing

PR #179, opened 2024-01-09
Spec on what should happen when the recover option is malformed.

maratal:
See https://github.com/ably/ably-cocoa/pull/1714/files#r1441646226 for the context.


textile/features.textile
```diff
@@ -533,6 +533,7 @@ h3(#realtime-connection). Connection
 * @(RTN16)@ @Connection@ recovery:
 ** @(RTN16i)@ Connection recovery is similar to connection resumption (see "RTN15c":#RTN15c), except that instead of the library resuming from a time at which that library instance was previously connected, it is doing so from external state provided in the client options, "(TO3i)":#TO3i. Since the library has no state at the time of connection, the channels must be explicitly attached by the user; continuity preservation is achieved by the @channelSerial@s for each channel being stored in the recovery key.
 ** @(RTN16f)@ When a library is instantiated with the @recover@ client option, it should initialize its internal @msgSerial@ counter to the @msgSerial@ component of the @recoveryKey@. (If the recover fails, the counter should be reset to 0 per "RTN15c7":#RTN15c7 )
+*** @(RTN16f1)@ If recover fails due to a malformed data in the provided client options "(TO3i)":#TO3i, then an error should be logged and the connection should be made like no recover option was provided.
```
lawrence-forooghian, line 536:
```suggestion
*** @(RTN16f1)@ If the recovery key provided in the @recover@ client option cannot be deserialized due to malformed data, then an error should be logged and the connection should be made like no @recover@ option was provided.
```

lawrence-forooghian (commented):
Adding @SimonWoolf as a reviewer too.

SimonWoolf (approved):
ok by me

(in an ideal world we'd maybe modify the next `connected` connectionstatechange event to add an error if none is present already saying that  we were unable to action the recover due to it being malformed. But that's faff, this important enough to start spending resources implementing that in every sdk)

PR #181, opened 2024-02-04
Updated the spec so that it reflects the behaviour implemented in ably-cocoa

### LINKED ISSUE #180: DeviceDetails.id (PCD2) should be nullable.
Currently it is declared as non-nil for some reason:

```
class DeviceDetails: // PCD*
  id: String // PCD2
```

which leads to the bunch of problems and inconsistencies in the SDK(s) implementations.

maratal:
Updated the spec so that it reflects the behaviour implemented in https://github.com/ably/ably-cocoa/pull/1847, where a new device ID is generated as soon as the previous ones are cleared, so that id is never null.

Originally posted by @lawrence-forooghian [here](https://github.com/ably/specification/issues/180#issuecomment-1898664043)

Closes #180

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1345,7 +1345,7 @@ h3(#activation-state-machine). Activation State Machine
 ***** @(RSH3a2a2)@ If a custom @registerCallback@ was provided to @Push#activate@, pass it the local @DeviceDetails@.
 ***** @(RSH3a2a3)@ Otherwise, makes an asynchronous HTTP PUT request to @/push/deviceRegistrations/:deviceId@ using the local @DeviceDetails@ with the push details as body. When the registration validation request is complete, a @RegistrationSynced@ or @SyncRegistrationFailed@ event should be fired.
 ***** @(RSH3a2a4)@ Transitions to @WaitingForRegistrationSync@.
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -1207,7 +1207,7 @@ h3(#activation-state-machine). Activation State Machine
 *** @(RSH3g1)@ On event @CalledDeactivate@:
 **** @(RSH3g1a)@ Transitions to @WaitingForDeregistration@.
 *** @(RSH3g2)@ On event @Deregistered@:
-**** @(RSH3g2a)@ Clears all local @DeviceDetails@.
+**** @(RSH3g2a)@ Generates a new pair of device @id@ and @deviceSecret@ as described in "RSH3a2b":#RSH3a2b. Clears all other local @DeviceDetails@.
```
lawrence-forooghian, line 1210: This only partially describes the ably-cocoa behaviour, doesn't it? In ably-cocoa, `id` and `deviceSecret` generation are completely decoupled from the `CalledActivate` event, meaning that RSH3a2b and RSH8b would also need to be changed if we wished for the spec to describe the behaviour of ably-cocoa.

Also, same point as https://github.com/ably/specification/pull/182#discussion_r1486624999 re spec point modification.

SimonWoolf:
I'm not really familiar with the context here or the push activation state machine implemented in sdks, so going to untag myself and suggest @owenpearson instead

lawrence-forooghian:
@maratal are you intending to update this one?

maratal:
> @maratal are you intending to update this one?

Will take a closer look.

maratal:
> @maratal are you intending to update this one?

@lawrence-forooghian finally fixed

PR #186, opened 2024-03-20
fix: Push Notification corner cases

ttypic:
After thoroughly examining our Push Notification flow, I have built a [diagram](https://miro.com/app/board/uXjVKe-Go5w=/#tpicker-content), and some of the problems have become clear.

## We can't ignore  `CalledDeactivate` event when we are in `NotActive` state

Currently, we are ignoring the `CalledDeactivate` event when we are in the `NotActive` state, which causes a lot of problems later on.

### Example:

- We ended up in `NotActive` with valid device identity token (e.g. haven't received registration update). 
- User logout from the app and calls deactivate (as we recommend to do), nothing happens. 
- User login with another `clientId` calls activate and get error `clientId mismatch`  

## We shouldn't use token / key auth during deregistration call

Users potentially can change apps, and we can handle this by relying solely on device token authentication during the deregistration call; otherwise, we might end up in a broken state.

### Example:

If we use device token auth together with token or key auth from another Ably app we'll receive `Token unrecognised` error.

## We should handle unauthorized errors or other problems with device identity token

When we encounter `401` status errors or code `40005` errors (invalid credentials), we should treat them as successful deregistrations and clear local storage. Otherwise, the device may remain in a broken state. If device identity token for some reason got invalid, broken or revoked we end up in permanent broken state.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1141,8 +1141,10 @@ h3(#activation-state-machine). Activation State Machine
 * @(RSH3)@ In platforms that support receiving push notifications, in order to connect the device's push features with Ably's, the library must perform the process described in the following abstract state machine. While this process should be implemented in whatever way better fits the concrete platform, it should be taken into account that its lifetime is that of the _app_ that runs it, which outlives that of the @RestClient@ instance or (typically) the process running the app. This typically forces some kind of on-disk storage to which the state machine's state must be persisted, so that it can be recovered later by new instances and processes running the app triggered by external events.
 ** @(RSH3a)@ State @NotActivated@ (the initial one).
 *** @(RSH3a1)@ On event @CalledDeactivate@:
-**** @(RSH3a1a)@ Makes @Push#deactivate@ return or call its callback with no error.
-**** @(RSH3a1b)@ Transitions to @NotActivated@.
+**** @(RSH3a1a)@ This clause has been deleted. It was valid up to and including specification version @3.0.0@.
+**** @(RSH3a1b)@ This clause has been deleted. It was valid up to and including specification version @3.0.0@.
+**** @(RSH3a1c)@ If the local device has @deviceIdentityToken@, does the same as @(RSH3d2)@.
+**** @(RSH3a1d)@ Otherwise, does the same as @(RSH3g2)@.
 *** @(RSH3a2)@ On event @CalledActivate@:
 **** @(RSH3a2a)@ If the local device has @deviceIdentityToken@, performs a validation of the local DeviceDetails via the following steps. "@(RSH3a2b)@":#RSH3a2b onwards then don't apply.
 ***** @(RSH3a2a1)@ Checks the compatibilty of the present client with the existing registration: if the @LocalDevice@ has a non-empty @clientId@, and the present identified client has a different (non-null) @clientId@, then a @SyncRegistrationFailed@ event should be fired containing an error with @code@ 61002, and skips to "@(RSH3a2a4)@":#RSH3a2a4.
@@ -1185,8 +1187,9 @@ h3(#activation-state-machine). Activation State Machine
 **** @(RSH3d1b)@ Transitions to @WaitingForNewPushDeviceDetails@.
 *** @(RSH3d2)@ On event @CalledDeactivate@:
 **** @(RSH3d2a)@ If a custom @deregisterCallback@ was provided to @Push#deactivate@, pass it the local @DeviceDetails@ 's id.
-**** @(RSH3d2b)@ Otherwise, make an asynchronous DELETE HTTP request to "/push/deviceRegistrations":/rest-api/#delete-device-registration using the local @DeviceDetails@ 's ID. This operation requires "push device authentication":#push-device-authentication.
+**** @(RSH3d2b)@ Otherwise, make an asynchronous DELETE HTTP request to "/push/deviceRegistrations":/rest-api/#delete-device-registration using the local @DeviceDetails@ 's ID. This operation requires "push device authentication":#push-device-authentication without other token or key authentication.
 **** @(RSH3d2c)@ Either way, when the registration is done, a @Deregistered@ or @DeregistrationFailed@ event should be fired.
+**** @(RSH3d2c1)@ @Deregistered@ event should be fired if the HTTP request returns a status code in the 2xx range, 401 (unauthorized), or code 40005 (invalid credentials). Otherwise @DeregistrationFailed@ event should be fired.
 **** @(RSH3d2d)@ Transitions to @WaitingForDeregistration@.
 *** @(RSH3d3)@ On event @GotPushDeviceDetails@ (note that this will only happen on platforms whose push device details, after first set, can change, e. g. FCM's registration token refresh):
 **** @(RSH3d3a)@ If a custom @registerCallback@ was provided to @Push#activate@, pass it the local @DeviceDetails@ updated with the push details.


textile/features.textile
```diff
@@ -1141,7 +1141,8 @@ h3(#activation-state-machine). Activation State Machine
 * @(RSH3)@ In platforms that support receiving push notifications, in order to connect the device's push features with Ably's, the library must perform the process described in the following abstract state machine. While this process should be implemented in whatever way better fits the concrete platform, it should be taken into account that its lifetime is that of the _app_ that runs it, which outlives that of the @RestClient@ instance or (typically) the process running the app. This typically forces some kind of on-disk storage to which the state machine's state must be persisted, so that it can be recovered later by new instances and processes running the app triggered by external events.
 ** @(RSH3a)@ State @NotActivated@ (the initial one).
 *** @(RSH3a1)@ On event @CalledDeactivate@:
-**** @(RSH3a1a)@ Makes @Push#deactivate@ return or call its callback with no error.
```
maratal, line 1144: We don't just delete items. Instead you should mark it "This spec was valid up until version...".

maratal: See https://github.com/ably/specification/blob/main/CONTRIBUTING.md#features-spec-points

textile/features.textile
```diff
@@ -1185,8 +1186,9 @@ h3(#activation-state-machine). Activation State Machine
 **** @(RSH3d1b)@ Transitions to @WaitingForNewPushDeviceDetails@.
 *** @(RSH3d2)@ On event @CalledDeactivate@:
 **** @(RSH3d2a)@ If a custom @deregisterCallback@ was provided to @Push#deactivate@, pass it the local @DeviceDetails@ 's id.
-**** @(RSH3d2b)@ Otherwise, make an asynchronous DELETE HTTP request to "/push/deviceRegistrations":/rest-api/#delete-device-registration using the local @DeviceDetails@ 's ID. This operation requires "push device authentication":#push-device-authentication.
+**** @(RSH3d2b)@ Otherwise, make an asynchronous DELETE HTTP request to "/push/deviceRegistrations":/rest-api/#delete-device-registration using the local @DeviceDetails@ 's ID. This operation requires "push device authentication":#push-device-authentication without other token or key authentication.
```
maratal, line 1189: Updates also has a rule, but I'm not sure if it worth it in this case. Looks like minor addition.

textile/features.textile
```diff
@@ -1185,8 +1186,9 @@ h3(#activation-state-machine). Activation State Machine
...
 **** @(RSH3d2c)@ Either way, when the registration is done, a @Deregistered@ or @DeregistrationFailed@ event should be fired.
+**** @(RSH3d2c1a)@ @Deregistered@ event should be fired if the HTTP request returns a status code in the 2xx range, 401 (unauthorized), or code 40005 (invalid credentials). Otherwise @DeregistrationFailed@ event should be fired.
```
maratal, line 1191: Why RSH3d2c1a and not just RSH3d2c1?

maratal (changes requested):
I've made a pre-review not digging into the purpose of these changes.

textile/features.textile
```diff
@@ -1185,8 +1187,9 @@ h3(#activation-state-machine). Activation State Machine
 **** @(RSH3d1b)@ Transitions to @WaitingForNewPushDeviceDetails@.
 *** @(RSH3d2)@ On event @CalledDeactivate@:
 **** @(RSH3d2a)@ If a custom @deregisterCallback@ was provided to @Push#deactivate@, pass it the local @DeviceDetails@ 's id.
-**** @(RSH3d2b)@ Otherwise, make an asynchronous DELETE HTTP request to "/push/deviceRegistrations":/rest-api/#delete-device-registration using the local @DeviceDetails@ 's ID. This operation requires "push device authentication":#push-device-authentication.
+**** @(RSH3d2b)@ Otherwise, make an asynchronous DELETE HTTP request to "/push/deviceRegistrations":/rest-api/#delete-device-registration using the local @DeviceDetails@ 's ID. This operation requires "push device authentication":#push-device-authentication without other token or key authentication.
 **** @(RSH3d2c)@ Either way, when the registration is done, a @Deregistered@ or @DeregistrationFailed@ event should be fired.
+**** @(RSH3d2c1)@ @Deregistered@ event should be fired if the HTTP request returns a status code in the 2xx range, 401 (unauthorized), or code 40005 (invalid credentials). Otherwise @DeregistrationFailed@ event should be fired.
```
owenpearson, line 1192: this feels wrong to me, haven't fully thought about the changes but doesn't this mean that if a client temporarily has invalid auth credentials then it would appear as deregistered even if the device registration still exists?

ttypic: Deregistration call should relying solely on device identity token authentication (without any key or temporary token auth), as I highlighted in `RSH3d2b`. Device identity token is permanent and should never change after push registration. If it becomes invalid, it's impossible to update. We can narrow down to `40100` error code, we had race-condition problem before and created a lot of states where `deviceIdentityToken` is not align with `deviceId` so every time we are trying to deregister we get `"Token deviceId does not match"`, and the only solution right now is to reinstall the app. 

But I have feeling that it is much better to handle any given problem that can happen to `deviceIdentityToken`. So I added  `Invalid credentials` that you can get if your token is not well-formed or `Token unrecognised` that you can get if you provide key / token auth from another Ably app. This way we can be sure that even if there are some problem in SDK which resulted in broken state, we can deregister. Otherwise end-user have to reinstall the app.

paddybyers:
> * We ended up in `NotActive` with valid device identity token (e.g. haven't received registration update)

If we're `NotActive`, shouldn't we delete all remnants of aborted registration attempts?

ttypic:
> > * We ended up in `NotActive` with valid device identity token (e.g. haven't received registration update)
> 
> If we're `NotActive`, shouldn't we delete all remnants of aborted registration attempts?

@paddybyers no, right now spec says that on `NotActive` you can have old registration / token update attempts (we clear all remnants  only when we get `Derigistered` event during on `WaitForDeregistration` state). That's why we try to restore the state on `CalledActivate` event, but we are ignoring `CalledDeactivate` and that's the problem.

maratal:
I can't find the code path leading to `NotActivated` while having saved token details. In `ably-cocoa` you'll end up in `ARTPushActivationStateNotActivated` in these four cases:
- `WaitingForDeviceRegistration` and failed (you don't have saved token yet)
- `WaitingForPushDeviceDetails` and failed (you don't have saved token yet)
- `WaitingForPushDeviceDetails` and called `deactivate` (you don't have saved token yet)
- `WaitingForDeregistration` and device is reset before going to `NotActivated`.

Should be similar on Android I guess? @ttypic

ttypic:
@maratal 

You can be in `WaitingForRegistrationSync` or `WaitingForDeregistration` states and haven't received next event (network issues, you closed the app, you closed ably client, etc). After you reopen the app you'll have `deviceIdentityToken` and `NotActive` state.

@paddybyers 

> So stale/invalid deviceId etc can exist in the NotActive state, and would only get deleted if there was an explicit request to deregister?

- It's very important to understand that the `NotActive` state can have valid `deviceId`, `deviceIdentityToken`. And push notifications can work fine. 
- Sdk user can't explicit request to deregister. They can only send `CalledDeactivate`, `CalledActivate` events by calling `ably.push.activate()`/`ably.push.deactivate()`. 
- Right now we are ignoring `CalledDeactivate` on `NotActive` state.
- If user logged out, then logged in with different `clientId`, called `deactivate()` and then `activate()` they will receive clientId mismatch error. (and still be subscribed to push notifications from another account).

> Do you think deleting unused device details should happen on any transition to `NotActive`?

I don't think so, because `NotActive` can have valid device details and we can simply delete it from local storage.

maratal:
>You can be in WaitingForRegistrationSync or WaitingForDeregistration

But in these two cases the previous state is not `NotActivated`, right? (at least in ably-cocoa). @ttypic

ttypic:
@maratal Yes, but the previous state doesn't matter. We were discussing path leading to `NotActivated` while having saved `deviceIdentityToken`. And these are the paths, you can look at  https://sdk.ably.com/builds/ably/specification/main/features/#RSH3a2a - this spec point specifically mention this exact situation.

maratal (approved):

textile/features.textile
umair-ably: @owenpearson - Evegenii is off for this week. Are we happy to resolve this, or is there anything I can do to help get this in?

owenpearson: @umair-ably, I'm still not sure I'm convinced, sorry. My question was more about what would happen if the client had invalid API auth credentials (ably key/token) rather than a problem with its local `deviceIdentityToken`. I haven't tested it, but from what I can understand, this change would mean that a device without valid API credentials would appear as if it has successfully deregistered even though its attempt to remove its own device registration is rejected by the API with a 401, which would be wrong since the registration would still exist in that case.

umair-ably: Ahh I see what you mean. If my understanding is correct then something like this would play out... 

1. User successfully registers with valid auth 
2. User hits invalid auth error for whatever reason (ably key or token)
4. Server returns 401. The server does no processing itself since invalid auth was used.  
5. Device clears local registration state, allowing user to attempt registration again.
6. Server still assumes registered state

Seems like the risk here is that the server and client will be out of sync e.g. server assumes registered, client assumes deregistered. In this case, I think the changes to `RSH3d2b` in this commit would address that with something like this:

1. User successfully registers with valid auth 
2. User calls deregister with only device identity token (not the usual ably key or token)
4. Server returns 2xx if there is a matching and now removed device identify token, or a 40x (perhaps a 404 or 204 imo) if it could not find the token to begin with.
5. On both success or failure from step 4, the device clears local registration state, allowing the user to attempt registering again. (this is the critical step to fix the race conditions described by Evgenii above)
6. Server and client are in sync 

Have I missed something here?

owenpearson: Yep, your understanding of the problem sounds correct 🙂 as for the solution, I'm not sure if you can send a `DELETE /push/deviceRegistrations` with just the deviceIdentityToken for authentication. I haven't tested it but I'd be quite surprised if that was the case. Moreover, if a client somehow ended up with an invalid deviceIdentityToken then you would likely end up with the same issue of client/server getting out of sync, although i guess this is quite an edge case since that would only really be caused by an sdk bug or a user sabotaging their own deviceIdentityToken.

I do think deleting local state upon a 404 makes sense, since in that case the server is actually saying that the device registration doesn't exist. Maybe if we're currently sending 401s on the basis that the deviceIdentityToken is no longer valid then we should change the API to respond with a 404 in that case?

owenpearson:
Having thought about this PR a bit more broadly, I do agree with @paddybyers that having any device push state while in the `NotActive` state is wrong, and that this PR addresses that by allowing that state to be deleted rather than preventing an invalid state in the first place. Although, I suppose even if we made a change to prevent the machine from having state while `NotActive`, there would still be clients which got into that invalid state from an older SDK version, so we do at least need something in place to exit that invalid state, I'm not really sure what the best way to do that would be though.

textile/features.textile
umair-ably: Evegenii put in a PR for ably-java before going on leave which also implies `RSH3d2b` works by calling delete solely with the deviceIdentityToken [here](https://github.com/ably/ably-java/pull/994). 

The deviceIdentityToken should never change once a user registers for push. Unless, as you said, a real edge case occurs in which we should delete it locally or else the user is stuck with it. 

> Maybe if we're currently sending 401s on the basis that the deviceIdentityToken is no longer valid then we should change the API to respond with a 404 in that case?

I don't think we are sending 401s right now on the basis they are no longer valid... it is purely down to invalid auth. The changes to `RSH3d2b` would allow us to do this though. 

I agree with yourself and Paddy that the wider design for this should mean `NOT ACTIVE` also means not persisting device details. There have been a few calls to simplify this state machine 😅 

I guess the context for why this solution was the one proposed was that it is currently the cause of several thousand broken states for end clients a day, which in turn means users are not receiving push notifications when they should be. I think this solution is adequate if my understanding of `RSH3d2b` is correct. Without `RSH3d2b`, there is a risk users think they've deregistered but they actually haven't (causing the inverse of current issues), however subsequent calls with valid auth would fix this (unlike now where this would still not resolve our issues)

owenpearson: Just discussed this with @ttypic on a call, I had missed the fact that this request would only be sent with a deviceIdentityToken (ie excluding api credentials) so I'll resolve this thread now.

owenpearson (approved):

PR #191, opened 2024-05-19
Added forgotten test case for RTL2f

### LINKED ISSUE #190: RTL2f spec is vauge on what test for it should do
It [says](https://sdk.ably.com/builds/ably/specification/main/features/#RTL2f) "_A test should exist to ensure that resumed is always false when a channel first becomes ATTACHED, it is true when the channel is ATTACHED following a successful [connection recovery](https://sdk.ably.com/builds/ably/specification/main/features/#RTN16), and is false when the channel is ATTACHED following a failed [connection recovery](https://sdk.ably.com/builds/ably/specification/main/features/#RTN16)_", but connection recovery is not the only case where `resumed` flag should be tested (and equal to `true`).

maratal:
Closes #190 
This shouldn't alter the behavior of the SDK, so I'm editing this in place.


ttypic (commented):
I don't really like that we are mentioning test cases in the spec point, but changes look valid. What do you think @SimonWoolf?

SimonWoolf:
I agree with Evgeny on both counts

I've [never liked](https://github.com/ably/specification/pull/166#discussion_r1320080062) the "A test should exist..." spec items. IMO the spec should specify behaviour, and it should be up to the implementer to write whatever tests they need to be sure that the behaviour is correct. (And in any case, the spec isn't exhaustive about what tests to write). But if we are going to have that in the spec, no particular objection to this one

ttypic (approved):

PR #192, opened 2024-06-16
Replaced RTL7c with RTL7g

maratal:
Channel should be attached whenever subscribe is called (except for the `FAILED` channels).


textile/features.textile
```diff
@@ -675,7 +675,8 @@ h3(#realtime-channel). RealtimeChannel
 * @(RTL7)@ @RealtimeChannel#subscribe@ function:
 ** @(RTL7a)@ Subscribe with a single listener argument subscribes a listener to all messages
 ** @(RTL7b)@ Subscribe with a name argument and a listener argument subscribes a listener to only messages whose @name@ member matches the string name
-** @(RTL7c)@ Implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@ state. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
+** @(RTL7c)@ This clause has been deleted. It was valid up to and including specification version @3.0@.
```
ttypic, line 678: I would treat this change as clarification, since we implemented this spec point in ably-js differently, and do not treat this as spec update

textile/features.textile
```diff
@@ -675,7 +675,8 @@ h3(#realtime-channel). RealtimeChannel
...
+** @(RTL7g)@ Implicitly attaches the @RealtimeChannel@ if the channel is not in the @FAILED@ state. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
```
SimonWoolf, line 679: suggest `...if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states`. (It doesn't make sense to require an implicit attach if the channel is already in the `attached` state, or is already in the process of attaching, or is suspended).

SimonWoolf (changes requested):

textile/features.textile
maratal: https://github.com/ably/specification/pull/192/commits/454a765188cf8c12a0f7939ec2c383f9cd130181

textile/features.textile
maratal: https://github.com/ably/specification/pull/192/commits/454a765188cf8c12a0f7939ec2c383f9cd130181

SimonWoolf (approved):

PR #194, opened 2024-06-22
TO3g editing in place for clarity.

maratal:
I want to make it more clear that the spec is about user data and presence messages (`MESSAGE` and `PRESENCE`).

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1606,7 +1606,7 @@ h4. ClientOptions
 ** @(TO3d)@ @tls@ boolean - defaults to true. If false, will not use TLS for all connections
 ** @(TO3e)@ @autoConnect@ boolean - defaults to true. If false, suppresses the automatic initiation of a connection when the library is instantiated
 ** @(TO3f)@ @useBinaryProtocol@ boolean - defaults to true. If false, forces the library to use the JSON encoding for REST and Realtime operations, instead of the default binary msgpack encoding
-** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of @MESSAGE@ or @PRESENCE@ @ProtocolMessage@s when connection is either in the @CONNECTING@ or @DISCONNECTED@ states. Otherwise those messages should be failed immediately if not in the @CONNECTED@ state.
+** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of @MESSAGE@ or @PRESENCE@ @ProtocolMessage@s See @RTL6c@ for connection and channel state condition details
 ** @(TO3h)@ @echoMessages@ boolean - defaults to true. If false, suppresses messages originating from this connection being echoed back on the same connection
 ** @(TO3i)@ @recover@ string - A connection recovery string, specified with the intention of inheriting the state of an earlier connection
 ** @(TO3n)@ @idempotentRestPublishing@ boolean - defaults to false for clients with version &lt; 1.2, otherwise true. If true, "RSL1k":#RSL1k1 applies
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -1606,7 +1606,7 @@ h4. ClientOptions
 ** @(TO3d)@ @tls@ boolean - defaults to true. If false, will not use TLS for all connections
 ** @(TO3e)@ @autoConnect@ boolean - defaults to true. If false, suppresses the automatic initiation of a connection when the library is instantiated
 ** @(TO3f)@ @useBinaryProtocol@ boolean - defaults to true. If false, forces the library to use the JSON encoding for REST and Realtime operations, instead of the default binary msgpack encoding
-** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of messages when connection states that anticipate imminent connection (connecting and disconnected). Instead, publish and presence state changes will fail immediately if not in the connected state
+** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of publish and presence messages when connection either in the `CONNECTING` or `DISCONNECTED` states. Otherwise messages should be failed immediately if not in the `CONNECTED` state.
```
sacOO7, line 1609:
```suggestion
** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of messages when connection is either in the `CONNECTING` or `DISCONNECTED` state. Otherwise they should be failed immediately if not in the `CONNECTED` state.
```
I think type of `messages` that needs to be queued is already explicitly mentioned at other places.

maratal: But the only place where they SHOULD be mentioned is this spec =) Although they've mentioned in the end, but making it clear from the beginning makes sense I guess.

maratal: No no, now it shows different text! Please get back mentioning type of messages.

maratal: https://github.com/ably/specification/commit/ef10bc10db4f55191c8914889a6e1a1fe69073cc

textile/features.textile
```diff
@@ -1606,7 +1606,7 @@ h4. ClientOptions
 ** @(TO3d)@ @tls@ boolean - defaults to true. If false, will not use TLS for all connections
 ** @(TO3e)@ @autoConnect@ boolean - defaults to true. If false, suppresses the automatic initiation of a connection when the library is instantiated
 ** @(TO3f)@ @useBinaryProtocol@ boolean - defaults to true. If false, forces the library to use the JSON encoding for REST and Realtime operations, instead of the default binary msgpack encoding
-** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of messages when connection states that anticipate imminent connection (connecting and disconnected). Instead, publish and presence state changes will fail immediately if not in the connected state
+** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of @MESSAGE@ or @PRESENCE@ messages when connection is either in the `CONNECTING` or `DISCONNECTED` states. Otherwise those messages should be failed immediately if not in the `CONNECTED` state.
```
ttypic, line 1609: `@MESSAGE@ or @PRESENCE@ messages` looks a bit odd, I think it's better `messages or presence messages` or `@Message@ or @PresenceMessage@`. Also I believe "`" should be replaced with "@".

```suggestion
** @(TO3g)@ @queueMessages@ boolean - defaults to true. If false, suppresses the default queueing of messages or presence messages when connection is either in the @CONNECTING@ or @DISCONNECTED@ states. Otherwise those messages should be failed immediately if not in the @CONNECTED@ state.
```

textile/features.textile
maratal: https://github.com/ably/specification/pull/194/commits/b1a46dd8360aa2578ec3b50aab7f6e079e178472

textile/features.textile
maratal: a015fe14

maratal: Replaced @ProtocolMessages@ with @ProtocolMessage@s

textile/features.textile
maratal: See this [thread](https://github.com/ably/specification/pull/194#discussion_r1651042915)

ttypic (approved):
LGTM

textile/features.textile
SimonWoolf: That thread addresses the first of the three points I made, other two are still unaddressed

maratal: https://github.com/ably/specification/pull/194/commits/1b4c8b2ffc48a294a1a9dfb34b5f27073f2efe16 @SimonWoolf

SimonWoolf (approved):

PR #195, opened 2024-07-05
RSA7f: clarify that send clientId regardless of basic/token auth

SimonWoolf:
Previous RSA7e said to only send clientId in the querystring & x-ably-clientid header in rest reqs if using basic auth. Which is just wrong, and not what any sdk does. (ably-js sends it unconditionally, ably-ruby sends it if token auth is used and it can see that the token has a wildcard client -- which is still wrong because the sdk might not know what the token clientId is if it's fed an opaque token string).

internal discussion: https://ably-real-time.slack.com/archives/C8SPU4589/p1720167508719079


textile/features.textile
```diff
@@ -203,7 +203,7 @@ h3(#rest-auth). Auth
 * @(RSA6)@ The @capability@ for new tokens is JSON stringified. If If the user-provided @tokenParams@ does not specify capabilities, the @capability@ field should be omitted from the @tokenRequest@, and Ably will supply a token with the capabilities of the underlying key. See "TK2b":#TK2b
 * @(RSA7)@ @clientId@ and authenticated clients:
 ** @(RSA7d)@ If @clientId@ is provided in @ClientOptions@ and @RSA4@ indicates that token auth is to be used, the @clientId@ field in the @TokenParams@ (@TK2c@) should be set to that @clientId@ when requesting a token
-** @(RSA7e)@ If @clientId@ is provided in @ClientOptions@ and @RSA4@ indicates that basic auth is to be used, then:
+** @(RSA7e)@ If a valid @clientId@ is provided in @ClientOptions@, then:
```
sacOO7, line 206: Wdyt about
```suggestion
** @(RSA7e)@ If a non-wildcard @clientId@ is provided in @ClientOptions@, then:
```

SimonWoolf:
I originally had this as adding a new RSA7f and deprecating RSA7e for <=3.0.0. But I've changed my mind, now just editing RSA7e in-place. Having RSA7e claim to be valid for <=3.0.0 ignores the reality that no SDK actually implemented RSA7e as written, so may as well change it to just reflect what they actually do

textile/features.textile
SimonWoolf: I don't like that for DRY reasons. RSA7c defines what a valid clientId is, if we add additional restrictions that should be the one place we need to edit.

sacOO7: Okay, makes sense.

sacOO7 (approved):
lgtm

PR #204, opened 2024-08-21
Rtn15h2 and connection states

SimonWoolf:
Per [slack thread](https://ably-real-time.slack.com/archives/C8SPU4589/p1724257493225679?thread_ts=1724236742.837859&cid=C8SPU4589), it was not sufficiently clear that the library should be in the CONNECTING state when it's attempting to reconnect. In particular the lib must not be in the CONNECTED state when not actually connected to the server, which is not currently made explicit. Clarified this spec item and added brief descriptions of all the connection states.

(I allowed the library to optionally go through DISCONNECTED on its way to CONNECTING in the case of an immediate reconnect since there are SDKs which do that and I don't want to make them suddenly be not in compliance, it's no big deal to do that if they want)

Also cleaned up RTN15a (which previously waffled on a bit about various things not relevant to actions to take if the transport is disconnected).


ttypic (approved):
LGTM

sacOO7 (approved):
lgtm

textile/features.textile
```diff
@@ -506,9 +515,11 @@ h3(#realtime-connection). Connection
...
 ** @(RTN15i)@ If an @ERROR@ @ProtocolMessage@ is received, this indicates a fatal error in the connection. The server will close the transport immediately after. The client should transition to the @FAILED@ state triggering all attached channels to transition to the @FAILED@ state as well. Additionally the @Connection#errorReason@ should be set with the error received from Ably
-** @(RTN15a)@ If a @Connection@ transport is disconnected unexpectedly or because a token has expired, then the @Connection@ manager will immediately attempt to reconnect and restore the connection state. Connection state recovery is provided by the Ably service and ensures that whilst the client is disconnected, all events are queued and channel state is retained on the Ably servers. When a new connection is made with the correct connection recovery key, the client is able to catch up by receiving the queued @ProtocolMessages@ from Ably.
+** @(RTN15a)@ If a transport is disconnected unexpectedly (without having received a @DISCONNECTED@ or @ERROR@ protocol message), it should respond as if it had received a non-token @DISCONNECTED@ (following @RTN15h3@).
```
sacOO7, line 522: This spec was correctly implemented in ably-go 4 years back -> https://github.com/ably/ably-go/pull/152/files

PR #205, opened 2024-08-30
Spec: ban warning on unknown fields for spec v3

SimonWoolf:
There's apparently an sdk that prints a warning message when it decodes a Message with an unknown field. This bans that for v3, for forward compatibility.

(I don't know which one, @mschristensen I think you mentioned it, do you know?)

I was thinking if I might have to bump the major version for this change, but looking at various SDKs, the only SDK so far to have implemented wire protocol v3 (and therefore spec v3) is ably-js, which doesn't complain about unknown fields, so already satisfied this. So I think I'm safe to add this as a requirement for v3. @ttypic wdyt?

(the purpose is to be able to use protocol v3 to trigger possible inclusion of new fields that are being added for various purposes like message interactions and data sync)


mschristensen (commented):
> (I don't know which one, @mschristensen I think you mentioned it, do you know?)

@SimonWoolf I think Paddy mentioned it, and I think he said Ably Java

lawrence-forooghian:
Does [RSF1](https://sdk.ably.com/builds/ably/specification/main/features/#RSF1) not already cover this?

SimonWoolf:
> Does [RSF1](https://sdk.ably.com/builds/ably/specification/main/features/#RSF1) not already cover this?

Oh, yes, good point, I missed that.
So if ably-java is logging a warning on unknown fields, it's in breach of that spec item already? Can we get that fixed?

lawrence-forooghian:
Yep, I’d say it's an ably-java bug in that case. Do you have an example of the error message? I’ll create an ably-java bug.

SimonWoolf:
> Do you have an example of the error message? I’ll create an ably-java bug.

I don't, I have this second hand (Mike commented above that Paddy mentioned it).
I've just tried to reprod but apparently you can no longer run the ably-java tests without having an android development environment installed for some reason, which I don't. Would you mind having a go?

lawrence-forooghian:
Yep, I’ll take a look. Does sandbox contain the realtime changes that trigger the warning?

lmars:
> you can no longer run the ably-java tests without having an android development environment installed for some reason

quick hack: delete this line: https://github.com/ably/ably-java/blob/e0e23176c717406f0c9a0e96752b6dc8d6663e47/settings.gradle#L3

paddybyers:
> I have this second hand (Mike commented above that Paddy mentioned it).

See for example https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/types/ProtocolMessage.java#L226

SimonWoolf:
> See for example https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/types/ProtocolMessage.java#L226

That's a `VERBOSE` log, not a warn or error, so wouldn't even be a violation of this new spec item

lawrence-forooghian:
> That's a `VERBOSE` log, not a warn or error, so wouldn't even be a violation of this new spec item

But it does violate the existing RSF1, given that it doesn't "ignore" it per that spec item's language. So it still sounds like an ably-java bug.

PR #206, opened 2024-09-06
Add opt-out of implicit attach when subscribing

### LINKED ISSUE #202: Add a way to subscribe to realtime channel and presence messages without triggering an attach
## Background

This arose in the context of https://github.com/ably-labs/ably-chat-swift/issues/18.

The JS Chat SDK has decided that in its API, subscribing to a room’s messages or presence messages should _not_ trigger a channel attach operation (see [decision record](https://ably.atlassian.net/wiki/spaces/CHA/pages/3156705308/CHADR-038+API+Design+Separating+Listeners+from+Lifecycle)). In order to do this, it uses ably-js private API in order to, essentially, call `RealtimeChannel#subscribe` and `RealtimePresence#subscribe` in a way that does not trigger the RTL7c or RTP6c implicit attach.

For the Swift and Kotlin Chat SDKs, I would like us to instead be able to use an API that’s properly specified in this spec, even if marked as "experimental and for Ably-authored SDKs" only.

## Suggested approach

Add some sort of `subscribeWithoutImplicitAttach` method to `RealtimeChannel` and `RealtimePresence`. Think of a way of not confusing users with this method.

lawrence-forooghian:
Add an `attachOnSubscribe` channel option, which, if set to false, allows users to opt out of the implicit attach that’s triggered by `subscribe`-ing to channel or presence messages.

The Chat SDK has decided that in its API, subscribing to a room’s messages or presence messages should not trigger a channel attach operation (see [decision record](https://ably.atlassian.net/wiki/spaces/CHA/pages/3156705308/CHADR-038+API+Design+Separating+Listeners+from+Lifecycle)). The JS Chat SDK uses private ably-js API to, essentially, call `subscribe` without triggering an implicit attach. For the Swift and Kotlin SDKs, we’d like to avoid this, hence this new API.

I’m not sure what’s the better approach out of adding a new channel option (I don’t love the fact that we just ignore the user-provided “on attach” callback) or adding a new method called something like `subscribeWithoutAttach` (which, given that it’d be part of the public API, might be quite confusing for users when it appears in their IDE autocomplete or whatever).

Resolves #202.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -686,7 +686,9 @@ h3(#realtime-channel). RealtimeChannel
 * @(RTL7)@ @RealtimeChannel#subscribe@ function:
 ** @(RTL7a)@ Subscribe with a single listener argument subscribes a listener to all messages
 ** @(RTL7b)@ Subscribe with a name argument and a listener argument subscribes a listener to only messages whose @name@ member matches the string name
-** @(RTL7c)@ Implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
+** @(RTL7c)@ This clause has been replaced by "@RTL7g@":#RTL7g. It was valid up to and including specification version @3.0@.
+** @(RTL7g)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
+** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, the optional callback must not be called.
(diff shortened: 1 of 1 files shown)


SimonWoolf (approved):
fine by me

lawrence-forooghian:
> Okay noob question, why don't we just create plain channel when subscribe is called.

The aim of `subscribe` is to register a listener that will be called when the channel recieves a message. Your example fetches a channel but doesn't register a listener.

SimonWoolf:
> We should consider whether we want this to be default behaviour in future major SDK versions.

I would endorse this, never liked implicit attach. (especially never liked the asymmetry of 'implicit attach on first subscribe but no implicit detach on last unsubscribe')

AndyTWF:
> > We should consider whether we want this to be default behaviour in future major SDK versions.
> 
> I would endorse this, never liked implicit attach. (especially never liked the asymmetry of 'implicit attach on first subscribe but no implicit detach on last unsubscribe')

In Chat we took the approach of having `subscribe` having no bearing at all on the chat room lifecycle (in all cases). Any listener registration is purely synchronous. Similar for any other operation that would have a side effect (e.g. presence get), we either have it behind a promise that won't resolve until attach is initiated explicitly, or it simply fails and informs the user that they need to attach first.

ttypic (approved):
I like this approach

maratal:
>we just ignore the user-provided “on attach” callback

Maybe do attach when user provides this callback and do not do by default if not?

lawrence-forooghian:
> Maybe do attach when user provides this callback and do not do by default if not?

That'd be a breaking API change — the current expectation is that if you call `subscribe` then it attaches.

maratal (approved):
LGTM

PR #208, opened 2024-09-30
Fixed spec for channel subscribe with optional callback

sacOO7:
- Subscribe method accepts optional callback to indicate success/failure of a specified operation.
- Old spec keeps callback into pending state which is not acceptable and confusing at the same time.
- Updated spec to indicate success when specified operation is successful.
- It wasn't intentional, but same is implemented in ably-java -> 
https://github.com/ably/ably-java/blob/ea94d2a87fee3373a6495bfc11ebaba5394fa6c9/lib/src/main/java/io/ably/lib/realtime/Presence.java#L311-L314

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -687,8 +687,8 @@ h3(#realtime-channel). RealtimeChannel
 ** @(RTL7a)@ Subscribe with a single listener argument subscribes a listener to all messages
 ** @(RTL7b)@ Subscribe with a name argument and a listener argument subscribes a listener to only messages whose @name@ member matches the string name
 ** @(RTL7c)@ This clause has been replaced by "@RTL7g@":#RTL7g. It was valid up to and including specification version @3.0@.
-** @(RTL7g)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
-** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, the optional callback must not be called.
+** @(RTL7g)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result.
+** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, listener is registered without implicit attach and optional callback (if provided) is called indicating success.
 ** @(RTL7d)@ Messages delivered are automatically decoded based on the @encoding@ attribute; see @RestChannel@ encoding features. Tests should exist to publish and subscribe to encoded messages using the "AES 128":https://github.com/ably/ably-common/blob/main/test-resources/crypto-data-128.json and "AES 256":https://github.com/ably/ably-common/blob/main/test-resources/crypto-data-256.json fixture test data
 ** @(RTL7e)@ If a message cannot be decoded or decrypted successfully, it should be delivered to the listener with the @encoding@ attribute set indicating the residual encoding state, and an error should be logged
 ** @(RTL7f)@ A test should exist ensuring published messages are not echoed back to the subscriber when @echoMessages@ is set to false in the @RealtimeClient@ library constructor
@@ -781,8 +781,8 @@ h3(#realtime-presence). RealtimePresence
 ** @(RTP6a)@ Subscribe with a single listener argument subscribes a listener to all presence messages
 ** @(RTP6b)@ Subscribe with an action argument and a listener argument - such as @ENTER@, @LEAVE@, @UPDATE@ or @PRESENT@ - subscribes a listener to receive only presence messages with that action. In lanuages where method overloading is supported the action argument may also be an array of actions to receive only presence messages with an action included in the supplied array.
 ** @(RTP6c)@ This clause has been replaced by "@RTP6d@":#RTP6d. It was valid up to and including specification version @3.0@.
-** @(RTP6d)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
-** @(RTP6e)@ If the @attachOnSubscribe@ channel option is @false@, the optional callback must not be called.
+** @(RTP6d)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result.
+** @(RTP6e)@ If the @attachOnSubscribe@ channel option is @false@, listener is registered without implicit attach and optional callback (if provided) is called indicating success.
 * @(RTP7)@ @RealtimePresence#unsubscribe@ function:
 ** @(RTP7c)@ Unsubscribe with no arguments unsubscribes all listeners
 ** @(RTP7a)@ Unsubscribe with a single listener argument unsubscribes the listener if previously subscribed with an action-specific subscription


lawrence-forooghian (commented):
@sacOO7 you've described this as a "fix" but what is it fixing? Please could you explain why you believe the new approach to be correct?

textile/features.textile
```diff
@@ -687,8 +687,8 @@ h3(#realtime-channel). RealtimeChannel
...
 ** @(RTL7c)@ This clause has been replaced by "@RTL7g@":#RTL7g. It was valid up to and including specification version @3.0@.
-** @(RTL7g)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
-** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, the optional callback must not be called.
+** @(RTL7g)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result.
+** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, listener is registered without implicit attach and optional callback (if provided) is called indicating success.
```
lawrence-forooghian, line 691: "success" of what?

sacOO7: We can also update statement to accommodate technical implementation. 
Point is, not to keep `optional callback` into dangling state, especially when we explicitly mention not to `implicit attach` on subscribe.

lawrence-forooghian (commented):
I don't really have strong opinions on whether this is the right thing to do or not. I agree that leaving callbacks in a pending state isn't great, so it does kind of seem sensible. But on the other, it makes the meaning of the callback a bit muddier; before it meant "the channel has attached", and in some languages (e.g. ably-cocoa) this meaning is [embedded in the method name](https://github.com/ably/ably-cocoa/blob/80f05fd4023b7e502797ba376f4be0efbd127fc5/Source/include/Ably/ARTRealtimeChannel.h#L84) e.g. <code>subscribeWith<strong>AttachCallback</strong>:callback:</code> so to remove its meaning of "the channel has attached" is kind of confusing.

sacOO7:
@coderabbitai review

sacOO7:
@coderabbitai PR summary posted is totally wrong. We only rephrased two statements. Can you try to rewrite relevant summary if possible.

textile/features.textile
VeskeR: I like the behavior descripbed by Andy in this [comment](https://github.com/ably/specification/pull/206#issuecomment-2341350082) from the PR that introduced this logic:
> Any listener registration is purely synchronous. Similar for any other operation that would have a side effect ..., we either have it behind a promise that won't resolve **until attach is initiated explicitly**, or it simply fails and informs the user that they need to attach first.

Now, we can't remove the optional callback entirely and make subscription purely synchronous, however, we already had the line in the spec:
> The listener will always be registered regardless of the implicit attach result

which technically means that regarding registering a listener we don't care whether or not implicit attach is turned on and if we actually succeed in attaching.

So for the optional callback part (stealing the idea from chat's implementation), if `attachOnSubscribe` = `false` the callback will be called eventually according to [RTL4d](https://sdk.ably.com/builds/ably/specification/main/features/#RTL4d) after user's initiated explicit attach, or it's resolved immediatelly with no argument if it's already `ATTACHED`.

wdyt?

VeskeR (changes requested):

sacOO7:
@VeskeR I don't think we care about `ATTACHED` state here. `attachOnSunscribe` indicates whether we should care about implicit channel ATTACH result or not.  In case we disable it, we don't care about ATTACHED/ INITIALIZED state, we only care about listener getting registered. That's where we are deciding whether to call optional callback or not.

sacOO7:
@lawrence-forooghian In case of ably-go, every call is synchronous, we don't have exceptions and method call returns both result, error in the result. We wait for channel attach normally -> https://github.com/ably/ably-go/blob/a81632dfeb26bd545232674622463e5709d42e03/ably/realtime_channel.go#L515. But, when we disable implicit ATTACH, it shouldn't wait and return nil error. 
Nil error here indicates success of the operation.

textile/features.textile
maratal: >whether to call optional callback or not

But RTL7a says that this callback is not a `subscribe` callback, it's explicitly a callback for an attach operation. In cocoa docstrings the wording even more concrete (have no idea why it's different there): "_An **attach** callback may optionally be passed in to this call to be notified of success or failure of the channel `attach` operation._". And the parameter name is `onAttach`:

```
- (ARTEventListener *_Nullable)subscribeWithAttachCallback:(nullable ARTCallback)onAttach callback:(ARTMessageCallback)callback;
```

maratal: Said that I don't think that calling this callback as success makes sense. It's either should be renamed or called with an error.

sacOO7: I don't think `RTL7a` mentions callback to be ATTACH callback. Also, none of the public API across SDKs for `subscribe` denote any implicit ATTACH operation : (
In the future, we can have some other channel based operations other than ATTACH, that all would come under `subscribe`. Optional callback would only represent success/failure in such a case.

sacOO7: @maratal you might also like to check https://github.com/ably/specification/pull/208#issuecomment-2383765560

SimonWoolf: seems to me that the docs should advise that if `attachOnSubscribe` is set to `false`, the user should not pass a callback. There's no reason to do so. So we could either ban passing a callback in that case (throwing an exception), or alternatively we could silently unconditionally call it with no error. Don't mind which.

sacOO7: There's no reason to pass a callback, sure. But we are considering this because we know the nature of operation. It's a simple operation where listener is being added. What if there's some other complex async operation in the future which expects callback when attachOnSubscribe is false. I mean, callback in the end denotes success or failure, doesn't care about nature of operation.

SimonWoolf: >  What if there's some other complex async operation in the future which expects callback when attachOnSubscribe is false

I mean.. ok, but it's not like we pre-emptively make other synchronous operations async just in case some unknown future change makes them async. Conceptually, `subscribe()` with `attachOnSubscribe=false` is a local, synchronous operation. It takes a callback in callback languages just because subscribe with attachOnSubscribe=true is async, and we don't want to make them different methods. 

But anyway -- I'm not disagreeing with you. I think calling the callback immediately with no error is perfectly reasonable. (Eg in languages which return a promise, like ably-js v2, we certainly don't want to reject that promise, so if we want to keep a rough analogy with callback to promise apis, calling the callback with success seems fine).

lawrence-forooghian: My feeling is the following:

- adding a listener is fundamentally a synchronous operation and doesn't require a callback
- due to past decisions we are now in a position where the API of `subscribe` is able to asynchronously communicate the result of an attach operation
- where possible (i.e. in languages in which the `subscribe` method takes an optional attach callback), it should be considered a programmer error to pass an attach callback when `attachOnSubscribe == false` ("programmer error" to be handled per what’s idiomatic for that language; e.g. throw an exception)
- if the previous point is not possible (for example because the ability to asynchronously communicate the result of the operation is baked into the signature of the method, such as a method returning a promise in JS, or an `async` method in Swift), then we have to bite the bullet and indicate that the asynchronous work completed successfully (even though in the case where `attachOnSubscribe == false` the "asynchronous work" is a no-op, not an attach operation)

and make sure that this is documented correctly on a per-language basis.

Does that make sense?

maratal: >Also, none of the public API across SDKs for subscribe denote any implicit ATTACH operation

It's a "passed in to this call to be notified of success or failure of the channel `attach` operation" in a docstring.

Ok, calling it immediately without error seems reasonable to me too, but I would still log some warning into the console that "callback is called but no attach operation was performed due to options provided", at least in ably-cocoa where the name of the parameter is `onAttach`.

lawrence-forooghian: @sacOO7 @SimonWoolf @maratal do you have thoughts on what I proposed in https://github.com/ably/specification/pull/208#discussion_r1784424111?

maratal: >it's related to subscribe operation as a whole and not implicit attach.

It should be renamed everywhere to `attachCallback` to avoid confusion IMHO

>do you have thoughts on what I proposed

https://github.com/ably/specification/pull/208#discussion_r1784481377 @lawrence-forooghian

sacOO7: I think the problem lies only with cocoa where method name is subscribeWithAttach, in all other SDKs, there's just subscribe method that accepts a optional callback. User doesn't even care what type of implicit operation being performed internally.

lawrence-forooghian: > My opinion by looking at public API is, it's related to subscribe operation as a whole and not implicit attach.

> User doesn't even care what type of implicit operation being performed internally.

I disagree with this.

1. Our [public documentation for `#subscribe`](https://github.com/ably/sdk-api-reference/blob/af4c26721d7a76ed0d42887b3021dc8e9c3469f6/descriptions.md?plain=1#L309-L314) says 
   
   > A callback may optionally be passed in to this call to be notified of success or failure of the channel `attach()` operation.

   This means that users expect that, if the aforementioned optional callback is called without an error, the channel is now in an `ATTACHED` state, which might affect the way in which they use the channel.

2. Let's say that the optional callback did, as you say, represent the "subscribe operation as a whole and not implicit attach". What does it mean if this callback is called with an error? Does it mean that the listener was added or not, indicating that the "subscribe operation as a whole" failed? Does the user need to call `subscribe()` again for the listener to be added? (Answers: 1. it was added; 2. no, they don’t, but neither of these would be clear to the user.)

lawrence-forooghian: I’ll have a go at writing something.

lawrence-forooghian: I’ve opened #209, which I think captures the points raised in this PR.

sacOO7:
Closing in favor of #209

PR #209, opened 2024-10-07
Tighten up `attachOnSubscribe == false` behaviour

lawrence-forooghian:
This is a second attempt (resulting from the conversations from the first attempt in #208) at improving the behaviour (introduced in #202) of the `#subscribe` methods when the `attachOnSubscribe` channel option is `false`. See commit messages for more details.


textile/features.textile
```diff
@@ -688,7 +688,9 @@ h3(#realtime-channel). RealtimeChannel
...
 ** @(RTL7g)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
-** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, the optional callback must not be called.
+** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, then the behaviour depends on the public API that a given SDK uses to communicate the result of an "@RTL7g@":#RTL7g implicit attach:
+*** If the SDK’s API accepts an optional callback (or equivalent, for example it provides two overloads of @#subscribe@, one which accepts such a callback and one which doesn’t) to communicate the result of an "@RTL7g@":#RTL7g implicit attach, then it is a programmer error to provide such a callback when the @attachOnSubscribe@ channel option is @false@. This programmer error should be handled in an idiomatic fashion (e.g. by halting the program).
```
sacOO7, line 692: Do we want to throw exception for non optional callback here? Like when implicit attach is false

sacOO7: Like, I am not sure, but what if adding listener becomes async operation when called through wrapper SDKs.

lawrence-forooghian: We discussed this on a call and the above comments relate to believing that the optional callback communicates the result of the add-a-listener operation — after going through my explanation of why I believe this not to be the case (https://github.com/ably/specification/pull/208#discussion_r1790255796) we agreed that these comments can be resolved

lawrence-forooghian, line 692: Sachin indicated that he found the "accepts an optional callback" language a bit confusing; I’ll have a think about how to reword.

sacOO7: Yeah, now we have mutual understanding for the same and conversation can be closed.

textile/features.textile
```diff
@@ -688,7 +688,9 @@ h3(#realtime-channel). RealtimeChannel
...
+*** If the SDK’s API accepts a non-optional callback (or equivalent, for example it returns a @ChannelStateChange?@) to communicate the result of an "@RTL7g@":#RTL7g implicit attach, then when the @attachOnSubscribe@ channel option is @false@ this callback should be called with no argument.
```
sacOO7, line 693: Actually, `If the SDK’s API accepts a non-optional callback` feels confusing. 
Because, if we are applying this case for ably-go, `ably-go` doesn't accept any optional/non-optional callback for subscribe.

sacOO7: Rather, we can say, if result of subscribe operation is communicated as a async/sync blocking operation that includes result of implicit attach, then null error should be returned for the same.

sacOO7: Also, in case of `ably-go`, it returns `unsubscribe` method that removed listener instead of returning 
`ChannelStateChange`

lawrence-forooghian: I’ve removed the "accepts a non-optional callback" language and instead reworded it as these two cases:

- it accepts an optional callback
- anything else

textile/features.textile
```diff
@@ -688,7 +688,9 @@ h3(#realtime-channel). RealtimeChannel
...
 ** @(RTL7c)@ This clause has been replaced by "@RTL7g@":#RTL7g. It was valid up to and including specification version @3.0@.
 ** @(RTL7g)@ If the @attachOnSubscribe@ channel option is @true@, implicitly attaches the @RealtimeChannel@ if the channel is in the @INITIALIZED@, @DETACHING@, or @DETACHED@ states. The optional callback, if provided, is called according to "@RTL4d@":#RTL4d based on the implicit attach operation. The listener will always be registered regardless of the implicit attach result
-** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, the optional callback must not be called.
+** @(RTL7h)@ If the @attachOnSubscribe@ channel option is @false@, then the behaviour depends on the public API that a given SDK uses to communicate the result of an "@RTL7g@":#RTL7g implicit attach:
+*** If the SDK’s API accepts an optional callback to communicate the result of an "@RTL7g@":#RTL7g implicit attach, then it is a programmer error to provide such a callback when the @attachOnSubscribe@ channel option is @false@. This programmer error should be handled in an idiomatic fashion (e.g. by halting the program).
```
sacOO7, line 692:
```suggestion
*** If the SDK’s API accepts an optional callback to communicate the result of an "@RTL7g@":#RTL7g implicit attach, then it is a programmer error to provide such a callback when the @attachOnSubscribe@ channel option is @false@. This programmer error should be handled in an idiomatic fashion (e.g. throwing exception or returning error without callback being called).
```

sacOO7: by halting program do we mean exiting the application? like calling `exit(0)` or something

lawrence-forooghian: Yeah, that's roughly what Swift does — programmer errors are considered to be irrecoverable. That's why I said "in an idiomatic fashion".

How would, say, Java or Go want to handle this? Would they throw an `ErrorInfo`? If so, do you have any thoughts on what would be an appropriate `statusCode` and `code` to assign to this error, so that I can specify it here? (Deciding what codes to use is something that remains a bit of a mystery to me.)

sacOO7: I think throwing exception with common error code should be enough, as long as exception message convey the purpose

lawrence-forooghian: I’ve added this info.

sacOO7 (approved):
lgtm
finally : )

PR #210, opened 2024-10-10
Messages: Add new message fields to message spec

splindsay-92:
Updated `Message` class in the `textile/features.textile` file to support new message fields related to materialisation.

## Additions to the `Message` class:

* Added the `action` field to represent the type of the message, using the `MessageAction` enum. (
* Introduced the `serial` field to store the serial number of the message. 
* Added `refSerial` and `refType` fields for messages that reference other messages.
* Included `updatedAt` and `updateSerial` fields to track the update and deletion ordering. 
* Added the `operation` field, which is a object containing optional fields such as `clientId`, `description`, and `metadata`.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1293,8 +1293,8 @@ h2. Types
 h3(#types). Data types
 
 h4. Message
-
 * @(TM1)@ A @Message@ represents an individual message to be sent or received via the Ably Realtime service.
+* @(TM5)@ @Message@ @Action@ enum has the following values in order from zero: @MESSAGE_UNSET@, @MESSAGE_CREATE@, @MESSAGE_UPDATE@, @MESSAGE_DELETE@, @ANNOTATION_CREATE@, @ANNOTATION_DELETE@, @META_OCCUPANCY@
 * @(TM2)@ The attributes available in a @Message@ are:
 ** @(TM2a)@ @id@ string - unique ID for this message. This attribute is always populated for messages received over REST. For messages received over Realtime, if the message does not contain an @id@, it should be set to @protocolMsgId:index@, where @protocolMsgId@ is the id of the @ProtocolMessage@ encapsulating it, and @index@ is the index of the message inside the @messages@ array of the @ProtocolMessage@
 ** @(TM2b)@ @clientId@ string
@@ -1306,15 +1306,15 @@ h4. Message
 ** @(TM2i)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The @extras@ field is provided to contain message metadata and/or ancillary payloads in support of specific functionality, e.g. push. Each of these supported extensions is documented separately; for 1.1 the only supported extension is @push@, via the @extras.push@ member; 1.2 adds the @delta@ extension whose keys and values are described by the attributes of the type @DeltaExtras@, and the @headers@ extension, which contains arbitrary @string->string@ key-value pairs, settable at publish time, and @ref@ whose keys and values are described by the attributes of the type @ReferenceExtras@. Unless otherwise specified, the client library should not attempt to do any filtering or validation of the @extras@ field itself, but should treat it opaquely, encoding it and passing it to realtime unaltered.
 ** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
 ** @(TM2j)@ @action@ enum
-** @(TM2K)@ @serial@ string - an opaque string that uniquely identifies the message.
-** @(TM2L)@ @refSerial@ string - an opaque string that uniquely identifies some referenced message.
-** @(TM2M)@ @refType@ string - an opaque string that identifies the type of this reference.
-** @(TM2N)@ @updatedAt@ time in milliseconds since epoch. This field is always populated on messages received with an @action@ of @MESSAGE_UPDATE@.
-** @(TM2O)@ @deletedAt@ time in milliseconds since epoch. This field is always populated on messages received with an @action@ of @MESSAGE_DELETE@.
-** @(TM2P)@ @operation@ optional object that may contain the following `optional` attributes;
-*** @(TM2P1)@ @clientId@ string
-*** @(TM2P2)@ @description@ string
-*** @(TM2P3)@ @metadata@ object - used to contain any arbitrary key value pairs of type string.
+** @(TM2k)@ @serial@ string - an opaque string that uniquely identifies the message.
+** @(TM2l)@ @refSerial@ string - an opaque string that uniquely identifies some referenced message.
+** @(TM2m)@ @refType@ string - an opaque string that identifies the type of this reference.
+** @(TM2n)@ @operation@ - object that may contain the following `optional` attributes;
+*** @(TM2n1)@ @clientId@ string
+*** @(TM2n2)@ @description@ string
+*** @(TM2n3)@ @metadata@ object - used to contain any arbitrary key value pairs of type string.
+** @(TM2o)@ @updatedAt@ time in milliseconds since epoch at which an operation occurred. This field is always populated on messages received with an @action@ of @MESSAGE_UPDATE@ or @MESSAGE_DELETE@..
+** @(TM2p)@ @updateSerial@ string - an opaque string that uniquely identifies the operation. This field is always populated on messages received with an @action@ of @MESSAGE_UPDATE@ or @MESSAGE_DELETE@.
 * @(TM4)@ @Message@ has constructors @constructor(name: String?, data: Data?)@ and @constructor(name: String?, data: Data?, clientId: String?)@.
 * @(TM3)@ @fromEncoded@ and @fromEncodedArray@ are alternative constructors that take an (already deserialized) @Message@-like object (or array of such objects), and optionally a @channelOptions@, and return a @Message@ (or array of such @Messages@) that's decoded and decrypted as specified in @RSL6@, using the cipher in the @channelOptions@ if the message is encrypted, with any residual transforms (ones that the library cannot decode or decrypt) left in the @encoding@ property per @RSL6b@. This is intended for users receiving messages other than from a REST or Realtime channel (for example, from a queue), to avoid them having to parse the @encoding@ string themselves.
 
@@ -2149,8 +2149,6 @@ enum MessageAction: // TM5
     ANNOTATION_DELETE // TM5
     META_OCCUPANCY // TM5
 
-
-
 class ConnectionDetails: // CD*, internal
   clientId: String? // RSA12a, CD2a
   connectionKey: String // RTN15e, CD2b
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -2152,6 +2168,13 @@ class Message: // TM*
   id: String // TM2a
   name: String? // TM2g
   timestamp: Time // TM2f
+  action: enum // TM2j
+  serial: string // RTN7b, TM2K
```
AndyTWF, line 2172: The serial here is different to `RTN7b` afaik - `RTN7b` is a `ProtocolMessage` field. In fact, the `msgSerial` of the `ProtocolMessage` is a uint, not a timeserial.

AndyTWF: Ditto on the line below

textile/features.textile
```diff
@@ -1305,6 +1305,22 @@ h4. Message
...
 ** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
+** @(TM2j)@ @action@ enum - the @MessageAction@ type of the message. Any message received from Ably must have an @action@ field set to one of the following values:
+*** @(TM2j1)@ 0: @MESSAGE_UNSET@
+*** @(TM2j2)@ 1: @MESSAGE_CREATE@
+*** @(TM2j3)@ 2: @MESSAGE_UPDATE@
+*** @(TM2j4)@ 3: @MESSAGE_DELETE@
+*** @(TM2j5)@ 4: @MESSAGE_ANNOTATION_CREATE@
+*** @(TM2j6)@ 5: @MESSAGE_ANNOTATION_DELETE@
+** @(TM2K)@ @serial@ string - the serial of the message. This will be set to the @msgSerial@ of the @ProtocolMessage@ that encapsulates the message, with an additional offset that locates the message in the @ProtocolMessage@ messages array. If the message is received from Ably, this must be set.
```
AndyTWF, line 1315: Should this be the "origin timeserial assigned to encapsulating `ProtocolMessage` after publish"?

Per below - the `msgSerial` field in the `ProtocolMessage` isn't a timeserial

splindsay-92: I've updated this, I still want to make it clear that it is the serial of the encapsulating ProtocolMessage, but with the Idx.

textile/features.textile
splindsay-92: You're quite correct, I've updated this to point to the `@serial@` field I've added to the `@Message@` definition.

textile/features.textile
```diff
@@ -654,6 +655,7 @@ h3(#realtime-channel). RealtimeChannel
 * @(RTL6)@ @RealtimeChannel#publish@ function:
 ** @(RTL6a)@ Messages are encoded in the same way as the @RestChannel#publish@ method, and "RSL1g":#RSL1g (size limit) applies similarly
 *** @(RTL6a1)@ "RSL1k":#RSL1k (@idempotentRestPublishing@ option), "RSL1j1":#RSL1j1 (idempotent publishing test), and "RSL1l":#RSL1l (@publish(Message, params)@ form) do not apply to realtime publishes
+*** @(RTL6a2)@ Just like "RSL1j1":#RSL1j1 (publishing via REST), @Message@ objects can have the "TM2j2":#TM2j2 (message action field) set to @MESSAGE_CREATE@, if unset, this will be set by realtime to @MESSAGE_CREATE@.
```
AndyTWF, line 658:
```suggestion
*** @(RTL6a2)@ Similar to "RSL1j1":#RSL1j1 (publishing via REST), @Message@ objects can have the "TM2j2":#TM2j2 (message action field) set to @MESSAGE_CREATE@, if unset, this will be set by realtime to @MESSAGE_CREATE@.
```

Sounds a bit more formal :)

textile/features.textile
```diff
@@ -1305,6 +1307,23 @@ h4. Message
...
 ** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
+** @(TM2j)@ @action@ enum - the type of the message. Any message received from Ably must have an @action@ field set to one of the following values:
+*** @(TM2j1)@ 0: @MESSAGE_UNSET@
+*** @(TM2j2)@ 1: @MESSAGE_CREATE@
+*** @(TM2j3)@ 2: @MESSAGE_UPDATE@
+*** @(TM2j4)@ 3: @MESSAGE_DELETE@
+*** @(TM2j5)@ 4: @ANNOTATION_CREATE@
+*** @(TM2j6)@ 5: @ANNOTATION_DELETE@
+*** @(TM2j7)@ 6: @META_OCCUPANCY@
+** @(TM2K)@ @serial@ string - the origin timeserial of the message. This will be set, after publish, to the @channelSerial@ of the @ProtocolMessage@ that encapsulated the @Message@, with an additional offset that locates the @Message@ in the @ProtocolMessage@ messages array. If the message is received from Ably, this must be set.
```
SimonWoolf, line 1318: Remember that this is the SDK features spec, it specifies how SDK's should be implemented.

Put yourself in the position of an SDK developer reading this (and the rest of the TM2 series). Your job is to make your SDK compliant with these requirements. How do I do that? "This will be set, after publish, to the channelSerial of the ProtocolMessage..." -- ok, how do I do that? After publish as in after it's been sent on the wire? But by then I can't modify it. Is it telling me to set it? If it's received from ably it "must" be set? So what do I do if it isn't?

Answer: you can't, and actually there is no requirement for the SDK implementation here other than to expose the `serial`. Plus a brief explanation for a docstring or whatever. `* @(TM2k)@ @serial@ string, an opaque string that uniquely identifies this message` would do fine

textile/features.textile
```diff
@@ -1305,6 +1307,23 @@ h4. Message
...
+** @(TM2L)@ @refSerial@ string - if this message is an annotation that references another message, this must be set to the @serial@ of the referenced message. If the @refSerial@ is set, the @refType@ must also be set.
+** @(TM2M)@ @refType@ string - if the @refSerial@ is set, this must be set. If describes the type of the reference.
+** @(TM2N)@ @updatedAt@ time in milliseconds since epoch. If a message received from Ably has the @action@ field set to "MESSAGE_UPDATE" then the @updatedAt@ field must be set to the time the message was updated.
+** @(TM2O)@ @deletedAt@ time in milliseconds since epoch. If a message received from Ably has the @action@ field set to "MESSAGE_DELETE" then the @deleteAt@ field must be set to the time the message was deleted.
```
SimonWoolf, line 1322: (similar comments to above apply to most of these, I won't repeat)

textile/features.textile
```diff
@@ -1305,6 +1307,23 @@ h4. Message
...
+** @(TM2P)@ @operation@ operation is an JSON-encodable object that contains the following `optional` fields;
```
SimonWoolf, line 1323: I don't see why we'd need to make this a JsonObject. It's a well-defined optional substruct of the Message with three known fields, why shouldn't the library  expose it accordingly? (In JS there might not be a difference (other than the typings), it could just be a plain object, but eg in java this would be much more usable as a MessageOperation class or whatever vs a JSONObject)

textile/features.textile
```diff
@@ -2154,6 +2173,13 @@ class Message: // TM*
   id: String // TM2a
   name: String? // TM2g
   timestamp: Time // TM2f
+  action: enum // TM2j
```
SimonWoolf, line 2176: the IDL defines enums explicitly, eg look at `PresenceAction`

SimonWoolf (changes requested):

textile/features.textile
```diff
@@ -300,6 +300,7 @@ h3(#rest-channel). RestChannel
 ** @(RSL1h)@ The @publish(name, data)@ form should not take any additional arguments. If a client library has supported additional arguments to the @(name, data)@ form (e.g. separate arguments for @clientId@ and @extras@, or a single @attributes@ argument) in any 1.x version, it should continue to do so until version 2.0.
 ** @(RSL1i)@ If the total size of the message or (if publishing an array) messages, calculated per "TO3l8":#TO3l8, exceeds the @maxMessageSize@, then the client library should reject the publish and indicate an error with code 40009
 ** @(RSL1j)@ When @Message@ objects are provided, any valid @Message@ attribute (that is, an attribute specified in "TM2":#TM2) that is supplied by the caller must be included in the encoded message. (This does not mean it must be included _unaltered_; for example the @data@ and @encoding@ will be subject to processing per "RSL4":#RSL4)
+*** @(RSL1j1)@ On calling @publish@ and if @Message@ objects are provided, the library can set the "TM2j2":#TM2j2 (message action field) of each @Message@ to @MESSAGE_CREATE@, if unset, realtime will set the action to @MESSAGE_CREATE@.
```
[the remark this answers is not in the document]
splindsay-92: I think I was taking the wrong view of what needs to be included in the spec here, so I will remove the addition as a whole. Thanks for the additional feedback though!

textile/features.textile
splindsay-92: Removed

textile/features.textile
```diff
@@ -1305,6 +1307,23 @@ h4. Message
...
 ** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
+** @(TM2j)@ @action@ enum - the type of the message. Any message received from Ably must have an @action@ field set to one of the following values:
```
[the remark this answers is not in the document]
splindsay-92: I'll update to make things clearer here.

textile/features.textile
splindsay-92: Thanks for the feedback Simon, this is really useful! I'll update the message string too

textile/features.textile
splindsay-92: Updated to be consistent with `PresenceAction`

textile/features.textile
```diff
@@ -1305,6 +1305,16 @@ h4. Message
...
+** @(TM2j)@ @action@ enum
+** @(TM2K)@ @serial@ string - an opaque string that uniquely identifies the message.
+** @(TM2L)@ @refSerial@ string - an opaque string that uniquely identifies some referenced message.
+** @(TM2M)@ @refType@ string - an opaque string that identifies the type of this reference.
+** @(TM2N)@ @operation@ - object that may contain the following `optional` attributes;
+*** @(TM2N1)@ @clientId@ string
+*** @(TM2N2)@ @description@ string
+*** @(TM2N3)@ @metadata@ object - used to contain any arbitrary key value pairs of type string.
+** @(TM2O)@ @updatedAt@ time in milliseconds since epoch at which an operation occurred. This field is always populated on messages received with an @action@ of @MESSAGE_UPDATE@ or @MESSAGE_DELETE@..
+** @(TM2P)@ @updateSerial@ string - an opaque string that uniquely identifies the operation. This field is always populated on messages received with an @action@ of @MESSAGE_UPDATE@ or @MESSAGE_DELETE@.
```
SimonWoolf, line 1317: - the spec uses the convention that the letters in spec items other than the initial section name (eg the p in TM2p) is lowercase
- you reference a TM5 from the IDL that lists the actions, where is it?

SimonWoolf (approved):
Needs a couple small further changes, but pre-approving after you make those.

textile/features.textile
splindsay-92: I've made the case changes and included TM5.

AndyTWF (approved):

PR #212, opened 2024-10-21
Spec: SUBSCRIBE mode renamed to MESSAGE_SUBSCRIBE

mschristensen:
The `SUBSCRIBE` channel mode has now been renamed to `MESSAGE_SUBSCRIBE` to better match the existing `PRESENCE_SUBSCRIBE` mode as well as upcoming channel state related modes (to be addressed separately), and for better parity with capability strings.

SEE: https://github.com/ably/realtime/discussions/6352

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1343,7 +1343,8 @@ h4. ProtocolMessage
 ** @(TR3f)@ 5: @ATTACH_RESUME@
 ** @(TR3q)@ 16: @PRESENCE@
 ** @(TR3r)@ 17: @PUBLISH@
-** @(TR3s)@ 18: @SUBSCRIBE@
+** @(TR3s)@ 18: @SUBSCRIBE@ Synonymous with @MESSAGE_SUBSCRIBE@. Retained for backward compatibility.
+** @(TR3u)@ 18: @MESSAGE_SUBSCRIBE@ Synonymous with @SUBSCRIBE@. This variant should be preferred.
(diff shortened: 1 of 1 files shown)


SimonWoolf (changes requested):
from https://github.com/ably/realtime/discussions/6352:
> But note: for backwards compatibility, the existing `SUBSCRIBE` mode retains its current meaning, synonymous with `MESSAGE_SUBSCRIBE`. 
else this will be a breaking change for people who are currently specifying SUBSCRIBE. (which would be fine in a major sdk version break, but that's not being proposed here, and it would be way overkill when we can just make it backwards-compatible)

mschristensen:
@SimonWoolf I've updated to try to clarify that `MESSAGE_SUBSCRIBE` is synonymous with `SUBSCRIBE` and that the latter is still supported by still including it in the spec. Please let me know if you think there's a way to express this more clearly than what I've done

SimonWoolf:
I think it's worth making clear which one is the main one going forward and which one is the deprecated synonym, eg for purposes of exposing in channel.modes (RTL4m); the PR in its current state has them symmetric.

(I don't think we need to worry too much about backwards compatibility re rtl4m, I doubt anyone's parsing that; but we definitely do for taking an array of modes in channeloptions)

mschristensen:
@SimonWoolf updated

SimonWoolf (approved):

Issue #211, opened 2024-10-17
RFC: only emit a leave if there was an existing member


SimonWoolf:
Proposal: have the spec specify that you only emit a leave event if a member was removed.

Presence events are only emitted by the SDK if they pass a 'newness check': [RTP2g](https://sdk.ably.com/builds/ably/specification/main/features/#RTP2g). The general idea is that presence is a CRDT, and you don't need to emit an event that does not change your local presence set state, only newer versions of the presence message for a particular member (which might container newer versions of the data).

That reasoning would also imply that we should only emit a leave event if a matching member was previously present in the local view of the presence set, else the leave does not cause a change in local state. But we don't currently do that. Why not? This feels like an obvious thing to do, I feel like I must be missing something?

SimonWoolf:
> For example, if a client explicitly leaves presence and immediately detaches (causing an implicit leave), is there a guarantee here that the explicit leave will always win-out?

Realtime transports preserve order. So if you leave before sending a detach, then yes, the explicit leave should arrive first.

(That said, if someone's designed something that depends on the client being able to specify the data in the leave event, that's a bit of a design smell. Half the point of presence as a feature is that the service will leave on your behalf if you eg have internet issues and disconnect abruptly, and in those situations the client clearly can't control the leave data)

SimonWoolf:
This was discussed in the RTF and agreed, I will turn this into a spec pr

SimonWoolf:
https://github.com/ably/specification/pull/222

PR #222, opened 2024-11-13
Features spec: only emit a leave if there was an existing matching member

SimonWoolf:
per https://github.com/ably/specification/issues/211

The actual behaviour change here is small (only whether non-sync leaves where there is no matching member currently present should emit a leave event), but I found the current structure of this spec section a bit confusing (why are broadcast and members-map add/remove requirements defined separately, despite having the same requirements?), so ended up rewriting it. (Without deprecating the old spec items since this isn't actually a behaviour change except in that one case)

change was discussed and agreed at the 2024-11-13 RTF


textile/features.textile
```diff
@@ -741,16 +741,23 @@ h3(#realtime-presence). RealtimePresence
...
+*** @(RTP2d1)@ It must be emitted to @RealtimePresence@ subscribers (with its original presence action, and an event name set to the stringified form of that action)
+*** @(RTP2d2)@ It must be added to the presence map with the action set to @PRESENT@
+** @(RTP2h)@ When a presence message with an action of @LEAVE@ arrives, if and only if there is a member with a matching @memberKey@ currently in the presence map (and the incoming event satisfies the @RTP2b@ newness check against it):
+*** @(RTP2h1)@ If a @SYNC@ is not in progress:
+**** @(RTP2h1a)@ The incoming message must be emitted to @RealtimePresence@ subscribers (with an event name set to the stringified form of the @LEAVE@ action)
+**** @(RTP2h1b)@ That member must be deleted from the presence map
+*** @(RTP2h2)@ If a @SYNC@ is in progress:
+**** @(RTP2h2a)@ The incoming message must be stored in the presence map with the action set to @ABSENT@
+**** @(RTP2h2b)@ When the @SYNC@ completes, then all @ABSENT@ members in the presence map must be deleted. (No leave events should be emitted other than those required by @RTP19@)
+** @(RTP2g)@ This clause has been replaced by @RTP2d1@ and @RTP2h1a@
```
lawrence-forooghian, line 760: Can we stick to the language from https://github.com/ably/specification/blob/main/CONTRIBUTING.md#features-spec-points?

textile/features.textile
```diff
@@ -741,16 +741,23 @@ h3(#realtime-presence). RealtimePresence
...
 **** @(RTP2b1a)@ If the timestamps compare equal, the newly-incoming message is considered newer than the existing one
 *** @(RTP2b2)@ Else split the @id@ of both presence messages (which will be of the form @connid:msgSerial:index@, e.g. @aaaaaa:0:0@) on the separator @:@, and parse the latter two as integers. Compare them first by @msgSerial@ numerically, then (if @msgSerial@ is equal) by @index@ numerically, larger being newer in both cases
 ** @(RTP2c)@ As there are no guarantees that during a @SYNC@ operation presence events will arrive in order, all presence messages from a @SYNC@ must also be compared for newness in the same way as they would from a @PRESENCE@
-** @(RTP2d)@ When a presence message with an action of @ENTER@, @UPDATE@, or @PRESENT@ arrives, it should be added to the presence map with the action set to @PRESENT@
-** @(RTP2e)@ If a @SYNC@ is not in progress, then when a presence message with an action of @LEAVE@ arrives,  that @memberKey@ should be deleted from the presence map, if present
```
lawrence-forooghian, line 751: RTP2e and RTP2f need to be marked as deleted or replaced (see https://github.com/ably/specification/blob/main/CONTRIBUTING.md#features-spec-points) so that we don't create new spec points with these identifiers.

textile/features.textile
```diff
@@ -741,16 +741,23 @@ h3(#realtime-presence). RealtimePresence
...
-** @(RTP2f)@ If a @SYNC@ is in progress, then when a presence message with an action of @LEAVE@ arrives, it should be stored in the presence map with the action set to @ABSENT@. When the @SYNC@ completes, any @ABSENT@ members should be deleted from the presence map. (This is because in a @SYNC@, we might receive a @LEAVE@ before the corresponding @ENTER@).
-** @(RTP2g)@ Any incoming presence message that passes the newness check should be emitted on the @RealtimePresence@ object, with an event name set to its original action. Note: this action may not be the same one that it will have when stored in the presence map. For example: an incoming presence message with an @ENTER@ action will be emitted as an @enter@ event, and the emitted presence message will have its action set to @ENTER@. However, it will be stored in the presence map with a @PRESENT@ action.
+** @(RTP2d)@ When a presence message with an action of @ENTER@, @UPDATE@, or @PRESENT@ arrives (if if it satisfies the @RTP2a@ newness check):
```
lawrence-forooghian, line 750:
```suggestion
** @(RTP2d)@ When a presence message with an action of @ENTER@, @UPDATE@, or @PRESENT@ arrives (if it satisfies the @RTP2a@ newness check):
```

textile/features.textile
```diff
@@ -741,16 +741,23 @@ h3(#realtime-presence). RealtimePresence
...
-** @(RTP2e)@ If a @SYNC@ is not in progress, then when a presence message with an action of @LEAVE@ arrives,  that @memberKey@ should be deleted from the presence map, if present
-** @(RTP2f)@ If a @SYNC@ is in progress, then when a presence message with an action of @LEAVE@ arrives, it should be stored in the presence map with the action set to @ABSENT@. When the @SYNC@ completes, any @ABSENT@ members should be deleted from the presence map. (This is because in a @SYNC@, we might receive a @LEAVE@ before the corresponding @ENTER@).
-** @(RTP2g)@ Any incoming presence message that passes the newness check should be emitted on the @RealtimePresence@ object, with an event name set to its original action. Note: this action may not be the same one that it will have when stored in the presence map. For example: an incoming presence message with an @ENTER@ action will be emitted as an @enter@ event, and the emitted presence message will have its action set to @ENTER@. However, it will be stored in the presence map with a @PRESENT@ action.
```
lawrence-forooghian, line 753: So if I’ve understood correctly, the functional change in this PR is that RTP2g’s "emit an event for _all_ newness-check-satisfying incoming presence messages" has been split into two rules: one for `LEAVE` (to only emit an event if there was an existing local member) and another for the other actions?

SimonWoolf: Yep. (hence the PR title)

lawrence-forooghian (approved):

PR #230, opened 2024-11-19
[Chat][Room-Release] Remove CHA-RL3c

sacOO7:
- CHA-RL3c—If the room is in the @RELEASING@ status, the operation will return the result of the pending @release@ operation or throw any exception it throws.
- This statement conflicts with the atomic behavior of the room lifecycle manager [CHA-RL7].
- This also conflicts with [CHA-RL3k], where every new operation starts only after the previous one has finished.
- Also, `release` operation never throws any exceptions and always sets channels into either `detached` or `failed` status, setting RoomStatus as `Released` at the end.
- This means once release operation starts, it will always finish emitting `Releasing` -> `Released` states.
- New `Release` operation will be automatically covered as per CHA-RL3a.
- Related discussion for the same -> https://github.com/ably/ably-chat-js/issues/399

### DIFF
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -116,7 +116,7 @@ h4(#rooms-lifecycle-operations). Room Lifecycle Operations
 ** @(CHA-RL3a)@ @[Testable]@ If the room is already in the @RELEASED@ status, this operation is no-op.
 ** @(CHA-RL3b)@ @[Testable]@ If the room is in the @DETACHED@ status, then the room is immediately transitioned to @RELEASED@ and the operation succeeds.
 ** @(CHA-RL3j)@ @[Testable]@ If the room is in the @INITIALIZED@ status, then the room is immediately transitioned to @RELEASED@ and the operation succeeds.
-** @(CHA-RL3c)@ @[Testable]@ If the room is in the @RELEASING@ status, then the operation will return the result of the pending @release@ operation, or throw any exception that it throws.
+** @(CHA-RL3c)@ This specification point has been removed.
 ** @(CHA-RL3i)@ This specification point has been removed.
 ** @(CHA-RL3l)@ @[Testable]@ When the release operation commences, the room will transition into the @RELEASING@ status and any transient disconnect timeouts shall be cleared.
 ** @(CHA-RL3d)@ @[Testable]@ All features channels must be detached in sequence.


textile/chat-features.textile
```diff
@@ -116,7 +116,7 @@ h4(#rooms-lifecycle-operations). Room Lifecycle Operations
 ** @(CHA-RL3a)@ @[Testable]@ If the room is already in the @RELEASED@ status, this operation is no-op.
 ** @(CHA-RL3b)@ @[Testable]@ If the room is in the @DETACHED@ status, then the room is immediately transitioned to @RELEASED@ and the operation succeeds.
 ** @(CHA-RL3j)@ @[Testable]@ If the room is in the @INITIALIZED@ status, then the room is immediately transitioned to @RELEASED@ and the operation succeeds.
-** @(CHA-RL3c)@ @[Testable]@ If the room is in the @RELEASING@ status, then the operation will return the result of the pending @release@ operation, or throw any exception that it throws.
+** @(CHA-RL3c)@ This specification point has been removed.
```
AndyTWF, line 119: Perhaps a slight reword might work here - the intended behaviour is "we're releasing, so atomically this operation doesn't run until the previous one completes", so perhaps:

@(CHA-RL3c)@ @[Testable]@ If the a release operation in process, the call will wait until the previous operation is complete before proceeding. For example, if the room releases successfully, then the subsequent call will resolve per @CHA-RL3a@.

WDYT?

sacOO7: - Actually, as I mentioned in the [PR summary](https://github.com/ably/specification/pull/230#issue-2671661962), the release operation never throws any exception and will always reach the `RELEASED` state.
- So, the statement `If the release operation in process` is never true according to [CHA-RL7] : |
- This implies that, per CHA-RL7, new operations will always have `_releaseInProgress` set to false.
- Additionally, I am confident chat-js code will [never enter this block](https://github.com/ably/ably-chat-js/blob/ee554e3d19e2ec718a3b3c90974753ee05ad6b52/src/core/room-lifecycle-manager.ts#L799), of all tests we have as a part of chat-js test-suite. So, the check and related implementation is redundant.

sacOO7: Not just for this spec point, but we shouldn't have `in progress` checks as a part of any other Room operations.

PR #235, opened 2024-11-20
Update field names for mutable messages, and specify how message.serial is set from channelSerial

SimonWoolf:
Per discussion and subsequent call
https://ably-real-time.slack.com/archives/C02NY1VT3LY/p1732031493162379

Didn't bother deprecating and assigning new spec names since it's only been implemented by us in one sdk so far, we can just retcon


textile/features.textile
```diff
@@ -1332,17 +1332,17 @@ h4. Message
...
-** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
+** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably over a realtime transport does not contain a @timestamp@, the SDK must set it to the @timestamp@ of the encapsulating @ProtocolMessage@
 ** @(TM2j)@ @action@ enum
-** @(TM2k)@ @serial@ string - an opaque string that uniquely identifies the message.
+** @(TM2p)@ @version@ string - an opaque string that uniquely identifies the message, and is different for different versions. If a message received from Ably over a realtime transport does not contain a @version@, the SDK must set it to @<channelSerial>:<padded_index>@ from the @channelSerial@ field of the enclosing @ProtocolMessage@, and @padded_index@ is the index of the message inside the @messages@ array of the @ProtocolMessage@, left-padded with 0s to three digits (for example, the second entry might be @foo:001@).
```
AndyTWF, line 1337: Do we need a spec point here that limits ProtocolMessage to 999 messages? I vaguely remember discussing it but couldn't remember the outcome.

textile/features.textile
```diff
@@ -1332,17 +1332,17 @@ h4. Message
...
-** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
+** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably over a realtime transport does not contain a @timestamp@, the SDK must set it to the @timestamp@ of the encapsulating @ProtocolMessage@
```
AndyTWF, line 1335: Worth clarifying that this will always represent the time at which the latest event/action was received by the ably servers?

Edit: I guess its implicit by the fact we have a `createdAt` timestamp, so may not be necessary.

AndyTWF (approved):

textile/features.textile
SimonWoolf: I don't think we need it here, because there's nothing an sdk can do at this point. (and in general an sdk doesn't need to know about a limit to enforce it). But thanks for bringing it up 'cause that reminds me there is one place we do need it, which is RTL6d, I'll add that

textile/features.textile
SimonWoolf: Also this isn't customer documentation, this is the SDK spec, SDK devs don't need to care what it represents they just need to handle the field correctly

AndyTWF: Sure, sounds good

SimonWoolf:
Made one small further tweak -- specified that SDKs should copy over version/timestamp to serial/createdAt for CREATEs for messages received over REST as well, not just realtime. https://github.com/ably/specification/pull/235/commits/be1855a3e3bd07088d1de0915a32d7b7ec455fa2

AndyTWF (approved):

PR #237, opened 2024-11-21
Add some spec points about disabled chat features

lawrence-forooghian:


textile/chat-features.textile
```diff
@@ -194,7 +194,10 @@ Each chat room can be configured individually, allowing options to be passed as
 *** @(CHA-RC1g5)@ @[Testable]@ The room is removed from the room map and a @CHA-RL3@ release operation is initiated for the room object. A future is returned which resolves when the release operation completes.
 * @(CHA-RC2)@ Chat rooms are configurable, so as to enable or disable certain features. When requesting a room, options as to which features should be enabled, and the configuration they should take, must be provided (@RoomOptions@).
 ** @(CHA-RC2a)@ @[Testable]@ If a room is requested with invalid configuration, for example: a negative typing timeout, an @ErrorInfo@ with code @40001@ must be thrown.
+** @(CHA-RC2c)@ @[Testable]@ A room’s messages feature (@CHA-M*@) can not be disabled.
```
AndyTWF, line 197: Might be worth a spec point saying that contributors should be ordered, and that messages is always first (this was a conscious choice so that messages would always be the first thing working)

lawrence-forooghian: Ah, I wasn't aware of this requirement; what is the order of the remaining contributors?

AndyTWF: At the moment - one isn't determined, but chat-js is presence, typing, reactions, occupancy

lawrence-forooghian: Would you mind pushing a commit here to add this requirement worded in a way that satisfies the requirements you have in mind?

AndyTWF: Done - I added CHA-RL10 (because 9 is reserved by another PR) and then references CHA-RC2e which specifies the order. The old 2e is now 2f - as the new spec point fits better next to 2d.

lawrence-forooghian: Cool 👍 — happy to approve now?

AndyTWF (approved):

PR #238, opened 2024-11-25
Update/Limit default typing timeout

sacOO7:
- The current default typing timeout is set to `5 seconds`.
- From the customer's perspective, the receiver will continue to see the typing event from the sender even after the sender has stopped typing for 5 seconds.
- This can feel awkward, even in a [demo app](https://ably-real-time.slack.com/archives/C03JDBVM5MY/p1732287870937399?thread_ts=1732287678.314829&cid=C03JDBVM5MY)
- Ideally, default values should reflect realistic use cases and avoid producing unexpected behavior.
- I believe the default typing timeout should be capped at `1 second` since the event is transmitted in real-time, and we don’t want the receiver to see “typing” events after the sender has stopped typing.
  
**PS**: While we currently have a negative timeout check for typing events, we should also introduce a `max_typing_timeout` check to prevent users from setting unrealistic values.

### DIFF
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -402,7 +402,7 @@ Typing Indicators allow chat room users to indicate to others that they are typi
 ** @(CHA-T2d)@ @[Testable]@ If the room status is @Attached@, then the @get@ call shall invoke the underlying @presence.enter()@ call.
 ** @(CHA-T2e)@ @[Testable]@ If the room status is @Detached@, an @ErrorInfo@ with code 40000, explaining that attach must be called first, will be thrown.
 ** @(CHA-T2f)@ @[Testable]@ For any other room status, an @ErrorInfo@ with code 40000 will be thrown.
-* @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 5000ms.
+* @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 1000ms.
 * @(CHA-T4)@ Users may indicate that they have started typing.
 ** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
 *** @(CHA-T4a1)@ @[Testable]@ When a typing session is started, the client is entered into presence on the typing channel.


textile/chat-features.textile
```diff
@@ -402,7 +402,7 @@ Typing Indicators allow chat room users to indicate to others that they are typi
 ** @(CHA-T2d)@ @[Testable]@ If the room status is @Attached@, then the @get@ call shall invoke the underlying @presence.enter()@ call.
 ** @(CHA-T2e)@ @[Testable]@ If the room status is @Detached@, an @ErrorInfo@ with code 40000, explaining that attach must be called first, will be thrown.
 ** @(CHA-T2f)@ @[Testable]@ For any other room status, an @ErrorInfo@ with code 40000 will be thrown.
-* @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 5000ms.
+* @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 1000ms.
```
lawrence-forooghian, line 405: @sacOO7 if we decide to go ahead with this change, please can you [mark this spec point as replaced](https://github.com/ably/specification/blob/main/CONTRIBUTING.md#additional-notes-on-features-spec-point-removal-and-replacement) since we've already implemented this spec point in Swift

sacOO7: Sure, will do that.
Also, post your thoughts about default timeout / max capping for the same as per https://github.com/ably/specification/pull/238#issue-2690234910

AndyTWF:
The reason we set this number high was quite deliberate - we wanted to make it so that when users experiment with it, it was something that had a sense of permanence to it (we felt going too short would make it look temperamental, whereas what we wanted to show was a long enough time that the developer could observe it). For reference, when we tried with other platforms timeouts, most of them were around 3 seconds.

> capped at 1 second since the event is transmitted in real-time, and we don’t want the receiver to see “typing” events after the sender has stopped typing.

Wouldn't we get that anyway? The receiver is still going to receive presence events in order, and we provide realtime ordering guarantees, so they'll see something regardless.

> we should also introduce a max_typing_timeout check to prevent users from setting unrealistic values.

Is there a technical reason for us to do this? I'd rather not apply arbitrary constraints on users unless there's a technical reason to do so. Sure, 10 seconds is long - but if that's what somebody wants, we probably shouldn't stop them.

sacOO7:
> Wouldn't we get that anyway? The receiver is still going to receive presence events in order, and we provide realtime ordering guarantees, so they'll see something regardless.

I’m not sure I fully understand this. The receiver will still get the `stop` event after 5 seconds, even though the sender has already stopped typing (before 5 seconds). Wouldn’t it be better to have default values that make more sense to the customer, rather than providing values that might seem like an error or misconfiguration? 

I do see the point that they might want to configure the `typing` timeout themselves, but at the same time, we'd be forcing them to set a specific timeout value. The problem is that this value might not align with what they consider a "standard" timeout. So, having a current timeout of 5 seconds doesn't seem ideal, especially since the customer will likely need to adjust it anyway. 

The current demo app works as per default configuration, but from the user's perspective, the typing event looks like a bug or misconfiguration. I’m also concerned that developers might end up with a negative impression, which goes against the experience we’re trying to create.

AndyTWF:
> The receiver will still get the stop event after 5 seconds, even though the sender has already stopped typing (before 5 seconds). Wouldn’t it be better to have default values that make more sense to the customer, rather than providing values that might seem like an error or misconfiguration?

After 5 seconds, IFF they haven't called `typing.stop()` in the meantime. If someone actually sends the message, then you'd expect the developer to call this to cancel typing earlier.

5 seconds was chosen (as well as the above reasons) because that sort of range was what we saw in other applications we looked at. Every customer is different, some may want a near-instant timeout, others might want a bit of leniency. I don't think we can realistically predict a value that's going to work in all cases. Of course, this is only if someone starts typing and then just sits there.

PR #232, opened 2024-11-19
chat: add CHA-RL9 to explain "attaching piggyback"

AndyTWF:
Closes #228

PR #249, opened 2024-11-28
chat: room attachment requirements

AndyTWF:
Adding room attachment requirements to Typing and Presence.leave

### DIFF
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -416,12 +416,17 @@ Typing Indicators allow chat room users to indicate to others that they are typi
 * @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 5000ms.
 * @(CHA-T4)@ Users may indicate that they have started typing.
 ** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
-*** @(CHA-T4a1)@ @[Testable]@ When a typing session is started, the client is entered into presence on the typing channel.
-*** @(CHA-T4a2)@ @[Testable]@ When a typing session is started, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session is automatically ended by leaving presence.
+*** @(CHA-T4a3)@ @[Testable]@ If the room status is @Attaching@, then the operation shall invoke a @CHA-RL9@ pending room status operation and wait for it to complete. If this operation fails, then the error from @CHA-RL9@ must be raised. If this succeeds (i.e the room becomes @ATTACHED@), the library will proceed per @CHA-T4a1@.
+*** @(CHA-T4a4)@ @[Testable]@ If the room status is anything other than @Attached@ or @Attaching@, an @ErrorInfo@ using the @RoomInInvalidState@ error code from the "chat-specific error codes":#error-codes with a status code of @400@ must be thrown. The message shall explain that attach must be called first.
+*** @(CHA-T4a1)@ @[Testable]@ If the room status is @Attached@, then the @start@ call will invoke the underlying @presence.enter()@ call.
+*** @(CHA-T4a2)@ @[Testable]@ When a typing session is started per @CHA-T4a1@, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session is automatically ended by leaving presence.
 ** @(CHA-T4b)@ @[Testable]@ If typing is already in progress, the @CHA-T3@ timeout is extended to be @timeoutMs@ from now.
 * @(CHA-T5)@ Users may indicate that they have stopped typing.
 ** @(CHA-T5a)@ @[Testable]@ If typing is not in progress, this operation is no-op.
-** @(CHA-T5b)@ @[Testable]@ If typing is in progress, he @CHA-T3@ timeout is cancelled. The client then leaves presence.
+** @(CHA-T5b)@ This specification point has been removed. It was superseded by @CHA-T5e@.
+** @(CHA-T5c)@ @[Testable]@ If typing is in progress, and the room status is @Attaching@, then the operation shall invoke a @CHA-RL9@ pending room status operation and wait for it to complete. If this operation fails, then the error from @CHA-RL9@ must be raised. If this succeeds (i.e the room becomes @ATTACHED@), the library will proceed per @CHA-T5b@.
+** @(CHA-T5d)@ @[Testable]@ If typing is in progress, and the room status is anything other than @Attached@ or @Attaching@, an @ErrorInfo@ using the @RoomInInvalidState@ error code from the "chat-specific error codes":#error-codes with a status code of @400@ must be thrown. The message shall explain that attach must be called first.
+** @(CHA-T5e)@ @[Testable]@ If typing is in progress, and the room status is @Attached@, the client leaves presence and the @CHA-T3@ timeout is cancelled.
 * @(CHA-T6)@ Users may subscribe to typing events - updates to a set of clientIDs that are typing. This operation, like all subscription operations, has no side-effects in relation to room lifecycle.
 ** @(CHA-T6a)@ @[Testable]@ Users may provide a listener to subscribe to "typing events":#chat-structs-typing-event in a room.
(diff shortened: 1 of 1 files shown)


sacOO7:
PR is upgrade to https://github.com/ably/specification/pull/232

sacOO7:
Can we also update 
`@(CHA-PR10e)@ @[Testable]@ If the room status is @Attached@, then the @update@ call will invoke the underlying @presence.enter()@ call.`
to 
`@(CHA-PR10e)@ @[Testable]@ If the room status is @Attached@, then the @update@ call will invoke the underlying @presence.update()@ call.
`

sacOO7:
While I agree that we need to ensure the room state is `ATTACHED` before performing any room operations, we currently face a lot of specification duplication for user-triggered room actions.

In Kotlin, we've defined the `ensureAttached` method as part of the `Room` class, which returns a promise (in JavaScript) or a callback in other languages.

[[DefaultRoom.ensureAttached() method](https://github.com/ably/ably-chat-kotlin/blob/16ead0cec25650a1d860144307e51f508f8e07e0/chat-android/src/main/java/com/ably/chat/Room.kt#L243-L272)](https://github.com/ably/ably-chat-kotlin/blob/16ead0cec25650a1d860144307e51f508f8e07e0/chat-android/src/main/java/com/ably/chat/Room.kt#L243-L272):
- This method is simply called before executing any user-triggered real-time operations, at multiple places. 
- It's easy to test and avoid lot of code duplication/ simplifies spec understanding.

sacOO7:
Perhaps we can streamline the specification so that we don't need to define multiple spec points (currently 3) for basic user-triggered real-time room operations. This will help in future specifications as well.

AndyTWF:
> While I agree that we need to ensure the room state is `ATTACHED` before performing any room operations, we currently face a lot of specification duplication for user-triggered room actions.
> 
> In Kotlin, we've defined the `ensureAttached` method as part of the `DefaultRoom` class, which returns a promise (in JavaScript) or a callback in other languages.
> 
> [DefaultRoom.ensureAttached() method](https://github.com/ably/ably-chat-kotlin/blob/16ead0cec25650a1d860144307e51f508f8e07e0/chat-android/src/main/java/com/ably/chat/Room.kt#L243-L272):
> 
> * This method is simply called before executing any user-triggered real-time operations, at multiple places.
> * It's easy to test and avoid lot of code duplication/ simplifies spec understanding.

I'm cautious to define exact implementation details (e.g. specific method names) in the spec, so as to give latitude to different SDKs to be idiomatic to their ecosystems.

But in a future PR I'll look at consolidating the points - perhaps a GA item

textile/chat-features.textile
```diff
@@ -413,12 +416,16 @@ Typing Indicators allow chat room users to indicate to others that they are typi
 * @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 5000ms.
 * @(CHA-T4)@ Users may indicate that they have started typing.
 ** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
-*** @(CHA-T4a1)@ @[Testable]@ When a typing session is started, the client is entered into presence on the typing channel.
```
sacOO7, line 416: We should mark `CHA-T4a1` as removed, instead it's redefined at the bottom. Is this acceptable?

AndyTWF: There's not any change in functionality (beyond what was originally intended) - so its just a rewording :)

textile/chat-features.textile
```diff
@@ -413,12 +416,16 @@ Typing Indicators allow chat room users to indicate to others that they are typi
...
+*** @(CHA-T4a3)@ @[Testable]@ If the room status is @Attaching@, then the operation shall invoke a @CHA-RL9@ pending room status operation and wait for it to complete. If this operation fails, then the error from @CHA-RL9@ must be raised. If this succeeds (i.e the room becomes @ATTACHED@), the library will proceed per @CHA-T4a1@.
+*** @(CHA-T4a4)@ @[Testable]@ If the room status is anything other than @Attached@ or @Attaching@, an @ErrorInfo@ using the @RoomInInvalidState@ error code from the "chat-specific error codes":#error-codes with a status code of @400@ must be thrown. The message shall explain that attach must be called first.
+*** @(CHA-T4a1)@ @[Testable]@ If the room status is @Attached@, then the @start@ call will invoke the underlying @presence.enter()@ call.
+*** @(CHA-T4a2)@ @[Testable]@ When a typing session is started per @CHA-T4a1@, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session is automatically ended by leaving presence.
```
sacOO7, line 422:
```suggestion
*** @(CHA-T4a2)@ @[Testable]@ When a typing session is started per @CHA-T4a1@, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session automatically ends by leaving presence as per @CHA-T5c@.
```

sacOO7: Not sure, if we should have `internal retry mechanism` to stop sending typing event if `presence.leave` fails.
Since, we will not be delegating this error to the user : (

AndyTWF: Yes - this is a known bug that we intend to fix at a later date, I would prefer not to spec for it until we know what that is

sacOO7: if implicit `leave` event fails due to `typing timeout`, receivers will see eternal `typing` in progress : (

textile/chat-features.textile
```diff
@@ -413,12 +416,16 @@ Typing Indicators allow chat room users to indicate to others that they are typi
...
 ** @(CHA-T4b)@ @[Testable]@ If typing is already in progress, the @CHA-T3@ timeout is extended to be @timeoutMs@ from now.
 * @(CHA-T5)@ Users may indicate that they have stopped typing.
 ** @(CHA-T5a)@ @[Testable]@ If typing is not in progress, this operation is no-op.
-** @(CHA-T5b)@ @[Testable]@ If typing is in progress, he @CHA-T3@ timeout is cancelled. The client then leaves presence.
```
sacOO7, line 421: `CHA-T5b` is redefined, same as https://github.com/ably/specification/pull/249/files#r1867359230

textile/chat-features.textile
AndyTWF: Yes - it's a known bug. Once someone fixes it in a satisfactory way (in any of the SDKs), we'll update the spec - I just don't want to write a spec point for it until we have a tried and tested fix :)

sacOO7: If we are tracking it somewhere then good, otherwise you might like to create separate github issue for this WDYT

textile/chat-features.textile
AndyTWF: The spec point is the same - just changed the ordering

AndyTWF: Renamedas CHA-T5e

textile/chat-features.textile
sacOO7: Just an idea,
- All implicit ops that don't notify user about the error are retried for max_retries = 3 and then added to the internal queue of implicit failed operations.
- All those operations will be performed again once the connection goes `connected`.

textile/chat-features.textile
sacOO7: Should we mark `CHA-T5b` as removed ?

textile/chat-features.textile
sacOO7: Seems as per [RTP16b](https://sdk.ably.com/builds/ably/specification/main/features/#RTP16b), presence events will be queued at channel level when channel in `Initialized` or `Attaching` state.

sacOO7: For other states, operation will result in error as per [RTP16c](https://sdk.ably.com/builds/ably/specification/main/features/#RTP16c)

sacOO7: You might like to add those points to JIRA ticket, I will approve the PR for now, to unblock the spec

sacOO7 (approved):
LGTM. 
You can add some points from conversation into internal JIRA ticket

lawrence-forooghian:
@AndyTWF is this one still relevant after single channel? If so, would you mind rebasing on that branch, please?

AndyTWF:
> @AndyTWF is this one still relevant after single channel? If so, would you mind rebasing on that branch, please?

Its being resolved by the open API review RFC so I think we can close

PR #250, opened 2024-11-29
chat: mention newestFirst in history spec

AndyTWF:


textile/chat-features.textile
```diff
@@ -299,7 +299,7 @@ Broadly speaking, messages are published via REST calls to the Chat HTTP API and
 ** @(CHA-M5c)@ @[Testable]@ If a channel leaves the @ATTACHED@ state and then re-enters @ATTACHED@ with @resumed=false@, then it must be assumed that messages have been missed. The @subscription point@ of any subscribers must be reset to the @attachSerial@.
 ** @(CHA-M5d)@ @[Testable]@ If a channel @UPDATE@ event is received and @resumed=false@, then it must be assumed that messages have been missed. The @subscription point@ of any subscribers must be reset to the @attachSerial@.
 ** @(CHA-M5e)@ Each subscription shall expose a method or callback that allows for messages to be queried. These messages are queried via the "REST API"#rest-fetching-messages.
-** @(CHA-M5f)@ @[Testable]@ This method must accept any of the standard history query options, except for @direction@, which must always be @backwards@.
+** @(CHA-M5f)@ @[Testable]@ This method must accept any of the standard history query options, except for @direction@, which must always be @newestFirst@.
```
vladvelici, line 302: `orderBy`?

AndyTWF: Doh!

vladvelici (approved):

textile/chat-features.textile
```diff
@@ -299,7 +299,7 @@ Broadly speaking, messages are published via REST calls to the Chat HTTP API and
 ** @(CHA-M5c)@ @[Testable]@ If a channel leaves the @ATTACHED@ state and then re-enters @ATTACHED@ with @resumed=false@, then it must be assumed that messages have been missed. The @subscription point@ of any subscribers must be reset to the @attachSerial@.
 ** @(CHA-M5d)@ @[Testable]@ If a channel @UPDATE@ event is received and @resumed=false@, then it must be assumed that messages have been missed. The @subscription point@ of any subscribers must be reset to the @attachSerial@.
 ** @(CHA-M5e)@ Each subscription shall expose a method or callback that allows for messages to be queried. These messages are queried via the "REST API"#rest-fetching-messages.
-** @(CHA-M5f)@ @[Testable]@ This method must accept any of the standard history query options, except for @direction@, which must always be @backwards@.
+** @(CHA-M5f)@ @[Testable]@ This method must accept any of the standard history query options, except for @orderBy@, which must always be @newestFirst@.
```
sacOO7, line 302:
```suggestion
** @(CHA-M5f)@ @[Testable]@ This method can accept any of the standard history query options, except for @orderBy@, which must always be @newestFirst@.
```

AndyTWF: The original wording is correct here imo - it's saying the method must `accept` any of the standard arguments, i.e. its talking about the method signature. It would not be acceptable for the method not to accept the `limit` param, for example.

sacOO7: But, we also have default values for `QueryParam` arguments.
Okay maybe

```suggestion
** @(CHA-M5f)@ @[Testable]@ This method signature must have all history query options as arguments, except for @orderBy@, which must always be @newestFirst@.
```
This feels less confusing. Because there's a diff. in accepting the argument and defining it

sacOO7: This feels more explicit and there won't be implicit misunderstandings wdyt

AndyTWF: Downside with this wording is that it would be overly-prescriptive. This wording strongly requires that each history query option must be an individual arg to the function - that's not the case in JS (the history options are an object).

sacOO7: Okay, got it.
But, I would still recommend to make statement more explicit that fits across SDKs in general.
I will approve the PR for now 👍

sacOO7 (approved):

lawrence-forooghian:
This one looks pretty uncontentious — gonna resolve the comments, rebase, and merge 🙂

PR #259, opened 2024-12-18
Refactored RTN22

sacOO7:


textile/features.textile
```diff
@@ -502,7 +502,7 @@ h3(#realtime-connection). Connection
...
 ** @(RTN7d)@ If the @queueMessages@ client option (@TO3g@) has been set to @false@, then when a connection enters the @DISCONNECTED@ state, any messages which have not yet been @ACK@d should be considered to have failed, with the same effect as in @RTN7c@
-* @(RTN22)@ Ably can request that a connected client re-authenticates by sending the client an @AUTH@ @ProtocolMessage@. The client must then immediately start a new authentication process as described in "RTC8":#RTC8
+* @(RTN22)@ Ably can request that a connected client re-authenticates by sending the client an @AUTH@ @ProtocolMessage@. The client must then immediately start a new authentication process as described in "RTC8":#RTC8. Ably sends @AUTH@ @ProtocolMessage@ 30 seconds before the token expires. Note: If the token has a ttl less than 30 seconds, e.g. 5 seconds, ably won't be able to send @AUTH@ @ProtocolMessage@.
```
SimonWoolf, line 505: > Ably sends `AUTH` `ProtocolMessage` 30 seconds before the token expires. N(ote: If the token has a ttl less than 30 seconds, e.g. 5 seconds, ably won't be able to send `AUTH` `ProtocolMessage`

The spec specifies SDK behaviour. The added sentences don't have any requirements, they're just look like background information. What are SDK devs expected to do differently as a result of this information?

Like: currently, the server does not _guarantee_ that it sends an AUTH exactly 30 seconds before expiry, nor does it guarantee that it won't send one if you connect with a <30s token. Both things are at the server's discretion, and might change at any point. By putting this in the spec, we are saying that SDK devs should be able to rely on the current behaviour. Is that necessary? If so, for what purpose are they relying on it?

sacOO7: - I believe this won't affect core SDKs, but currently we added check as a part of [laravel token refresh PR](https://github.com/ably/laravel-broadcaster/pull/56/files)
- I had a hunch about value change, so the PR. Btw, few questions regarding this 
1.  How frequently do we change this value? or when was the last time value was changed ?
2. Currently, `laravel-echo` client is restricted to use [protocol 2 or ably-js@1.x](https://github.com/ably/laravel-broadcaster?tab=readme-ov-file#using-laravel-echo-on-client-side), can we still change the value from server side?

sacOO7: Okay current laravel impl. goes as follows 
1. Each requested token has default 1 hr expiry.
2. For each private/presence channel join, internally new jwt token is requested. As a part of request, it sends previous token. Server validates the previous token, checks if it is valid, and has previous private/presence channels. If it has, token is upgraded with the latest requested channel while maintaining old channels. While token is upgraded, [we set new token expiry same as old token](https://github.com/ably/laravel-broadcaster/blob/082d4e01b81048d33e8b2d04d998d6e70f3bb057/src/AblyBroadcaster.php#L218). So, after token expires ( default 1 hr expiry ), the client loses access to all channels along with their capabilities.
4. As a part of new token request after token expiry, although old token is provided ( since it's expired ), we return token with only `public` channels access. This means after token expiry, each channel will detach and will re-request token. This will cause presence enter/leave as customer explained -> https://ably-real-time.slack.com/archives/CURL4U2FP/p1733135454672739

sacOO7: We can upgrade token expiry to new value every time user joins a new channel, but then there won't be any point of having `token expiry`. So, wanted to make sure, we adhere to token expiry and it's not upgraded randomly.

sacOO7: Also, about adding statement `Ably sends @AUTH@ @ProtocolMessage@ 30 seconds before the token expires. Note: If the token has a ttl less than 30 seconds, e.g. 5 seconds, ably won't be able to send @AUTH@ @ProtocolMessage@.`, this is useful for testing purpose ( mainly integration tests ). But if you feel, value can change in the future, we can skip it 👍

SimonWoolf: > We can upgrade token expiry to new value every time user joins a new channel, but then there won't be any point of having token expiry

It seems to me that this claim contradicts the change you've made in [the PR](https://github.com/ably/laravel-broadcaster/pull/56/files).

If the point of having tokens expire is that the customer, via their auth server, must approve (i.e. have the ability to disapprove) every user's right to be in each private channel at least once per hour (or whatever period), then surely having the auth server automatically re-approve all permissions if the user's token is about to expire (without asking the customer) defeats that point.

(And if we don't believe that, if we think it's acceptable for the auth server to automatically re-issue the token with all its previous permissions if it's near expiry, then why not extend the expiry every time a new channel is added? But I don't think we do believe this; as you say that would defeat the point of token expiry).

Should the broadcaster not re-ask the customer to confirm permissions (using `verifyUserCanAccessChannel`) for each non-public channel in the token if the token is near expiry (which would be something like, on any auth request with < 10 minutes to go, so including channel access requests made towards the end, not just 30 seconds), and then it can extend the expiry time without breaking the auth model?

sacOO7: Okay, I will give a thought about this and will get back 👍

sacOO7: It seems sensible to verify for each non-public channel if the token is near expiry, before refreshing to new token with upgraded expiry. Will update PR accordingly in the future 👍

sacOO7: Added TODO for the same https://github.com/ably/laravel-broadcaster/pull/56#issuecomment-2550959155

PR #261, opened 2025-01-06
Port api reference to specification repo

SimonWoolf:
Context: https://github.com/ably/sdk-api-reference/pull/44#issuecomment-2573287969

Internal discussion with @m-hulbert : https://ably-real-time.slack.com/archives/C01LR5XFHFX/p1735928965057959

The existing sdk api reference repo was clearly not working well. The existence of this repo is not much known to anyone but the SDK team, who have been at some point instructed to update it in lockstep with the specification repo and have the deved team approve all PRs, which is no longer tenable given resource constraints on that team. The reason for it being its own repo instead of a file in the specification repo is lost to history.

There is certainly still value in having a file for shared docstrings between different sdks. But it should be a file in the specifications repo, where it can be versioned along with the spec, modified by the same pr that adds the same fields to the spec, and rendered to https://sdk.ably.com/ for the convenience of SDK devs.

This PR does that. (As a repo merge, preserving full history on both sides). After it is merged, I will archive the sdk-api-reference repo.

FYI @ttypic @lawrence-forooghian @owenpearson @sacOO7 @paddybyers


ttypic (commented):
Looks great! I totally agree with you, the only small problem I can see textile render isn't processing "/" correctly right now
image

SimonWoolf:
> isn't processing "/" correctly right now

looks like this is a 7 year old bug in redcarpet which still isn't fixed :/ https://github.com/vmg/redcarpet/issues/477

i'll figure something out

SimonWoolf:
switched from redcarpet to kramdown

Issue #265, opened 2025-01-10
[chat-spec][Room monitoring] Make user driven room operation result deterministic


sacOO7:
- Currently, when connection goes into `DISCONNECTED` state, RoomStatus is still `ATTACHED`.
- It remains `ATTACHED` till connection goes into either `CONNECTED`, `DISCONNECTED` or `FAILED` state.
- This allows room to perform all operations as per [CHA-RL9b](https://sdk.ably.com/builds/ably/specification/main/chat-features/#:~:text=(-,CHA%2DRL9b,-)%20%5BTestable%5D)
- But, operation doesn't guarantee proper result if `CONNECTION` goes into `SUSPENDED` state, since all room operation messages are queued at this point and connection keeps retrying indefinitely in `SUSPENDED` state till it gets `CONNECTED`.
- Link to internal conversation -> https://ably-real-time.slack.com/archives/C03JDBVM5MY/p1736419714906639?thread_ts=1736416097.802209&cid=C03JDBVM5MY

sacOO7:
Fix will need adding one more state to `Room` as `RoomStatus.DISCONNECTED`

sacOO7:
Emitting room event for the same will come with an additional property, `RetryRoomAttachIn=5`

sacOO7:
Currently, I have created PR to go with option 3 and it does seem feasible for now. 
I think we can keep it the same in the long term as well, provided it allows for deterministic user driven realtime ops.

PR #267, opened 2025-01-14
[chat-spec] Make user triggered ops deterministic in disconnected state

sacOO7:
- Closes #249
- Related to #265 

Note 

1. As per [3 proposed solutions](https://github.com/ably/specification/issues/265#issuecomment-2582492587), PR proposes a **Quick Fix 3rd Option**.
2. This is implemented as a part of the newly introduced `CHA-RL11a1`


lawrence-forooghian:
@sacOO7 I had a quick look at this one and it seems like most of the changes here (deduplication) are no longer needed after single-channel, in which only the presence feature now performs channel presence operations. Could you please confirm whether we can close this one?

sacOO7:
I think we can close this one. @AndyTWF is planning to redefine CHA-RL11 sometime in near future

PR #268, opened 2025-01-15
features/spec: add sandbox backward compatibility to endpoint spec

surminus:
In the initial implementation I wrote for ably-go I included backward compatibility to allow being able to simply specifying `sandbox` rather than having to specify `nonprod:sandbox`, and to ensure this compatibility is included in all client implementations, then I am including it in the spec.

Happy for suggestions on where it should go and wording, it seemed appropriate as a sub-clause.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -78,6 +78,7 @@ h2(#endpoint-configuration). Client library endpoint configuration
 *** @(REC1b2)@ If the @endpoint@ option is a domain name, determined by it containing at least one period (@.@) then the @primary domain@ is the value of the @endpoint@ option.
 *** @(REC1b3)@ Otherwise, if the @endpoint@ option is a non-production routing policy name of the form @nonprod:[name]@ then the @primary domain@ is @[name].realtime.ably-nonprod.net@.
 *** @(REC1b4)@ Otherwise, if the @endpoint@ option is a production routing policy name of the form @[name]@ then the @primary domain@ is @[name].realtime.ably.net@.
+**** @(REC1b4a)@ Unless the @endpoint@ option is a routing policy name of @sandbox@ then the @primary domain@ is @sandbox.realtime.ably-nonprod.net@ to maintain backward compatibility.
 ** @(REC1c)@ If the deprecated @environment@ option is specified then it defines a production routing policy name @[name]@:
 *** @(REC1c1)@ If any one of the deprecated options @restHost@, @realtimeHost@ are also specified then the options as a set are invalid.
 *** @(REC1c2)@ Otherwise, the primary domain is @[name].realtime.ably.net@.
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -78,6 +78,7 @@ h2(#endpoint-configuration). Client library endpoint configuration
 *** @(REC1b2)@ If the @endpoint@ option is a domain name, determined by it containing at least one period (@.@) then the @primary domain@ is the value of the @endpoint@ option.
 *** @(REC1b3)@ Otherwise, if the @endpoint@ option is a non-production routing policy name of the form @nonprod:[name]@ then the @primary domain@ is @[name].realtime.ably-nonprod.net@.
 *** @(REC1b4)@ Otherwise, if the @endpoint@ option is a production routing policy name of the form @[name]@ then the @primary domain@ is @[name].realtime.ably.net@.
+**** @(REC1b4a)@ Unless the @endpoint@ option is a routing policy name of @sandbox@ then the @primary domain@ is @sandbox.realtime.ably-nonprod.net@ to maintain backward compatibility.
```
SimonWoolf, line 81: This is kindof ambiguously phrased -- I'd usually parse "Unless X then Y" as "Y applies if !X". which here would mean, set the primary domain to sandbox.realtime.ably-nonprod.net unless endpoint="sandbox". (sure, with common sense it's clear it's not that, but not good practice)

I'd be tempted to put this as an REC1b5 between b3 and b4 at the same level, so it's part of the one if-else waterfall, saying "Otherwise, if the `endpoint` option is `sandbox` then the `primary domain` is `sandbox.realtime.ably.net`"? Or maybe modify REC1b3 to say "...a non-production routing policy name of the form `nonprod:[name]`, or 'sandbox', ...". wdyt?

sacOO7: Agreed, we should simplify phrasing the statement

sacOO7:
@surminus can you check why CI is failing? also can update PR according to comment

textile/features.textile
surminus: Agree with this and have updated.

surminus:
@sacOO7 I've updated it and fixed CI (it was due to duplicates)

sacOO7:
> @sacOO7 I've updated it and fixed CI (it was due to duplicates)

Great, I will take a look 👍 .
 I am also going through the relevant spec and the current ably-go implementation.

textile/features.textile
```diff
@@ -95,8 +96,9 @@ h2(#endpoint-configuration). Client library endpoint configuration
...
-*** @(REC2c5)@ Otherwise, if the @primary domain@ is determined to be a routing policy name via @(REC1c2)@ then the set of @fallback domains@ is @[name].a.fallback.ably-realtime.com@, @[name].b.fallback.ably-realtime.com@, @[name].c.fallback.ably-realtime.com@, @[name].d.fallback.ably-realtime.com@, @[name].e.fallback.ably-realtime.com@.
-*** @(REC2c6)@ Otherwise, if the @primary domain@ is determined by the deprecated @restHost@ or @realtimeHost@ option via @(REC1d)@ then the set of fallback domains is empty.
+*** @(REC2c5)@ Otherwise, if the routing policy name via @(REC1b5)@ is @sandbox@ then the set of @fallback domains@ is @sandbox.a.fallback.ably-realtime-nonprod.com@, @sandbox.b.fallback.ably-realtime-nonprod.com@, @sandbox.c.fallback.ably-realtime-nonprod.com@, @sandbox.d.fallback.ably-realtime-nonprod.com@, @sandbox.e.fallback.ably-realtime-nonprod.com@.
```
sacOO7, line 99: Same goes for `REC2c5`, it's exact interpretation of `REC2c3`, where if we set endpoint to `nonprod:sandbox`, we will get same behaviour. I don't think we need different spec for `sandbox` specific endpoint

sacOO7 (commented):
Ideally, spec points should focus on aspects directly related to the implementation of business logic in the code. They shouldn’t address test-specific details, as they don’t contribute to the business logic itself. In general, even during normal development, we avoid mixing test-specific implementations with the actual business logic. 

At the same time, I also feel this violates `REC1b4`, as it introduces the `sandbox` keyword as an exception to the spec point.

sacOO7:
Okay, [ endpoint ADR conclusion](https://ably.atlassian.net/wiki/spaces/ENG/pages/3428810778/ADR-119+ClientOptions+for+new+DNS+structure#Issues-and-options) specifies, 

`
SDKs, when updated to use target the new domains, should also have internal and test references to environment: ‘sandbox' updated to use the new option, ie endpoint: 'nonprod:sandbox';
`

This is not a breaking change, since this will be only for tests. Also, it says 

`
a temporary CNAME is added that resolves sandbox.realtime.ably.net to sandbox.realtime.ably-nonprod.net, (which can be removed once all references to environment: 'sandbox' have been removed).
`

That means specifying `sandbox` will also work as expected, for the time being, without need to have additional spec for the same.

surminus:
> a temporary CNAME is added that resolves sandbox.realtime.ably.net to sandbox.realtime.ably-nonprod.net, (which can be removed once all references to environment: 'sandbox' have been removed).

I actually missed this in the ADR but I really don't think we should do this, since implementing this on the infrastructure side will be complicated, since we have to consider SSL certificates etc, which reside in a different AWS account, and I would prefer not to do have to do that since it is **significantly** easier to build in backwards compatibility to the client itself, since it's a simple conditional. 

However, I am happy to remove the compatibility entirely and we can just agree that all tests are updated to use `nonprod:sandbox` with the release of the new client.

sacOO7:
Yeah, thanks! 
Actually, I was responsible for adding fallback support across SDKs before, and it’s definitely a tedious task that varies with each SDK. It's always good to have few spec points mention the required functionality to reduce implementation complexity.

PR #271, opened 2025-01-25
Fixed endpoint spec for REC1b1 and REC2c1

sacOO7:
- `REC1b1` forgot to mention one more deprecated clientOption `fallbackHostsUseDefault`.
- Updated `REC2c1` to use new default fallback hosts


surminus (approved):
Thanks!

SimonWoolf (approved):

PR #273, opened 2025-02-11
Add API for tracking usage of wrapper SDKs

lawrence-forooghian:
This specifies the API agreed in [ADR-117](https://ably.atlassian.net/wiki/spaces/ENG/pages/3413016589/ADR-117+Tracking+wrapper+SDK+usage+continued), to allow us to track usage of a "modern wrapper SDK" (the term used in that ADR to refer to a wrapper SDK which accepts an already-instantiated core SDK client). We will use it to track usage of the Chat SDK.

The functionality described here has already been partially implemented in ably-cocoa (https://github.com/ably/ably-cocoa/pull/2014) and is being used by ably-chat-swift (https://github.com/ably/ably-chat-swift/pull/211).

I found it quite hard to write a spec for this, and the language might still be both too vague and too convoluted. I'm putting this up for review as a first attempt in order not to end up getting stuck trying to write something perfect.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -138,6 +138,7 @@ h3(#restclient). RestClient
 *** @(RSC7d6)@ Libraries may offer a @ClientOptions#agents@ property, for use only by other Ably-authored SDKs, on a need-to-have basis.
 **** @(RSC7d6a)@ The product/version key-value pairs supplied to that property should be injected into all @Agent@ library identifiers emitted by connections made as a result of REST or Realtime instances created using those @ClientOptions@.
 **** @(RSC7d6b)@ An API commentary must be provided for this property. This commentary must make it clear that this interface is only to be used by Ably-authored SDKs.
+*** @(RSC7d7)@ To _attribute a given REST request to a given wrapper SDK_ means to, given a @WrapperSDKProxyOptions@ object describing the wrapper SDK, inject the product/version key-value pairs contained in these options' @agents@ property into the @Agent@ library identifier emitted for that REST request. "@WP6@":#WP6 defines the circumstances in which a REST request must be attributed to a wrapper SDK.
 * @(RSC18)@ If @ClientOptions#tls@ is true, then all communication is over HTTPS. If false, all communication is over HTTP however "Basic Auth":https://en.wikipedia.org/wiki/Basic_access_authentication over HTTP will result in an error as private keys cannot be submitted over an insecure connection. See @Auth@ below
 * @(RSC8)@ Supports two protocols:
 ** @(RSC8a)@ "MessagePack":https://msgpack.org/ binary protocol (this is the default for environments having a suitable level or support for binary data)
@@ -194,6 +195,10 @@ h3(#restclient). RestClient
 ** @(RSC22b)@ Returns an array of @BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>@s. Optionally, in languages where this is idiomatic, an overload may be implemented whereby the method can be called with a single @BatchPublishSpec@ and return a single @BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>@. This is not a feature of the REST API, whose response will still be an array, so if implementing this overload, the SDK will have to extract the element from the array.
 ** @(RSC23)@ This clause has been replaced by "@RSC24@":#RSC24. It was valid up to and including specification @2.1@.
 ** @(RSC24)@ @RestClient#batchPresence@ function takes an array of channel name strings and sends them as a comma separated string in the @channels@ query parameter in a GET request to @/presence@, returning a @BatchResult<BatchPresenceSuccessResult | BatchPresenceFailureResult>@ object.
+* @(RSC26)@ @RestClient#createWrapperSDKProxy@:
+** @(RSC26a)@ Returns a wrapper SDK proxy client whose underlying client is this @RestClient@. See "@WP1@":#WP1 for definitions of these terms.
+** @(RSC26b)@ Accepts a single argument, a @WrapperSDKProxyOptions@.
+** @(RSC26c)@ An API commentary must be provided for this method. This commentary must make it clear that this interface is only to be used by Ably-authored SDKs.
 
 h3(#rest-auth). Auth
 
@@ -464,6 +469,11 @@ h3(#realtimeclient). RealtimeClient
 ** @(RTC17a)@ Returns the current value of the @#clientId@ attribute of the @RealtimeClient@ object’s @#auth@ attribute
 * @(RTC10)@ The client library should never register any listeners for internal use with the public @EventEmitter@ interfaces (such as @Connection#on@) or message/event subscription interfaces (such as @RealtimeChannel#subscribe@) in such a way that a user of the library calling @Connection#off()@ or @RealtimeChannel#unsubscribe()@ to remove all listeners would result in the library not working as expected
 * @(RTC11)@ Unexpected internal exceptions, as defined in "@RSC20@":#RSC20, must be handled as gracefully as possible and reported to Ably's error reporting service when enabled. The aim when handling unexpected exceptions should be to ensure that no invalid or inconsistent state can potentially be left after handling the exception; depending on circumstances the remedial action could include failing the transport, failing the connection, rejecting a message, reinitialising the library completely, etc.
+* @(RTC14)@ @RealtimeClient#createWrapperSDKProxy@:
+** @(RTC14a)@ Returns a wrapper SDK proxy client whose underlying client is this @RealtimeClient@. See "@WP1@":#WP1 for definitions of these terms.
+** @(RTC14b)@ Accepts a single argument, a @WrapperSDKProxyOptions@.
+** @(RTC14c)@ An API commentary must be provided for this method. This commentary must make it clear that this interface is only to be used by Ably-authored SDKs.
+
 
 h3(#realtime-connection). Connection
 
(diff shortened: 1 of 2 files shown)


lawrence-forooghian:
Ah, need to add docstrings, will do that now.

textile/features.textile
```diff
@@ -899,6 +909,27 @@ h3(#backoff-jitter). Incremental backoff and jitter
...
+* @(WP3)@ This specification does not prescribe how a wrapper SDK proxy client should be implemented nor its concrete type. An implementation may choose to replace the return value of methods and properties with an implementation-defined type that differs from that returned by the underlying client; for example, a wrapper SDK proxy client's @channels@ property may return an object of some class which wraps the underlying @Channels@ instance and forwards all method calls to it. And, in turn, this type's @#get@ method may return an object of a class which wraps the underlying @RestChannel@ or @RealtimeChannel@ class and forwards all method calls to it.
+* @(WP4)@ A wrapper SDK proxy client should not provide a @createWrapperSDKProxy@ method.
+* @(WP5)@ A wrapper SDK proxy client does not have any state of its own; rather, it shares all state with its underlying client.
+* @(WP6)@ When the following methods are called via a wrapper SDK proxy client, then the resulting REST request must be attributed to the wrapper SDK (as defined in "@RSC7d7@":#RSC7d7), using the @WrapperSDKProxyOptions@ that were used to create the wrapper SDK proxy client:
+** @(WP6a)@ @RestChannel#publish@
```
AndyTWF, line 924: Do we need a specific list of methods? Would we not want to make it so for any methods that interact with the server?

textile/features.textile
```diff
@@ -899,6 +909,27 @@ h3(#backoff-jitter). Incremental backoff and jitter
 h3(#realtime-compatibility). Forwards compatibility
 * @(RTF1)@ The library must apply the "robustness principle":https://en.wikipedia.org/wiki/Robustness_principle in its processing of requests and responses with the Ably system. In particular, deserialization of ProtocolMessages and related types, and associated enums, must be tolerant to unrecognised attributes or enum values. Such unrecognised values must be ignored.
 
+h3(#wrapper-sdk-proxy). Wrapper SDK proxy client
+
+A _wrapper SDK_ is an Ably-authored non-core SDK.
+
+The core SDK provides an API for wrapper SDKs to supply Ably with analytics information that allows us track the usage of these SDKs.
```
AndyTWF, line 916: Something that we should perhaps consider here - right now, connections to the server are made by the underlying SDK, and because `autoconnect=true` we never see connections from the wrapper SDKs.

Obviously with autoconnect, that's not something we can realistically do when the option is turned on. But perhaps we should decide if we want to provide wrapper details in connections in any way, and also specify it?

textile/features.textile
```diff
@@ -899,6 +909,27 @@ h3(#backoff-jitter). Incremental backoff and jitter
...
+
```
AndyTWF, line 917: Another thing that comes to mind - in Chat we have a use-case whereby the SDK agents are changed "on the fly" (not part of the constructor). This is in the React case, whereby we set the agents when a client is passed to the React component.

Is this something that's supported here / needs further speccing?

textile/features.textile
lawrence-forooghian: So, in [my assumptions of the requirements in ADR-117](https://ably.atlassian.net/wiki/spaces/ENG/pages/3413016589/ADR-117+Tracking+wrapper+SDK+usage+continued#Requirements) I assumed that we did not want to attribute Realtime connection events to a wrapper SDK. This was based on the assumption that `autoconnect` would be used, though. But, if you look at [Matt's response to my question about what we'd like to track](https://ably.atlassian.net/wiki/spaces/ENG/pages/3413016589/ADR-117+Tracking+wrapper+SDK+usage+continued#Which-REST-requests-do-we-want-to-be-able-to-attribute-to-a-modern-wrapper-SDK%3F), he seems to believe that connections are not something we are interested in tracking.

textile/features.textile
lawrence-forooghian: That's not a use case that I was aware of. We'd probably want to have an API that allows you to mutate the `agents` of a wrapper SDK proxy client in that case. I don't know whether we should specify it here or implement it as a JS-only API.

textile/features.textile
lawrence-forooghian: The list here is based on [Matt saying in ADR-117 that we're interested in tracking billable operations](https://ably.atlassian.net/wiki/spaces/ENG/pages/3413016589/ADR-117+Tracking+wrapper+SDK+usage+continued#Which-REST-requests-do-we-want-to-be-able-to-attribute-to-a-modern-wrapper-SDK%3F).

I agree that it would be nice to try and write the spec here in a more general fashion, but I'm struggling to think of how we'd express it. Because initially we'd try and say something like "all REST requests initiated via a wrapper SDK proxy should be attributed to that wrapper SDK", but there are some REST requests where the meaning of "initiated via a wrapper SDK proxy" gets a bit muddy when the underlying client is not being used exclusively by a single wrapper SDK, because the requests happen as part of some common process that is difficult to attribute to a single user action, for example token requests or the REST calls involved in registering a device for push notifications. I guess we could explicitly exclude these REST calls from the "all REST requests initiated via a wrapper SDK proxy should be attributed to that wrapper SDK" spec point.

textile/features.textile
AndyTWF: Ah perfect, that's already been discussed :) Yeah I agree attachments are more valuable than just connections.

textile/features.textile
AndyTWF: It's probably worth specifying, if only with a comment that says it primarily pertains to React-like circumstances.

ttypic: Do we really want to mutate the agent? I’d rather think of a proxy as a decorator (those two GoF patterns are almost identical). Now we have `RealtimeClient` and `WithChatAgent(RealtimeClient)`, which powers `ChatClient`.

Can we use `WithReactAgent(ChatClient)` inside React internals? And what if `WithReactAgent(ChatClient)` is powered by `WithReactAgent(WithChatAgent(RealtimeClient))`?

Does that make sense?

lawrence-forooghian: Thinking about this some more, and I'm wondering whether what we could do instead in chat-js is, upon realising that we're operating in a React context, discard the initial wrapper SDK proxy client that we created upon instantiation of the chat client, and call `createWrapperSDKProxyClient` again to create a new wrapper SDK proxy client which includes the React agent. I assume this would require some changes to the JS chat SDK to allow the client that's being used internally to be switched out. Would this be an option?

textile/features.textile
ttypic: I think tracking connections is fundamentally incorrect in terms of analytics. The connection is initiated by `RealtimeClient`, so it shouldn’t be counted as Chat SDK usage. Otherwise, the wrapper SDK usage will obscure all `RealtimeClient` usage, leading to incorrect assumptions

textile/features.textile
```diff
@@ -899,6 +909,27 @@ h3(#backoff-jitter). Incremental backoff and jitter
...
+* @(WP2)@ A wrapper SDK proxy client must provide the same public API (that is, methods and properties) as its underlying client. Calling a method or accessing a property of a wrapper SDK proxy client should behave as if the same action had been performed directly on the underlying client, except for some modified usage tracking behaviour which is described in "@WP6@":#WP6 and "@WP7@":#WP7.
+* @(WP3)@ This specification does not prescribe how a wrapper SDK proxy client should be implemented nor its concrete type. An implementation may choose to replace the return value of methods and properties with an implementation-defined type that differs from that returned by the underlying client; for example, a wrapper SDK proxy client's @channels@ property may return an object of some class which wraps the underlying @Channels@ instance and forwards all method calls to it. And, in turn, this type's @#get@ method may return an object of a class which wraps the underlying @RestChannel@ or @RealtimeChannel@ class and forwards all method calls to it.
+* @(WP4)@ A wrapper SDK proxy client should not provide a @createWrapperSDKProxy@ method.
```
ttypic, line 921: Why do you think we should prevent this? I think with React-case it can be useful. Also if you can do this, I believe it shows that implementation has good design

textile/features.textile
AndyTWF: > I think tracking connections is fundamentally incorrect in terms of analytics. The connection is initiated by `RealtimeClient`, so it shouldn’t be counted as Chat SDK usage. Otherwise, the wrapper SDK usage will obscure all `RealtimeClient` usage, leading to incorrect assumptions

Yeah exactly, all of our current chat analytics are based on attachments + frontend API requests anyway

textile/features.textile
AndyTWF: > Thinking about this some more, and I'm wondering whether what we could do instead in chat-js is, upon realising that we're operating in a React context, discard the initial wrapper SDK proxy client that we created upon instantiation of the chat client, and call `createWrapperSDKProxyClient` again to create a new wrapper SDK proxy client which includes the React agent. I assume this would require some changes to the JS chat SDK to allow the client that's being used internally to be switched out. Would this be an option?

That could work, right now React calls an underlying method on the chat client to "add the agent". That method could just get a new wrapper instance. As getting a new wrapper instance still wraps the same realtime client, we can just replace the wrapper in the one place that'll need it.

textile/features.textile
lawrence-forooghian: I think that it could be hard to implement, because — at least in the ably-cocoa implementation — the wrapper proxy SDK relies on private API of the underlying client; that is, it wouldn't work if the underlying client only conformed to the public interface.

lawrence-forooghian: (I mean, we _could_ make the wrapper SDK proxy client and the public `RealtimeClient` class conform to some common private interface, but I'm not yet convinced it's worth it)

lawrence-forooghian: I've changed it to say

> A wrapper SDK proxy client does not need to provide a `createWrapperSDKProxy` method.

Would that be OK for you? And then individual platforms can choose to implement if it would be useful to them and not difficult to implement.

textile/features.textile
lawrence-forooghian: Can we resolve this conversation?

textile/features.textile
lawrence-forooghian: Can we resolve this conversation?

textile/features.textile
AndyTWF: Yep looks good!

textile/features.textile
AndyTWF: Yes

AndyTWF (approved):

textile/features.textile
ttypic: Yeah, sounds good to me

textile/features.textile
ttypic: Yeah, we can

ttypic (approved):

PR #275, opened 2025-02-12
fix TM2p: remove sentence about populating version from channelSerial

SimonWoolf:
this was my mistake; this doesn't work because channelSerial is regional, but version must be the origin serial, i.e. populated from the serial in the publishing region.

For now the server will just always populate version if it should be set (currently only in mutableMessages namespaces).

didn't bother to deprecate and replace, instead I'll just fix myself in every library this was implemented in.
- ably-js I've fixed in [the annotations support pr](https://github.com/ably/ably-js/pull/1953)
- ably-cocoa: https://github.com/ably/ably-cocoa/pull/2030
- ably-java: https://github.com/ably/ably-java/pull/1065

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1344,8 +1344,8 @@ h4. Message
 ** @(TM2i)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The @extras@ field is provided to contain message metadata and/or ancillary payloads in support of specific functionality, e.g. push. Each of these supported extensions is documented separately; for 1.1 the only supported extension is @push@, via the @extras.push@ member; 1.2 adds the @delta@ extension whose keys and values are described by the attributes of the type @DeltaExtras@, and the @headers@ extension, which contains arbitrary @string->string@ key-value pairs, settable at publish time, and @ref@ whose keys and values are described by the attributes of the type @ReferenceExtras@. Unless otherwise specified, the client library should not attempt to do any filtering or validation of the @extras@ field itself, but should treat it opaquely, encoding it and passing it to realtime unaltered.
 ** @(TM2f)@ @timestamp@ time in milliseconds since epoch. If a message received from Ably over a realtime transport does not contain a @timestamp@, the SDK must set it to the @timestamp@ of the encapsulating @ProtocolMessage@
 ** @(TM2j)@ @action@ enum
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -1344,8 +1344,8 @@ h4. Message
...
-** @(TM2k)@ @serial@ string - an opaque string that uniquely identifies the message. If a message received from Ably (whether over realtime or REST, eg history) with an @action@ of @MESSAGE_CREATE@ does not contain a @serial@, the SDK must set it equal to its @TM2p@ @version@.
+** @(TM2p)@ @version@ string - an opaque string that uniquely identifies the message, and is different for different versions. (May not be populated).
+** @(TM2k)@ @serial@ string - an opaque string that uniquely identifies the message. If a message received from Ably (whether over realtime or REST, eg history) with an @action@ of @MESSAGE_CREATE@ does not contain a @serial@ but does contain a @version@, the SDK must set it equal to the @TM2p@ @version@. (May not be populated).
```
sacOO7, line 1348: - `serial` field is always populated for edited/updated message right? 
- Ideally, it should be set to `serial`/`version` of the original message for all subsequent edits and deletes. I don't think we can populate it at client side using `version` for `edits` and `deletes`

sacOO7: Same goes for `createdAt`, it always represents value of original `message` -> `createdAt` for subsequent `edits` and `deletes`. That's how we wrote test assertions in `ably-java` as well

[Updated msg serial assertion](https://github.com/ably/ably-java/blob/fd12be945738e76b5fd78b066f8383e785be193a/lib/src/test/java/io/ably/lib/chat/ChatMessagesTest.java#L202) and 
[Deleted msg serial assertion](https://github.com/ably/ably-java/blob/fd12be945738e76b5fd78b066f8383e785be193a/lib/src/test/java/io/ably/lib/chat/ChatMessagesTest.java#L296)

sacOO7: chat spec [CHA-M10](https://sdk.ably.com/builds/ably/specification/main/chat-features/#chat-structs-message:~:text=strings%20are%20identical.-,(CHA%2DM10),-A%20Message%20can) mentions 
```
A Message can be modified by applying a new action to it, such as an update or delete. Applying an action must generate a new Message instance with the same serial and an updated version.
```
So, just want to make sure we populate `serial` for edited and updated messages since those are needed for l lexicographic sorting ( CHA-M10e )

SimonWoolf: > serial field is always populated for edited/updated message right?

yes

> Ideally, it should be set to serial/version of the original message for all subsequent edits and deletes. I don't think we can populate it at client side using version for edits and deletes

yep, that is why that field exists 🙂

sacOO7: Thanks for the clarification 👍

sacOO7, line 1348:
```suggestion
** @(TM2k)@ @serial@ string - an opaque string that uniquely identifies the message. If a message received from Ably (whether over realtime or REST, eg history) with an @action@ of @MESSAGE_CREATE@ does not contain a @serial@, the SDK must set it equal to the @TM2p@ @version@. (May not be populated).
```
Not sure if additional condition is needed, but we are good either way

SimonWoolf: I agree that the effect is the same either way, so you're right that the conditional is technically removable, but I think having it helps make the intent a bit clearer

PR #276, opened 2025-02-27
Add docstring for `Message.connectionKey`

lawrence-forooghian:
(We forgot to write one.)

### DIFF
--- a/textile/api-docstrings.md
+++ b/textile/api-docstrings.md
@@ -670,6 +670,7 @@ Contains an individual message that is sent to, or received from, Ably.
 | refType: String? ||| TM2m | If this message references another, the type of reference that is. |
 | createdAt: Time? ||| TM2o | The timestamp of the very first version of a given message (will differ from `timestamp` only if the message has been updated or deleted). |
 | operation: Operation? ||| TM2n | In the case of an updated or deleted message, this will contain metadata about the update or delete operation. |
+| connectionKey: String? ||| TM2h | Allows a REST client to publish a message on behalf of a Realtime client. If you set this to the [private connection key]{@link Connection.key} of a Realtime connection when publishing a message using a [`RestClient`]{@link RestClient}, the message will be published on behalf of that Realtime client. This property is only populated by a client performing a publish, and will never be populated on an inbound message. |
 
 ## class Operation
 


textile/api-docstrings.md
```diff
@@ -670,6 +670,7 @@ Contains an individual message that is sent to, or received from, Ably.
 | refType: String? ||| TM2m | If this message references another, the type of reference that is. |
 | createdAt: Time? ||| TM2o | The timestamp of the very first version of a given message (will differ from `timestamp` only if the message has been updated or deleted). |
 | operation: Operation? ||| TM2n | In the case of an updated or deleted message, this will contain metadata about the update or delete operation. |
+| connectionKey: String? ||| TM2h | Allows a REST client to publish a message on behalf of a Realtime client. If you set this to the [private connection key]{@link Connection.key} of a Realtime connection when publishing a message using a [`RestClient`]{@link RestClient}, the message will be published on behalf of that Realtime client. |
```
VeskeR, line 673: could also mention that:
"this is only ever populated by a publishing client when [publishing on behalf of another client](https://sdk.ably.com/rest/messages#publish-on-behalf)"
taken from https://sdk.ably.com/builds/ably/specification/main/features/#TM2h

lawrence-forooghian: What's the key information that you'd like me to add? I don't really want to repeat that it's part of the "publish on behalf of a Realtime client" feature. Is it the fact that this will never be populated on an inbound message?

VeskeR: > Is it the fact that this will never be populated on an inbound message?

yes, this is the important bit: never populated on inbound message, only ever sent by a client.
(ideally we would have something like `InboundMessage` and `OutboundMessage`, but that is a totally different discussion)

VeskeR (approved):
LGTM, thanks

textile/api-docstrings.md
lawrence-forooghian: Added "This property is only populated by a client performing a publish, and will never be populated on an inbound message."

I did propose adding such types to the spec in https://github.com/ably/specification/pull/156, but we didn't go ahead with it. We do have the `InboundMessage` type in ably-js, though (this TS-only type was the eventual result of the discussion in the aforementioned PR); I'll update the JS PR to omit this property from that type.

PR #277, opened 2025-02-28
Refactor: Spec updates for Typing presence to heartbeat-based changes

splindsay-92:
This Pull Request updates the specifications and behavior related to Typing Indicators in Chat.
It removes the old presence based typing system and introduces a new heartbeat-based system, which uses ephemeral messages over a typical WS publish. 

### Main Changes
  - Removed the spec points for the old presence based typing system.
  - Added spec points for the new heartbeat based system.
  - Added new configuration options for heartbeat interval properties.
  - Added new format for typing events that include both `change` and `currentlyTyping` fields.
  - Added  spec around use of ephemeral messages to ensure non-persistence in history.

### DIFF
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -425,8 +425,8 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 * @(CHA-T9)@ @[Testable]@ Users must be able to retrieve a list of the currently typing client IDs through a @typing.get()@ method. The list must be a copy of the up-to-date internal list of typing client IDs.
 * @(CHA-T15)@ @[Testable]@ Users must be able to provide a listener to subscribe to new typing events that the Chat client emits. This operation must have no side effects on the status of the room or the underlying realtime channel.
 * @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
-** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving events). The value shall be set to 2000ms.
-*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing and emit a @typing.stopped@ event for that user to all registered listeners @(CHA-T15)@.
+** @(CHA-T10a)@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving events). The default value shall be set to 2000ms.
+*** @(CHA-T10a1)@ This grace period is used to determine how long to wait, beyond the heartbeat interval, before removing a client from the typing set. This is used to prevent flickering when a user is typing and stops typing for a short period of time. See @CHA-T13b1@ for a detailed description of how this is used.
 * @(CHA-T3)@ This specification point has been removed. It was replaced by @CHA-T10@.
 * @(CHA-T4)@ Users may indicate that they have started typing using the @keystroke@ method.
 ** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
@@ -435,14 +435,16 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 *** @(CHA-T4a2)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
 *** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the @name@ field set to @typing.started@, the format of which is detailed "here":#realtime-typing-message.
 *** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a heartbeat timer shall be set according to the @CHA-T10@ timeout interval.
+*** @(CHA-T4a5)@ @[Testable]@ The client must wait for the publish to succeed or fail before returning the result to the caller. If the publish fails, the client must throw an @ErrorInfo@.
 ** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
 ** @(CHA-T4c)@ If typing is already in progress (i.e. a heartbeat timer set according to @CHA-T4a4@ exists and has not expired):
-*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
+*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event, instead it shall return a successful no-op to the caller.
 * @(CHA-T5)@ Users may explicitly indicate that they have stopped typing using @stop@ method.
 ** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer does not exist or is expired), this operation is a no-op.
 ** @(CHA-T5b)@ This specification point has been removed.
 ** @(CHA-T5c)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
-** @(CHA-T5d)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-message.
+** @(CHA-T5d)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-message. The client must wait for the publish to succeed or fail before returning the result to the caller.
+** @(CHA-T5d1)@ @[Testable]@ The client must wait for the publish to succeed or fail before returning the result to the caller. If the publish fails, the client must throw an @ErrorInfo@.
 ** @(CHA-T5e)@ @[Testable]@ On successfully publishing the message in @(CHA-T5d)@, the @CHA-T10@ timer shall be unset.
 * @(CHA-T6)@ Users may subscribe to typing events - updates to a set of clientIDs that are typing. This operation, like all subscription operations, has no side-effects in relation to room lifecycle.
 ** @(CHA-T6a)@ @[Testable]@ Users may provide a listener to subscribe to "typing event V2":#chat-structs-typing-event-V2 in a chat room.
@@ -455,13 +457,14 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 ** @(CHA-T13b)@ @[Testable]@ When a typing event is received on the realtime client:
 *** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the Chat client shall add the typers clientId (@msg.clientId@) to the typing set and a emit the updated set, via a new "typing event":#chat-structs-typing-event-V2 , to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
 *** @(CHA-T13b2)@ @[Testable]@ Each additional @typing.started@ event from the same client shall reset the @(CHA-T13b1)@ timeout for that client.
-*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the clientId from the typing set and emit a typing stop event for the given client to all registered listeners (@(CHA-T15)@).
-*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, the Chat client shall remove the clientId from the typing set and emit the updated set to any registered listeners (@(CHA-T15)@). It shall also cancel the @(CHA-T13b1)@ timeout for that typer.
-*** @(CHA-T13b5)@ @[Testable]@ If the event represents a client that has stopped typing, but the clientId for that client is not present in the typing set, then no event shall be emitted to any registered listeners (@(CHA-T15)@).
+*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the clientId from the typing set and emit a typing stop event for the given client to all registered listeners (@CHA-T15@).
+*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, the Chat client shall remove the clientId from the typing set and emit the updated set to any registered listeners (@CHA-T15@). It shall also cancel the @(CHA-T13b1)@ timeout for that typer.
+*** @(CHA-T13b5)@ @[Testable]@ If the event represents a client that has stopped typing, but the clientId for that client is not present in the typing set, then no event shall be emitted to any registered listeners (@CHA-T15@).
 * @(CHA-T14)@ @[Testable]@ Multiple asynchronous calls to keystroke/stop typing must eventually converge to a consistent state.
-** @(CHA-TM14a)@ When a call to @keystroke@ or @stop@ is made, it should attempt to acquire a mutex lock.
-** @(CHA-TM14b)@ Once the lock is acquired, if another call is made to either function, the second call shall be queued and wait until it can acquire the lock before executing.
-*** @(CHA-TM14b1)@ During this time, each new subsequent call to either function shall abort the previously queued call. In doing so, there shall only ever be one pending call and while the mutex is held, thus the most recent call shall "win" and execute once the mutex is released.
+** @(CHA-TM14a)@ @[Testable]@ When a call to @keystroke@ or @stop@ is made, it should attempt to acquire a mutex lock.
+** @(CHA-TM14b)@ @[Testable]@ Once the lock is acquired, if another call is made to either function, the second call shall be queued and wait until it can acquire the lock before executing.
+*** @(CHA-TM14b1)@ @[Testable]@ During this time, each new subsequent call to either function shall abort the previously queued call, this should be registered as a successful no-op to the caller. In doing so, there shall only ever be one pending call while the mutex is held, thus the most recent call shall "win" and execute once the mutex is released.
+*** @(CHA-TM14b2)@ @[Testable]@ If an error occurs while the lock is held, the mutex shall be released and the error shall be thrown to the caller.
 
 h2(#occupancy). Occupancy
 
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -438,7 +438,7 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 ** @(CHA-T4c)@ If typing is already in progress (i.e. a heartbeat timer set according to @CHA-T4a4@ exists and has not expired):
 *** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
 * @(CHA-T5)@ Users may explicitly indicate that they have stopped typing using @stop@ method.
-** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer exists and has not expired), this operation is a no-op.
+** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer does not exist or is expired), this operation is a no-op.
 ** @(CHA-T5b)@ This specification point has been removed.
 ** @(CHA-T5c)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
 ** @(CHA-T5d)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-heartbeat.
@@ -450,7 +450,7 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 *** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
 *** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
 * @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to the user.
-** @(CHA-T13a)@ @[Testable]@ If the typing event is malformed (unknown fields should be ignored), then it shall not be emitted to the subscriber.
+** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server is malformed, i.e does not conform to the expected "payload":#realtime-typing-events (unknown fields should be ignored), then it shall not be handled and emitted to the subscriber.
 ** @(CHA-T13b)@ @[Testable]@ When a typing event is received from the realtime client:
 *** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the chat client shall add the typer to the typing set and a emit the updated set to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
 *** @(CHA-T13b2)@ @[Testable]@ Each additional typing heartbeat from the same client shall reset the @(CHA-T13b1)@ timeout.
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -423,9 +423,9 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 * @(CHA-T2)@ This specification point has been removed. It was replaced by @CHA-T9@.
 * @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
 * @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
-* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided at the @RoomOptions.typing.heartbeatIntervalMs@ property, or idiomatic equivalent. The default is 15000ms.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
 ** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
-*** @(CHA-T10a1)@ @[Testable]@ If a heartbeat is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a heartbeat within 17000ms of the last heartbeat (15s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
+*** @(CHA-T10a1)@ @[Testable]@ If a heartbeat is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a heartbeat within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
 * @(CHA-T4)@ Users may indicate that they have started typing.
 ** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
 ** @(CHA-T4a)@ If typing is not already in progress (i.e. a @CHA-T10@ timeout has not been defined):
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -423,33 +423,25 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 * @(CHA-T2)@ This specification point has been removed. It was replaced by @CHA-T9@.
 * @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
 * @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
-* @(CHA-T3)@ @[Testable]@ Users may configure a typing timeout interval (the time after which the client will assume the user has stopped typing). This configuration is provided at the @RoomOptions.typing.timeoutMs@ property, or idiomatic equivalent. The default value is unset.
 * @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided at the @RoomOptions.typing.heartbeatIntervalMs@ property, or idiomatic equivalent. The default is 15000ms.
-* @(CHA-T11)@ @[Testable]@ Users may configure an inactivity timeout interval (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). This configuration is provided as part of the @RoomOptions@ @typing.inactivityTimeoutMs@ property, or idiomatic equivalent. The default is 500ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a heartbeat is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a heartbeat within 17000ms of the last heartbeat (15s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
 * @(CHA-T4)@ Users may indicate that they have started typing.
 ** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
 ** @(CHA-T4a)@ If typing is not already in progress (i.e. a @CHA-T10@ timeout has not been defined):
 *** @(CHA-T4a1)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
 *** @(CHA-T4a2)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
 *** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.started@, the format of which is detailed "here":#realtime-typing-heartbeat.
-*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, if a value for the @CHA-T3@ timeout interval is defined, a timeout shall be created.
-*** @(CHA-T4a5)@ @[Testable]@ Upon successful publish, a timeout shall be set according to the @CHA-T10@ timeout interval.
+*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a timeout shall be set according to the @CHA-T10@ timeout interval.
 ** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
 ** @(CHA-T4c)@ If typing is already in progress (i.e. a @CHA-T10@ timeout has been defined):
 *** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
-*** @(CHA-T4c2)@ @[Testable]@ If defined, the @CHA-T3@ timeout interval must be reset to @timeoutMs@ from now.
-* @(CHA-T12)@ Users may stop typing implicitly via lack of activity.
-** @(CHA-T12a)@ @[Testable]@ If a @CHA-T10@ timeout expires, then the @CHA-T10@ timeout must be unset. The client shall not send any messages on the channel.
-** @(CHA-T12b)@ @[Testable]@ If a @CHA-T3@ timeout expires, then the client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-heartbeat.
-*** @(CHA-T12b1)@ @[Testable]@ Upon successful publish, the @CHA-T10@ timeout must be unset.
-*** @(CHA-T12b2)@ @[Testable]@ Upon failed publish, the client must log this happening with the error as part of the log context and take no further action.
-*** @(CHA-T12b3)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, client must log this happening with the error as part of the log context and take no further action.
 * @(CHA-T5)@ Users may explicitly indicate that they have stopped typing.
 ** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. no @CHA-T10@ timeout has been defined), this operation is no-op.
 ** @(CHA-T5b)@ This specification point has been removed.
 ** @(CHA-T5c)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
 ** @(CHA-T5d)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-heartbeat.
-** @(CHA-T5e)@ @[Testable]@ On successfully publishing the message in @(CHA-T5d)@, the @CHA-T3@ and @CHA-T10@ timers shall be unset.
+** @(CHA-T5e)@ @[Testable]@ On successfully publishing the message in @(CHA-T5d)@, the @CHA-T10@ timer shall be unset.
 * @(CHA-T6)@ Users may subscribe to typing events - updates to a set of clientIDs that are typing. This operation, like all subscription operations, has no side-effects in relation to room lifecycle.
 ** @(CHA-T6a)@ @[Testable]@ Users may provide a listener to subscribe to "typing event V2":#chat-structs-typing-event-V2 in a chat room.
 ** @(CHA-T6b)@ @[Testable]@ A subscription to typing may be removed, after which it shall receive no further events.
@@ -459,7 +451,7 @@ For the purpose of this section, an @ephemeral message@ is understood to mean a 
 * @(CHA-T13)@ When a typing heartbeat is received from the realtime client, the Chat client shall emit appropriate events to the user.
 ** @(CHA-T13a)@ @[Testable]@ If the typing heartbeat is malformed (unknown fields should be ignored), then it shall not be emitted to the subscriber.
 ** @(CHA-T13b)@ @[Testable]@ When a typing heartbeat is received from the realtime client;
-*** @(CHA-T13b1)@ @[Testable]@ If the heartbeat represents a new client typing, then the chat client shall add the typer to the typing set and a emit the updated set to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T11@ grade period.
+*** @(CHA-T13b1)@ @[Testable]@ If the heartbeat represents a new client typing, then the chat client shall add the typer to the typing set and a emit the updated set to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ grade period.
 *** @(CHA-T13b2)@ @[Testable]@ Each additional typing heartbeat from the same client shall reset the @(CHA-T13b1)@ timeout.
 *** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the subject from the typing set and emitting the updated set to any subscribers.
 *** @(CHA-T13b4)@ @[Testable]@ If the heartbeat represents a client that has stopped typing, then the chat client shall remove the typer from the typing set and emit the updated set to any subscribers. It shall also cancel the @(CHA-T13b1)@ timeout for the typing client.
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -412,43 +412,42 @@ Presence allows chat room users to indicate to others that they're online, as we
 h2(#typing). Typing Indicators
 
 Typing Indicators allow chat room users to indicate to others that they are typing. This is most common in 1:1 and community chat, where the client would display "Alice and Bob are typing..." to users.
-@(deprecated)@ This system is presence-based, to avoid the cost of heartbeat/polling messages, with entering presence deemed to mean "started typing" and leaving presence deemed to mean "stopped typing".
 
-This system uses heartbeat messages, alongside specific message metadata (TBD) which disables persistence for them, reducing the cost of servicing the heartbeats.
+This system uses ephemeral non-persisted heartbeat messages. They do not appear in channel history, resume or rewind.
 
 @Typing Indicators@ shall be exposed to consumers via the @typing@ property of a @Room@.
 
-* @(CHA-T1)@ @(deprecated)@ Typing Indicators for a Room is exposed on a dedicated Realtime channel. These channels use the format @<roomId>::$chat::$typingIndicators@. For example, if your room id is @my-room@ then the typing channel will be @my-room::$chat::$typingIndicators@.
-* @(CHA-T2)@ @(deprecated)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs. The behaviour depends on the current room status, as presence operations in a Realtime Client cause implicit attaches.
-* @(CHA-T8)@ Typing Indicators for a Room is exposed on the channel used for chat messages, in the format @<roomId>::$chat::$chatMessages@. For example, if your room id is @my-room@ then the presence channel will be @my-room::$chat::$chatMessages@.
+For the purpose of this section, an @ephemeral message@ is understood to mean a regular message where @message.extras.ephemeral@ is set to @true@.
+
+* @(CHA-T1)@ This specification point has been removed. It was superseded by @CHA-T8@.
+* @(CHA-T2)@ This specification point has been removed. It was replaced by @CHA-T9@.
+* @(CHA-T8)@ Typing Indicators for a Room are exposed on the main rom channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
 * @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
-** @(CHA-T2b)@ This specification point has been removed.
-*** @(CHA-T2b1)@ This specification point has been removed.
-*** @(CHA-T2b2)@ This specification point has been removed.
-** @(CHA-T2c)@ @(deprecated)@ @[Testable]@ If the room status is @Attaching@, then the operation shall invoke a @CHA-RL9@ pending room status operation and wait for it to complete. If this operation fails, then the error from @CHA-RL9@ must be raised. If this succeeds (i.e the room becomes @ATTACHED@), the library must invoke the underlying @presence.get()@ call.
-** @(CHA-T2g)@ @(deprecated)@ @[Testable]@ If the room status is anything other than @Attached@ or @Attaching@, an @ErrorInfo@ using the @RoomInInvalidState@ error code from the "chat-specific error codes":#error-codes with a status code of @400@ must be thrown. The message shall explain that attach must be called first.
-** @(CHA-T2d)@ @(deprecated)@ @[Testable]@ If the room status is @Attached@, then the @get@ call shall invoke the underlying @presence.enter()@ call.
-
-** @(CHA-T2e)@ This specification point has been removed. It was superseded by @CHA-T2g@.
-** @(CHA-T2f)@ This specification point has been removed. It was superseded by @CHA-T2g@.
-* @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 5000ms.
-* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.heartbeatIntervalMs@ property, or idiomatic equivalent. The default is 15000ms.
-* @(CHA-T11)@ @[Testable]@ Users may configure an inactivity timeout interval for when they are receiving heartbeats from other users. This configuration is provided as part of the @RoomOptions@ @typing.inactivityTimeoutMs@ property, or idiomatic equivalent. The default is 2000ms.
+* @(CHA-T3)@ @[Testable]@ Users may configure a typing timeout interval (the time after which the client will assume the user has stopped typing). This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default value is unset.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided as part of the @RoomOptions@ @typing.heartbeatIntervalMs@ property, or idiomatic equivalent. The default is 15000ms.
+* @(CHA-T11)@ @[Testable]@ Users may configure an inactivity timeout interval (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). This configuration is provided as part of the @RoomOptions@ @typing.inactivityTimeoutMs@ property, or idiomatic equivalent. The default is 2000ms.
 * @(CHA-T4)@ Users may indicate that they have started typing.
-** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
-*** @(CHA-T4a1)@ @(deprecated)@ @[Testable]@ When a typing session is started, the client is entered into presence on the typing channel.
-*** @(CHA-T4a2)@ @(deprecated)@ @[Testable]@ When a typing session is started, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session is automatically ended by leaving presence.
-
-*** @(CHA-T4a3)@ @[Testable]@ When a typing session is started, the client shall publish a heartbeat message with the name field set to `typing.started`, the format of which is detailed "here":#realtime-typing-heartbeat.
-*** @(CHA-T4a4)@ @[Testable]@ When a typing session is started, a timeout shall be set according to the @CHA-T3@ timeout interval, if the interval is undefined, no timout shall be set. When this timeout expires, the client shall publish a heartbeat message with the name field set to `typing.stopped`, the format of which is detailed "here":#realtime-typing-heartbeat.
-*** @(CHA-T4a5)@ @[Testable]@ When a typing session is started, a timeout shall be set according to the @CHA-T10@ timeout interval. Until this timeout expires, the client shall not publish any further `typing.started` "heartbeats":#realtime-typing-heartbeat.
-
-** @(CHA-T4b)@ @[Testable]@ If typing is already in progress, the @CHA-T3@ timeout is extended to be @timeoutMs@ from now.
-** @(CHA-T4c)@ @[Testable]@ If typing is already in progress, the client shall not publish any further `typing.started` heartbeats until the @CHA-T10@ timeout expires. After expiration, if the client is still typing, it shall publish a `typing.started` "heartbeat":#realtime-typing-heartbeat and restart the @CHA-T10@ timeout.
-
-* @(CHA-T5)@ Users may indicate that they have stopped typing.
-** @(CHA-T5a)@ @[Testable]@ If typing is not in progress, this operation is no-op.
-** @(CHA-T5b)@ @(deprecated)@ @[Testable]@ If typing is in progress, the @CHA-T3@ timeout is cancelled. The client then leaves presence.
+** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
+** @(CHA-T4a)@ If typing is not already in progress (i.e. a @CHA-T10@ timeout has not been defined):
+*** @(CHA-T4a1)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a2)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.started@, the format of which is detailed "here":#realtime-typing-heartbeat.
+*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a timeout shall be set according to the @CHA-T3@ timeout interval, if the interval is undefined, no timeout shall be set. When this timeout expires, the client shall publish a heartbeat message with the name field set to `typing.stopped`, the format of which is detailed "here":#realtime-typing-heartbeat.
+*** @(CHA-T4a5)@ @[Testable]@ Upon successful publish, a timeout shall be set according to the @CHA-T10@ timeout interval. Until this timeout expires, the client shall not publish any further `typing.started` "heartbeats":#realtime-typing-heartbeat.
+** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
+** @(CHA-T4c)@ If typing is already in progress (i.e. a @CHA-T10@ timeout has been defined):
+*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
+*** @(CHA-T4c2)@ @[Testable]@ If defined, the @CHA-T3@ timeout interval must be reset to @timeoutMs@ from now.
+* @(CHA-T12)@ Users may stop typing implicitly via lack of activity.
+** @(CHA-T12a)@ @[Testable]@ If a @CHA-T10@ timeout expires, then the @CHA-T10@ timeout must be unset. The client shall not send any messages on the channel.
+** @(CHA-T12b)@ @[Testable]@ If a @CHA-T3@ timeout expires, then the client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-heartbeat.
+*** @(CHA-T12b1)@ @[Testable]@ Upon successful publish, the @CHA-T10@ timeout must be unset.
+*** @(CHA-T12b2)@ @[Testable]@ Upon failed publish, the client must log this happening with the error as part of the log context and take no further action.
+*** @(CHA-T12b3)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, client must log this happening with the error as part of the log context and take no further action.
+* @(CHA-T5)@ Users may explicitly indicate that they have stopped typing.
+** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. no @CHA-T10@ timeout has been defined), this operation is no-op.
+** @(CHA-T5b)@ This specification point has been removed.
+** @(CHA-T5c)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
 ** @(CHA-T5b)@ @(deprecated)@ @[Testable]@ If typing is in progress, the @CHA-T3@ and @CHA-T10@ timeouts shall be cancelled, and the client shall publish a heartbeat message with the name field set to `typing.stopped`.
 * @(CHA-T6)@ Users may subscribe to typing events - updates to a set of clientIDs that are typing. This operation, like all subscription operations, has no side-effects in relation to room lifecycle.
 ** @(CHA-T6a)@ @[Testable]@ Users may provide a listener to subscribe to "typing event V1":#chat-structs-typing-event in a room. Note, "typing event V1":#chat-structs-typing-event is a deprecated, please use "typing event V2":#chat-structs-typing-event-V2.
@@ -799,14 +798,13 @@ h3(#realtime-room-reactions). Ephemeral Room Reactions
   }
 </pre>
 
-h3(#realtime-typing-heartbeats). Typing Heartbeats
+h3(#realtime-typing-events). Typing Events
 
 <pre>
   {
     "name": "typing.started" | "typing.stopped"
     "encoding": "json"
-    "data": {}
-    },
+    "data": nil
     "timestamp": "1726232498871", // Only on incoming messages
     "extras": {}
   }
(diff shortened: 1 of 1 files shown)


textile/chat-features.textile
```diff
@@ -408,35 +408,51 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
-*** @(CHA-T4a1)@ @[Testable]@ When a typing session is started, the client is entered into presence on the typing channel.
-*** @(CHA-T4a2)@ @[Testable]@ When a typing session is started, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session is automatically ended by leaving presence.
-** @(CHA-T4b)@ @[Testable]@ If typing is already in progress, the @CHA-T3@ timeout is extended to be @timeoutMs@ from now.
-* @(CHA-T5)@ Users may indicate that they have stopped typing.
-** @(CHA-T5a)@ @[Testable]@ If typing is not in progress, this operation is no-op.
-** @(CHA-T5b)@ @[Testable]@ If typing is in progress, he @CHA-T3@ timeout is cancelled. The client then leaves presence.
+** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
+** @(CHA-T4a)@ If typing is not already in progress (i.e. a @CHA-T10@ timeout has not been defined):
```
sacOO7, line 428:
```suggestion
** @(CHA-T4a)@ If typing started without any previous timeout set (i.e. heartbeat using @CHA-T10@ timeout not set):
```

textile/chat-features.textile
```diff
@@ -408,35 +408,51 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T4a1)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a2)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.started@, the format of which is detailed "here":#realtime-typing-heartbeat.
+*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a timeout shall be set according to the @CHA-T10@ timeout interval.
```
sacOO7, line 432:
```suggestion
*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a heartbeat timer shall be set according to the @CHA-T10@ timeout interval.
```

textile/chat-features.textile
```diff
@@ -408,35 +408,51 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
+** @(CHA-T4c)@ If typing is already in progress (i.e. a @CHA-T10@ timeout has been defined):
```
sacOO7, line 434:
```suggestion
** @(CHA-T4c)@ If typing is already in progress (i.e. a heartbeat timer set according to CHA-T4a4):
```

textile/chat-features.textile
```diff
@@ -408,35 +408,51 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
+* @(CHA-T5)@ Users may explicitly indicate that they have stopped typing.
+** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. no @CHA-T10@ timeout has been defined), this operation is no-op.
```
sacOO7, line 437:
```suggestion
** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. heartbeat timer using @CHA-T10@ as a timeout not set), this operation is a no-op.
```

textile/chat-features.textile
```diff
@@ -408,35 +408,51 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-** @(CHA-T2g)@ @[Testable]@ If the room status is anything other than @Attached@ or @Attaching@, an @ErrorInfo@ using the @RoomInInvalidState@ error code from the "chat-specific error codes":#error-codes with a status code of @400@ must be thrown. The message shall explain that attach must be called first.
-** @(CHA-T2d)@ @[Testable]@ If the room status is @Attached@, then the @get@ call shall invoke the underlying @presence.enter()@ call.
-** @(CHA-T2e)@ This specification point has been removed. It was superseded by @CHA-T2g@.
-** @(CHA-T2f)@ This specification point has been removed. It was superseded by @CHA-T2g@.
-* @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 5000ms.
```
sacOO7, line 425: `CHA-T3 ` is not marked as removed  in updated spec

textile/chat-features.textile
```diff
@@ -408,35 +408,51 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a heartbeat is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a heartbeat within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
 * @(CHA-T4)@ Users may indicate that they have started typing.
-** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
```
sacOO7, line 427: `CHA-T4a` has been redefined, is it okay to do that?

sacOO7: Same goes for `CHA-T5a`

textile/chat-features.textile
```diff
@@ -774,6 +790,18 @@ h3(#realtime-room-reactions). Ephemeral Room Reactions
...
+h3(#realtime-typing-events). Typing Events
+
+<pre>
+  {
+    "name": "typing.started" | "typing.stopped"
+    "encoding": "json"
+    "data": nil
+    "timestamp": "1726232498871", // Only on incoming messages
+    "extras": {}
+  }
```
sacOO7, line 802:
```suggestion
  {
    "name": "typing.started" | "typing.stopped"
    "encoding": "json"
    "data": "clientId"
    "timestamp": "1726232498871", // Only on incoming messages
    "extras": { "ephemeral": true }
  }
```

textile/chat-features.textile
```diff
@@ -923,14 +951,26 @@ h3(#chat-structs-presence-event). Presence Event
...
     "currentlyTyping": ["clientId-1", "clientID-2"],
   }
 </pre>
 
+h3(#chat-structs-typing-event-V2). Typing Event V2
+
+<pre>
+  {
+    "change": {
+        "event": "typing.started",
```
sacOO7, line 967: Instead of `event` key, it is `type` in the DR => https://ably.atlassian.net/wiki/spaces/CHA/pages/3830513673/CHADR-093+Typing+Heartbeats+Ephemeral#Event-Changes
Which one should be preferred?

sacOO7 (commented):
can we update spec that mentions use of `keyStroke` and `stop` methods for typing.

textile/chat-features.textile
sacOO7: maybe this needs to be updated

textile/chat-features.textile
sacOO7: CHA-T5a needs to be updated/rephrased

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
-*** @(CHA-T4a1)@ @[Testable]@ When a typing session is started, the client is entered into presence on the typing channel.
-*** @(CHA-T4a2)@ @[Testable]@ When a typing session is started, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session is automatically ended by leaving presence.
-** @(CHA-T4b)@ @[Testable]@ If typing is already in progress, the @CHA-T3@ timeout is extended to be @timeoutMs@ from now.
-* @(CHA-T5)@ Users may indicate that they have stopped typing.
-** @(CHA-T5a)@ @[Testable]@ If typing is not in progress, this operation is no-op.
-** @(CHA-T5b)@ @[Testable]@ If typing is in progress, he @CHA-T3@ timeout is cancelled. The client then leaves presence.
+** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
+** @(CHA-T4a)@ If typing is not already in progress (i.e. a heartbeat timer using @CHA-T10@ as a timeout has not been defined):
```
sacOO7, line 428:
```suggestion
** @(CHA-T4a)@ If typing is not already in progress (i.e. a heartbeat timer not set on successful publish @CHA-T4a4@):
```

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
+* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a heartbeat is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a heartbeat within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
 * @(CHA-T4)@ Users may indicate that they have started typing.
```
sacOO7, line 426: Currently, we have 2 method for typing 
1. keystroke
2. stop
Both of those raise error if sending typing event fails. 
Ideally, typing event send is supposed to be an instantaneous process. 
So, if  `keystroke` and `stop` methods take more than 2 seconds to return a successful response, should we raise it as a timeout error? Ofc, we do check for channel attached state, so this case should be negligible, but just wanted to raise the point.

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T2)@ This specification point has been removed. It was replaced by @CHA-T9@.
+* @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
+* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a heartbeat is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a heartbeat within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
```
sacOO7, line 425:
```suggestion
*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
```

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-** @(CHA-T2f)@ This specification point has been removed. It was superseded by @CHA-T2g@.
-* @(CHA-T3)@ @[Testable]@ Users may configure a timeout interval for when they are typing. This configuration is provided as part of the @RoomOptions@ @typing.timeoutMs@ property, or idiomatic equivalent. The default is 5000ms.
+For the purpose of this section, an @ephemeral message@ is understood to mean a regular message where @message.extras.ephemeral@ is set to @true@.
+
+* @(CHA-T1)@ This specification point has been removed. It was superseded by @CHA-T8@.
+* @(CHA-T2)@ This specification point has been removed. It was replaced by @CHA-T9@.
+* @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
+* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the period between typing heartbeats sent by the client). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
```
sacOO7, line 423:
```suggestion
* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
```

textile/chat-features.textile
sacOO7:
```suggestion
* @(CHA-T4)@ Users may indicate that they have started typing using @keystroke@ method
```

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-*** @(CHA-T6c1)@ @[Testable]@ If the @presence.get()@ operation fails, then it shall be retried using a backoff with jitter, up to a timeout of 30 seconds.
-*** @(CHA-T6c2)@ @[Testable]@ If multiple presence events are received resulting in concurrent @presence.get()@ calls, then we guarantee that only the "latest" event is emitted. That is to say, if presence event A and B occur in that order, then only the typing event generated by B's call to @presence.get()@ will be emitted to typing subscribers.
+** @(CHA-T6c)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing heartbeat is received from the realtime client, the Chat client shall emit appropriate events to the user.
```
sacOO7, line 448:
```suggestion
* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to the user.
```

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T13a)@ @[Testable]@ If the typing heartbeat is malformed (unknown fields should be ignored), then it shall not be emitted to the subscriber.
```
sacOO7, line 449:
```suggestion
** @(CHA-T13a)@ @[Testable]@ If the typing event is malformed (unknown fields should be ignored), then it shall not be emitted to the subscriber.
```

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T13b)@ @[Testable]@ When a typing heartbeat is received from the realtime client;
+*** @(CHA-T13b1)@ @[Testable]@ If the heartbeat represents a new client typing, then the chat client shall add the typer to the typing set and a emit the updated set to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ grade period.
```
sacOO7, line 451:
```suggestion
*** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the chat client shall add the typer to the typing set and a emit the updated set to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ grade period.
```

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b2)@ @[Testable]@ Each additional typing heartbeat from the same client shall reset the @(CHA-T13b1)@ timeout.
+*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the subject from the typing set and emitting the updated set to any subscribers.
```
sacOO7, line 453:
```suggestion
*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the client from the typing set and emit typing stop event for the given client with an updated set.
```

textile/chat-features.textile
```diff
@@ -408,35 +408,52 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b4)@ @[Testable]@ If the heartbeat represents a client that has stopped typing, then the chat client shall remove the typer from the typing set and emit the updated set to any subscribers. It shall also cancel the @(CHA-T13b1)@ timeout for the typing client.
+*** @(CHA-T13b5)@ @[Testable]@ If the heartbeat represents a client that has stopped typing, but the client is not known to be currently typing, the event is ignored.
```
sacOO7, line 455:
```suggestion
*** @(CHA-T13b5)@ @[Testable]@ If the event represents that the client has stopped typing, but the client is not known to be currently typing, the event is ignored.
```

textile/chat-features.textile
```diff
@@ -411,35 +411,53 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the clientId from the typing set and emit typing stop event for the given client with an updated set.
+*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, then the chat client shall remove the clientId from the typing set and emit the updated set to any subscribers. It shall also cancel the @(CHA-T13b1)@ timeout for the typing client.
+*** @(CHA-T13b5)@ @[Testable]@ If the event represents that a client has stopped typing, but the clientId for that client is not present in the typing set, then the event is ignored.
+* @(CHA-T14)@ @[Testable]@ Handling multiple asynchronous calls to start/stop typing must converge to a consistent state. For example, a user begins typing, but publishing an event to the server takes some time, and the user calls @stop()@ typing before the first call completes. The client shall not send a @typing.stopped@ event and then a subsequent @typing.started@ event, it shall wait for the successful completion or failure of the first call before proceeding.
```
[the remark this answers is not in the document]
splindsay-92: I've tried to clear it up a bit, how does this sound?

```text
* @(CHA-T14)@ @[Testable]@ Multiple asynchronous calls to keystroke/stop typing must eventually converge to a consistent state.
** @(CHA-TM14a)@ When a call to @keystroke@ or @stop@ is made, it should attempt to acquire a mutex lock.
** @(CHA-TM14b)@ Once the lock is acquired, if another call is made to either function, the second call shall be queued and wait until it can acquire the lock before executing.
*** @(CHA-TM14b1)@ During this time, each new subsequent call to either function shall abort the previously queued call. In doing so, there shall only ever be one pending call and while the mutex is held, thus the most recent call shall "win" and execute once the mutex is released.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-* @(CHA-T7)@ @[Testable]@ Users may subscribe to discontinuity events to know when there's been a break in typing indicators. Their listener will be called when a discontinuity event is triggered from the room lifecycle. For typing, there shouldn't need to be user action as the underlying core SDK will heal the presence set.
+** @(CHA-T6c)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to the user.
+** @(CHA-T13a)@ @[Testable]@ If the typing event is malformed (unknown fields should be ignored), then it shall not be emitted to the subscriber.
```
sacOO7, line 453: Need to update this statement as to which fields will be malformed.
Current assumption is about `msg.name` which seems wrong, since we are specifically listening to the same field.
Check comment -> https://github.com/ably/ably-chat-kotlin/pull/117/files#r2015800350

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T4a)@ If typing is not already in progress (i.e. a heartbeat timer not set on successful publish @CHA-T4a4@):
+*** @(CHA-T4a1)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a2)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.started@, the format of which is detailed "here":#realtime-typing-heartbeat.
+*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a heartbeat timer shall be set according to the @CHA-T10@ timeout interval.
+** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
+** @(CHA-T4c)@ If typing is already in progress (i.e. a heartbeat timer set according to @CHA-T4a4@ exists and has not expired):
+*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
+* @(CHA-T5)@ Users may explicitly indicate that they have stopped typing using @stop@ method.
+** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer exists and has not expired), this operation is a no-op.
```
sacOO7, line 441:
```suggestion
** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer doesn't exist or expired), this operation is a no-op.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
+* @(CHA-T3)@ This specification point has been removed. It was replaced by @CHA-T10@.
+* @(CHA-T4)@ Users may indicate that they have started typing using the @keystroke@ method.
+** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
+** @(CHA-T4a)@ If typing is not already in progress (i.e. a heartbeat timer not set on successful publish @CHA-T4a4@):
```
sacOO7, line 432:
```suggestion
** @(CHA-T4a)@ This specification point has been removed. It was replaced by @CHA-T4c@.
```

sacOO7: No need to have a duplicate spec point. Or `CHA-T4c` should be moved here to specify condition of early return

sacOO7: `CHA-T4a` condition won't exist in the code and it's unnecessary duplication to have two statements with the same meaning in the spec.

textile/chat-features.textile
splindsay-92: This refers to the event received by the client over realtime, I'll clarify here. Also, I'll add a link to the realtime payload.

textile/chat-features.textile
splindsay-92: Good spot, thanks!

textile/chat-features.textile
sacOO7: Actually, we implemented early return like this -> https://github.com/ably/ably-chat-kotlin/blob/9b4eb93c3b9e1b008c5a511da896f11172ff8281/chat-android/src/main/java/com/ably/chat/Typing.kt#L235. So the else clause is implicit. You can check if you can simplify or keep as is. Just that, it felt duplication at first glance.

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-* @(CHA-T7)@ @[Testable]@ Users may subscribe to discontinuity events to know when there's been a break in typing indicators. Their listener will be called when a discontinuity event is triggered from the room lifecycle. For typing, there shouldn't need to be user action as the underlying core SDK will heal the presence set.
+** @(CHA-T6c)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to the user.
+** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server is malformed, i.e does not conform to the expected "payload":#realtime-typing-events (unknown fields should be ignored), then it shall not be handled and emitted to the subscriber.
```
sacOO7, line 453: Can we make this explicit, 
```suggestion
** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server has null or empty `msg.clientId`, then it shall not be handled and emitted to the subscriber.
```
This is the only valid we implemented as a part of kotlin typing. Seems other checks are not needed =>
https://github.com/ably/ably-chat-kotlin/blob/9b4eb93c3b9e1b008c5a511da896f11172ff8281/chat-android/src/main/java/com/ably/chat/Typing.kt#L196.
Check for `msg.name` is not needed, since we already register for `typing.started` and `typing.stopped`, so `msg.name` will always be either of those.
`encoding` will be dependent on underlying pub-sub SDK, `timestamp` and `extras` are not needed to be check wdyt

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
+* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
```
sacOO7, line 428:
```suggestion
*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms (10s heartbeat interval plus 2s grace period) since the last recorded @typing.start@ event, the client shall assume that the user has stopped typing.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to the user.
+** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server is malformed, i.e does not conform to the expected "payload":#realtime-typing-events (unknown fields should be ignored), then it shall not be handled and emitted to the subscriber.
+** @(CHA-T13b)@ @[Testable]@ When a typing event is received from the realtime client:
+*** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the chat client shall add the typer to the typing set and a emit the updated set to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
+*** @(CHA-T13b2)@ @[Testable]@ Each additional typing heartbeat from the same client shall reset the @(CHA-T13b1)@ timeout.
```
sacOO7, line 456:
```suggestion
*** @(CHA-T13b2)@ @[Testable]@ Each additional typing start event from the same client shall reset the @(CHA-T13b1)@ timeout.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T6c)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to the user.
+** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server is malformed, i.e does not conform to the expected "payload":#realtime-typing-events (unknown fields should be ignored), then it shall not be handled and emitted to the subscriber.
+** @(CHA-T13b)@ @[Testable]@ When a typing event is received from the realtime client:
+*** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the chat client shall add the typer to the typing set and a emit the updated set to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
```
sacOO7, line 455:
```suggestion
*** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the chat client shall add the typer to the typing set and emit the updated set to any subscribers. It shall also begin a timer with a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b2)@ @[Testable]@ Each additional typing heartbeat from the same client shall reset the @(CHA-T13b1)@ timeout.
+*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the clientId from the typing set and emit a synthetic typing stop event for the given client.
```
sacOO7, line 457:
```suggestion
*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timer expires, the client shall remove the clientId from the typing set and emit a locally synthesized typing stop event for the given client.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, then the chat client shall remove the clientId from the typing set and emit the updated set to any subscribers. It shall also cancel the @(CHA-T13b1)@ timeout for the typing client.
```
sacOO7, line 458:
```suggestion
*** @(CHA-T13b4)@ @[Testable]@ If the event represents that a client has stopped typing, then the chat client shall remove the clientId from the typing set and emit stop typing event with updated set to given subscribers. It shall also cancel the @(CHA-T13b1)@ timer for the given typing client.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b5)@ @[Testable]@ If the event represents that a client has stopped typing, but the clientId for that client is not present in the typing set, then the event is ignored.
```
sacOO7, line 459:
```suggestion
*** @(CHA-T13b5)@ @[Testable]@ If the event represents that a client has stopped typing, but the clientId for that client is not present in the typing set, then typing stop event is not emitted.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a heartbeat timer shall be set according to the @CHA-T10@ timeout interval.
+** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
+** @(CHA-T4c)@ If typing is already in progress (i.e. a heartbeat timer set according to @CHA-T4a4@ exists and has not expired):
+*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
+* @(CHA-T5)@ Users may explicitly indicate that they have stopped typing using @stop@ method.
+** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer does not exist or is expired), this operation is a no-op.
+** @(CHA-T5b)@ This specification point has been removed.
+** @(CHA-T5c)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
+** @(CHA-T5d)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-heartbeat.
```
sacOO7, line 444:
```suggestion
** @(CHA-T5d)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, data as @null@ or @empty@ the format of which is detailed "here":#realtime-typing-heartbeat.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
+* @(CHA-T3)@ This specification point has been removed. It was replaced by @CHA-T10@.
+* @(CHA-T4)@ Users may indicate that they have started typing using the @keystroke@ method.
+** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
+** @(CHA-T4a)@ If typing is not already in progress (i.e. a heartbeat timer not set on successful publish @CHA-T4a4@):
+*** @(CHA-T4a1)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a2)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.started@, the format of which is detailed "here":#realtime-typing-heartbeat.
```
sacOO7, line 435:
```suggestion
*** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.started@, data as @null@ or @empty@, the format of which is detailed "here":#realtime-typing-heartbeat.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
 
 h2(#typing). Typing Indicators
 
-Typing Indicators allow chat room users to indicate to others that they are typing. This is most common in 1:1 and community chat, where the client would display "Alice and Bob are typing..." to users. This system is presence-based, to avoid the cost of heartbeat/polling messages, with entering presence deemed to mean "started typing" and leaving presence deemed to mean "stopped typing".
+Typing Indicators allow chat room users to indicate to others that they are typing. This is most common in 1:1 and community chat, where the client would display "Alice and Bob are typing..." to users.
+
+This system uses ephemeral non-persisted heartbeat messages. They do not appear in channel history, resume or rewind.
```
sacOO7, line 416:
```suggestion
This system uses ephemeral non-persisted messages. They do not appear in channel history, resume or rewind.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+For the purpose of this section, an @ephemeral message@ is understood to mean a regular message where @message.extras.ephemeral@ is set to @true@.
+
+* @(CHA-T1)@ This specification point has been removed. It was superseded by @CHA-T8@.
+* @(CHA-T2)@ This specification point has been removed. It was replaced by @CHA-T9@.
+* @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
+* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
```
sacOO7, line 427:
```suggestion
** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving typing events to represent typing inactivity). The value shall be set to 2000ms.
```

textile/chat-features.textile
```diff
@@ -926,14 +958,26 @@ h3(#chat-structs-presence-event). Presence Event
...
+h3(#chat-structs-typing-event-V2). Typing Event V2
+
+<pre>
+  {
+    "change": {
+        "type": "typing.started",
+        "clientId": "clientId-2",
+    },
+    "currentlyTyping": ["clientId-1", "clientID-2"],
+  }
```
sacOO7, line 978:
```suggestion
  {
    "change": {
        "type": "typing.started",
        "clientId": "clientId-2",
    },
    "currentlyTyping": ["clientId-1", "clientId-2"],
  }
```

textile/chat-features.textile
```diff
@@ -777,6 +797,18 @@ h3(#realtime-room-reactions). Ephemeral Room Reactions
...
+
+<pre>
+  {
+    "name": "typing.started" | "typing.stopped"
+    "encoding": "json"
+    "clientId": "who-is-typing",
+    "timestamp": "1726232498871", // Only on incoming messages
+    "extras": { "ephemeral": true}
+  }
+</pre>
```
sacOO7, line 810:
```suggestion
h3(#realtime-typing-events). Typing Message - sent on @typing.keystroke@ and @typing.stop@

<pre>
  {
    "name": "typing.started" | "typing.stopped" // need to set
    "clientId": "who-is-typing", // automatically populated, no need to set
    "extras": { "ephemeral": true} // need to set
  }
</pre>
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
 
 h2(#typing). Typing Indicators
 
-Typing Indicators allow chat room users to indicate to others that they are typing. This is most common in 1:1 and community chat, where the client would display "Alice and Bob are typing..." to users. This system is presence-based, to avoid the cost of heartbeat/polling messages, with entering presence deemed to mean "started typing" and leaving presence deemed to mean "stopped typing".
+Typing Indicators allow chat room users to indicate to others that they are typing. This is most common in 1:1 and community chat, where the client would display "Alice and Bob are typing..." to users.
```
sacOO7, line 414:
```suggestion
Typing Indicators allow chat room users to indicate to others that they are typing. This is most common in 1:1 and community chat, where "Alice and Bob are typing..." message is displayed to users.
```

textile/chat-features.textile
sacOO7: Updated typing message format -> https://github.com/ably/specification/pull/277/files#r2019755781.

textile/chat-features.textile
sacOO7: typing-message-format -> https://github.com/ably/specification/pull/277/files#r2019755781

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-* @(CHA-T4)@ Users may indicate that they have started typing.
-** @(CHA-T4a)@ @[Testable]@ If typing is not already in progress, per explicit cancellation or the timeout interval in (@CHA-T3@), then a new typing session is started.
-*** @(CHA-T4a1)@ @[Testable]@ When a typing session is started, the client is entered into presence on the typing channel.
-*** @(CHA-T4a2)@ @[Testable]@ When a typing session is started, a timeout is set according to the @CHA-T3@ timeout interval. When this timeout expires, the typing session is automatically ended by leaving presence.
-** @(CHA-T4b)@ @[Testable]@ If typing is already in progress, the @CHA-T3@ timeout is extended to be @timeoutMs@ from now.
-* @(CHA-T5)@ Users may indicate that they have stopped typing.
-** @(CHA-T5a)@ @[Testable]@ If typing is not in progress, this operation is no-op.
-** @(CHA-T5b)@ @[Testable]@ If typing is in progress, he @CHA-T3@ timeout is cancelled. The client then leaves presence.
+For the purpose of this section, an @ephemeral message@ is understood to mean a regular message where @message.extras.ephemeral@ is set to @true@.
```
sacOO7, line 420:
```suggestion
For typing section, an @ephemeral message@ is understood to mean a regular message where @message.extras.ephemeral@ is set to @true@. i.e. "typing-message-format":#realtime-typing-message.
```

sacOO7: typing-message-format -> https://github.com/ably/specification/pull/277/files#r2019755781

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+
+* @(CHA-T1)@ This specification point has been removed. It was superseded by @CHA-T8@.
+* @(CHA-T2)@ This specification point has been removed. It was replaced by @CHA-T9@.
+* @(CHA-T8)@ Typing Indicators for a Room are exposed on the main room channel, in the format @<roomId>::$chat@. For example, if your room id is @my-room@ then the channel will be @my-room::$chat@.
+* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs.
```
sacOO7, line 425:
```suggestion
* @(CHA-T9)@ @[Testable]@ Users may retrieve a list of the currently typing client IDs using the @get@ method. This returns a copy of the internal @typing set@ (list to track currently typing users).
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving heartbeats). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing.
+* @(CHA-T3)@ This specification point has been removed. It was replaced by @CHA-T10@.
+* @(CHA-T4)@ Users may indicate that they have started typing using the @keystroke@ method.
+** @(CHA-T4d)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
```
sacOO7, line 431: Seems, this spec will be updated in the future as per internal slack discussion -> https://ably-real-time.slack.com/archives/C02NY1VT3LY/p1742981576970129

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T4a2)@ This specification point has been removed. It was replaced by @CHA-T4a3+@.
+*** @(CHA-T4a3)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.started@, the format of which is detailed "here":#realtime-typing-heartbeat.
+*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a heartbeat timer shall be set according to the @CHA-T10@ timeout interval.
+** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
+** @(CHA-T4c)@ If typing is already in progress (i.e. a heartbeat timer set according to @CHA-T4a4@ exists and has not expired):
+*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
+* @(CHA-T5)@ Users may explicitly indicate that they have stopped typing using @stop@ method.
+** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer does not exist or is expired), this operation is a no-op.
+** @(CHA-T5b)@ This specification point has been removed.
+** @(CHA-T5c)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
```
sacOO7, line 443: @(CHA-T5c) will be updated same as https://github.com/ably/specification/pull/277/files#r2019765235.
For now, we can ignore this.

textile/chat-features.textile
sacOO7: For, now we can ignore/ keep as is

textile/chat-features.textile
sacOO7, line 432:
```suggestion
** @(CHA-T4a)@ If typing is not already in progress (i.e. a heartbeat timer not yet set on successful publish @CHA-T4a4@):
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-*** @(CHA-T6c2)@ @[Testable]@ If multiple presence events are received resulting in concurrent @presence.get()@ calls, then we guarantee that only the "latest" event is emitted. That is to say, if presence event A and B occur in that order, then only the typing event generated by B's call to @presence.get()@ will be emitted to typing subscribers.
-* @(CHA-T7)@ @[Testable]@ Users may subscribe to discontinuity events to know when there's been a break in typing indicators. Their listener will be called when a discontinuity event is triggered from the room lifecycle. For typing, there shouldn't need to be user action as the underlying core SDK will heal the presence set.
+** @(CHA-T6c)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to the user.
```
sacOO7, line 452:
```suggestion
* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received on the realtime client, the chat client shall emit appropriate events to the user.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,56 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server is malformed, i.e does not conform to the expected "payload":#realtime-typing-events (unknown fields should be ignored), then it shall not be handled and emitted to the subscriber.
+** @(CHA-T13b)@ @[Testable]@ When a typing event is received from the realtime client:
```
sacOO7, line 454:
```suggestion
** @(CHA-T13b)@ @[Testable]@ When a typing event is received on the realtime client:
```

textile/chat-features.textile
splindsay-92: I don't think it's necessary to state data must be null or undefined, it can actually be anything, just typing won't handle it in any way.

textile/chat-features.textile
splindsay-92: I've modified this a bit, but removed `timestamp` and `encoding` still :)

textile/chat-features.textile
splindsay-92: Actually, it seems we have gone with the convention of capitalising `Chat` like we do with `Ably` :)

textile/chat-features.textile
splindsay-92: I've reworked this to state we should emit to registered listeners, and also defined a registered listener in a spec point.

textile/chat-features.textile
splindsay-92: I actually prefer the former, do you think it's not clear enough?

textile/chat-features.textile
```diff
@@ -411,36 +411,57 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving events). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing and emit a @typing.stopped@ event for that user to all registered listeners @(CHA-T15)@.
+* @(CHA-T3)@ This specification point has been removed. It was replaced by @CHA-T10@.
+* @(CHA-T4)@ Users may indicate that they have started typing using the @keystroke@ method.
```
lawrence-forooghian, line 431: It's not clear to me at what point the `keystroke` method returns and under what circumstances it throws errors.

- does it wait for the `CHA-T4a3` publish to complete, and how does it behave if that publish fails?
- is it a no-op if typing is in progress?

textile/chat-features.textile
```diff
@@ -411,36 +411,57 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T4a4)@ @[Testable]@ Upon successful publish, a heartbeat timer shall be set according to the @CHA-T10@ timeout interval.
+** @(CHA-T4b)@ This specification point has been removed. It was superseded by @CHA-T4c@.
+** @(CHA-T4c)@ If typing is already in progress (i.e. a heartbeat timer set according to @CHA-T4a4@ exists and has not expired):
+*** @(CHA-T4c1)@ @[Testable]@ The client must not send a @typing.started@ event.
+* @(CHA-T5)@ Users may explicitly indicate that they have stopped typing using @stop@ method.
+** @(CHA-T5a)@ @[Testable]@ If typing is not in progress (i.e. a @CHA-T10@ heartbeat timer does not exist or is expired), this operation is a no-op.
+** @(CHA-T5b)@ This specification point has been removed.
+** @(CHA-T5c)@ @[Testable]@ If the channel state is not @ATTACHED@ or @ATTACHING@, the operation shall throw an @ErrorInfo@ with code @40000@.
+** @(CHA-T5d)@ @[Testable]@ The client shall publish an ephemeral message to the channel with the name field set to @typing.stopped@, the format of which is detailed "here":#realtime-typing-message.
```
lawrence-forooghian, line 445: Similar question to that for `keystroke()` — does `stop()` wait for this publish to complete and how does it behave if the publish is unsuccessful?

textile/chat-features.textile
```diff
@@ -411,36 +411,57 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T15)@ @[Testable]@ Users must be able to provide a listener to subscribe to new typing events that the Chat client emits. This operation must have no side effects on the status of the room or the underlying realtime channel.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ @[Testable]@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving events). The value shall be set to 2000ms.
+*** @(CHA-T10a1)@ @[Testable]@ If a @typing.start@ event is not received within this period, the client shall assume that the user has stopped typing. For example, if the client has not received a @typing.start@ event within 12000ms of the last heartbeat (10s heartbeat interval plus 2s grace period), the client shall assume that the user has stopped typing and emit a @typing.stopped@ event for that user to all registered listeners @(CHA-T15)@.
```
lawrence-forooghian, line 429: It seems like the behaviour described here overlaps with that described in CHA-T13b1 - perhaps this spec point should just describe the existence of the grace period and give an overview of what it means, leaving the specification of the precise behaviour to CHA-T13b1, and making this one non-Testable.

textile/chat-features.textile
```diff
@@ -411,36 +411,57 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b2)@ @[Testable]@ Each additional @typing.started@ event from the same client shall reset the @(CHA-T13b1)@ timeout for that client.
+*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the clientId from the typing set and emit a typing stop event for the given client to all registered listeners (@(CHA-T15)@).
+*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, the Chat client shall remove the clientId from the typing set and emit the updated set to any registered listeners (@(CHA-T15)@). It shall also cancel the @(CHA-T13b1)@ timeout for that typer.
+*** @(CHA-T13b5)@ @[Testable]@ If the event represents a client that has stopped typing, but the clientId for that client is not present in the typing set, then no event shall be emitted to any registered listeners (@(CHA-T15)@).
+* @(CHA-T14)@ @[Testable]@ Multiple asynchronous calls to keystroke/stop typing must eventually converge to a consistent state.
+** @(CHA-TM14a)@ When a call to @keystroke@ or @stop@ is made, it should attempt to acquire a mutex lock.
```
lawrence-forooghian, line 462: Should these subpoints not also be Testable?

textile/chat-features.textile
splindsay-92: How about this?

```text
*** @(CHA-T10a1)@ This grace period is used to determine how long to wait, beyond the heartbeat interval, before removing a client from the typing set. This is used to prevent flickering when a user is typing and stops typing for a short period of time. See @CHA-T13b1@ for a detailed description of how this is used.
```

lawrence-forooghian: Sounds good!

textile/chat-features.textile
splindsay-92: I've added spec point specifically to mention that clients must wait for publish and throw an error info if it fails. I've also clarified in`@(CHA-T4c1)@` about no-ops, is that okay?

textile/chat-features.textile
splindsay-92: I've added a line as with the above comment on keystroke, so it details the client should wait for publish and throw an error info on failure.

textile/chat-features.textile
```diff
@@ -777,6 +801,16 @@ h3(#realtime-room-reactions). Ephemeral Room Reactions
...
 
+h3(#realtime-typing-message). Typing Message
+
+<pre>
+  {
+    "name": "typing.started" // Required, must be either "typing.started" or "typing.stopped"
+    "clientId": "who-is-typing", // Required, cannot be empty string or null
+    "extras": { "ephemeral": true } // Required, must be set to true
+  }
+</pre>
```
sacOO7, line 812:
```suggestion
<pre>
  {
    "name": "typing.started" // Required, must be either "typing.started" or "typing.stopped"
    "clientId": "who-is-typing", // Automatically set by pubsubclient, cannot be empty string or null when msg is received
    "extras": { "ephemeral": true } // Required, must be set to true
  }
</pre>
```
Required meaning can be interpreted as needs to set while sending typing event.
For guest user, pubsub client won't set it, but as a chat client, we don't need to set it. 
We should make this explicit somewhere wdyt

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-* @(CHA-T7)@ @[Testable]@ Users may subscribe to discontinuity events to know when there's been a break in typing indicators. Their listener will be called when a discontinuity event is triggered from the room lifecycle. For typing, there shouldn't need to be user action as the underlying core SDK will heal the presence set.
+** @(CHA-T6c)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to all registered listeners @(CHA-T15)@.
+** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server is malformed, i.e does not conform to the expected "payload":#realtime-typing-message (unknown fields should be ignored), then it shall not be handled and emitted to the subscriber.
```
sacOO7, line 456:
```suggestion
** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server has null/empty clientId, then it shall not be handled and emitted to the subscriber.
```

sacOO7: Spec point is easy to interpret if checks are explicit.

textile/chat-features.textile
sacOO7: Also, https://github.com/ably/specification/pull/277/files#r2021249000 will explicitly convey, null/empty clientId should be checked at the receiver side.

sacOO7 (commented):
Thanks for making suggested changes BTW 👍

textile/chat-features.textile
sacOO7: If we go by `required`, it's easy for someone to publish typing msg as 
```
val msg = Message(name = "typing.started", clientId = "custom-clientId", extras = { "ephemeral": true } }
```
instead of 
```
val msg = Message(name = "typing.started", extras = { "ephemeral": true } }
```

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
-*** @(CHA-T6c2)@ @[Testable]@ If multiple presence events are received resulting in concurrent @presence.get()@ calls, then we guarantee that only the "latest" event is emitted. That is to say, if presence event A and B occur in that order, then only the typing event generated by B's call to @presence.get()@ will be emitted to typing subscribers.
-* @(CHA-T7)@ @[Testable]@ Users may subscribe to discontinuity events to know when there's been a break in typing indicators. Their listener will be called when a discontinuity event is triggered from the room lifecycle. For typing, there shouldn't need to be user action as the underlying core SDK will heal the presence set.
+** @(CHA-T6c)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c1)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+*** @(CHA-T6c2)@ This specification point has been removed, it has been superseded by @CHA-T6d@.
+* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received from the realtime client, the Chat client shall emit appropriate events to all registered listeners @(CHA-T15)@.
```
sacOO7, line 455:
```suggestion
* @(CHA-T13)@ When a typing event (@typing.start@ or @typing.stop@) is received on the realtime client, the Chat client shall emit appropriate events to all registered listeners @(CHA-T15)@.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T13a)@ @[Testable]@ If the typing event received from the server is malformed, i.e does not conform to the expected "payload":#realtime-typing-message (unknown fields should be ignored), then it shall not be handled and emitted to the subscriber.
+** @(CHA-T13b)@ @[Testable]@ When a typing event is received on the realtime client:
+*** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the Chat client shall add the typers clientId (@msg.clientId@) to the typing set and a emit the updated set, via a new "typing event":#chat-structs-typing-event-V2 , to any subscribers. It shall also begin a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
```
sacOO7, line 458:
```suggestion
*** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the Chat client shall add the typers clientId (@msg.clientId@) to the typing set and a emit the updated set, via a new "typing event":#chat-structs-typing-event-V2 , to any subscribers. It shall also begin a timer with a timeout that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b2)@ @[Testable]@ Each additional @typing.started@ event from the same client shall reset the @(CHA-T13b1)@ timeout for that client.
```
sacOO7, line 459:
```suggestion
*** @(CHA-T13b2)@ @[Testable]@ Each additional @typing.started@ event from the same client shall reset the @(CHA-T13b1)@ timer for that client.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timeout expires, the client shall remove the clientId from the typing set and emit a typing stop event for the given client to all registered listeners (@CHA-T15@).
```
sacOO7, line 460:
```suggestion
*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timer expires, the client shall remove the clientId from the typing set and emit a typing stop event for the given client to all registered listeners (@CHA-T15@).
```

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, the Chat client shall remove the clientId from the typing set and emit the updated set to any registered listeners (@CHA-T15@). It shall also cancel the @(CHA-T13b1)@ timeout for that typer.
```
sacOO7, line 461:
```suggestion
*** @(CHA-T13b4)@ @[Testable]@ If the event represents that a client has stopped typing, then the Chat client shall remove the clientId from the typing set and emit a typing stop event with an updated set to any registered listeners (@CHA-T15@). It shall also cancel the @(CHA-T13b1)@ timer for the given typing client.
```

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+*** @(CHA-T13b5)@ @[Testable]@ If the event represents a client that has stopped typing, but the clientId for that client is not present in the typing set, then no event shall be emitted to any registered listeners (@CHA-T15@).
```
sacOO7, line 462:
```suggestion
*** @(CHA-T13b5)@ @[Testable]@ If the event represents a client that has stopped typing, but the clientId for that client is not present in the typing set, then no stop typing event shall be emitted to any registered listeners (@CHA-T15@).
```

textile/chat-features.textile
splindsay-92: Fair enough, I've removed `client`, but I've gone with the following as it makes it absolutely clear we are talking about a UI element. 
```text
Typing Indicators allow chat room users to indicate to others that they are typing. This is most common in 1:1 and community chat, where a UI would display somthing like "Alice and Bob are typing..." to users.
```

textile/chat-features.textile
splindsay-92: How about the following?
```text
{
    "name": "typing.started" // Required, must be either "typing.started" or "typing.stopped"
    "clientId": "who-is-typing", // Required, cannot be empty string or null - This is automatically set by the underlying pubsub SDK.
    "extras": { "ephemeral": true } // Required, must be set to true
  }
```
That way, if you're building this without an underlying pubsub client, it's clear that clientId must be sent. Otherwise, its still clear that this will automatically be set by the underlying client?

textile/chat-features.textile
splindsay-92: How about the following so it's clear it must have a `@clientId@` and conform to the expected payload?

```text
** @(CHA-T13a)@ @[Testable]@ Typing events received from the server must have a valid @clientId@ and must not be malformed, i.e they must conform to the expected "payload":#realtime-typing-message (unknown fields should be ignored). Invalid events shall not be handled or emitted to any subscribers. 
```

sacOO7: I mean that's still implicit and points to the gray area where something else needs to be checked. 
Why can't we keep it simple and explicit instead .. wdyt

textile/chat-features.textile
sacOO7: Sounds okay 👍

textile/chat-features.textile
lawrence-forooghian: Yep, looks good — please could you be more specific about the error in CHA-T4a5 and CHA-T5d1? Is there a reason for it not to just rethrow the publish error?

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+** @(CHA-T13b)@ @[Testable]@ When a typing event is received on the realtime client:
+*** @(CHA-T13b1)@ @[Testable]@ If the event represents a new client typing, then the Chat client shall add the typers @clientId@ to the typing set and a emit the updated set, via a new "typing event":#chat-structs-typing-event-V2 , to any listeners. It shall also begin a timer with a timeout period that is the sum of the @CHA-T10@ heartbeat interval and the @CHA-T10a@ graсe period.
+*** @(CHA-T13b2)@ @[Testable]@ Each additional @typing.started@ event from the same client shall reset the @(CHA-T13b1)@ timer for that client.
+*** @(CHA-T13b3)@ @[Testable]@ If the @(CHA-T13b1)@ timer expires, the client shall remove the @clientId@ from the typing set and emit a typing stopped event for the given client to all registered listeners (@CHA-T15@).
+*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, the Chat client shall remove the @clientId@ from the typing set and emit a typing stopped event to any registered listeners (@CHA-T15@). It shall also cancel the @(CHA-T13b1)@ timer for that the given typing client.
```
sacOO7, line 461:
```suggestion
*** @(CHA-T13b4)@ @[Testable]@ If the event represents a client that has stopped typing, the Chat client shall remove the @clientId@ from the typing set and emit a typing stopped event to any registered listeners (@CHA-T15@). It shall also cancel the @(CHA-T13b1)@ timer for the given typing client.
```

sacOO7 (approved):
LGTM ( you can check review comment for typo )

textile/chat-features.textile
```diff
@@ -411,36 +411,60 @@ Presence allows chat room users to indicate to others that they're online, as we
...
+* @(CHA-T15)@ @[Testable]@ Users must be able to provide a listener to subscribe to new typing events that the Chat client emits. This operation must have no side effects on the status of the room or the underlying realtime channel.
+* @(CHA-T10)@ @[Testable]@ Users may configure a heartbeat interval (the no-op period for @typing.keystroke@ when the heartbeat timer is set active at @CHA-T4a4@). This configuration is provided at the @RoomOptions.typing.heartbeatThrottleMs@ property, or idiomatic equivalent. The default is 10000ms.
+** @(CHA-T10a)@ A grace period shall be set by the client (the grace period on the @CHA-T10@ heartbeat interval when receiving events). The default value shall be set to 2000ms.
+*** @(CHA-T10a1)@ This grace period is used to determine how long to wait, beyond the heartbeat interval, before removing a client from the typing set. This is used to prevent flickering when a user is typing and stops typing for a short period of time. See @CHA-T13b1@ for a detailed description of how this is used.
```
sacOO7, line 429:
```suggestion
*** @(CHA-T10a1)@ This grace period is used to determine how long to wait ( since the last recorded received typing event ), beyond the heartbeat interval, before removing a client from the typing set. This is used to prevent flickering when a user is typing and stops typing for a short period of time. See @CHA-T13b1@ for a detailed description of how this is used.
```
You can rephrase accordingly, but main aim is to mention `grace` period is for received "typing start" events and `heartbeatThrottleMs` is set/reset when sending "typing start" events.

textile/chat-features.textile
sacOO7: Yeah, check chat-js typing keystroke publish, seems it throws error from publish as is. So, maybe we can modify both `CHA-T4a5` and `CHA-T5d1` accordingly.

splindsay-92: I've updated the comment, it now says we just re-throw the error to the caller :)

AndyTWF (approved):

PR #279, opened 2025-03-13
[PUB-925] Add new protocol message flags, actions, channel modes and message types for Objects

VeskeR:
Resolves [PUB-925](https://ably.atlassian.net/browse/PUB-925)

[PUB-925]: https://ably.atlassian.net/browse/PUB-925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+* @(TP5)@ The size of the @PresenceMessage@ for "TO3l8":#TO3l8 is calculated in the same way as for @Message@; see "TM6":#TM6
+
+h4. StateMessage
+
+* @(SM1)@ A @StateMessage@ represents an individual objects message to be sent or received via the Ably Realtime service
+* @(SM2)@ The attributes available in a @StateMessage@ are:
+** @(SM2a)@ @id@ string - unique ID for this objects message. This attribute is always populated for objects messages received over REST. For objects messages received over Realtime, if the objects message does not contain an @id@, it should be set to @protocolMsgId:index@, where @protocolMsgId@ is the id of the @ProtocolMessage@ encapsulating it, and @index@ is the index of the objects message inside the @state@ array of the @ProtocolMessage@
```
VeskeR, line 1421: The description for the id field was copy pasted from the identical for Message and PresenceMessage.
I would like to confirm this bit though:
> This attribute is always populated for objects messages received over REST
Will this hold true for objects message to??

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+** @(SM2b)@ @clientId@ string
+** @(SM2c)@ @connectionId@ string. If an objects message received from Ably does not contain a @connectionId@, it should be set to the @connectionId@ of the encapsulating @ProtocolMessage@
+** @(SM2d)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The @extras@ field is provided to contain message metadata and/or ancillary payloads in support of specific functionality. For 3.1 no specific functionality is specified for @extras@ in objects messages. Unless otherwise specified, the client library should not attempt to do any filtering or validation of the @extras@ field itself, but should treat it opaquely, encoding it and passing it to realtime unaltered
```
VeskeR, line 1424: > For 3.1 no specific functionality

I used `3.1` version here as I expect we add LiveObjects spec as part of the 3.1 protocol release. This was mentioned in [the comment in Jira](https://ably.atlassian.net/browse/PUB-925?focusedCommentId=61684).
Please note that this PR is based on `integration/liveobjects` branch that will contain all LiveObjects related spec additions, so we will have the opportunity to bump the protocol version before marging into main

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
 * @(TP4)@ @fromEncoded@ and @fromEncodedArray@ are alternative constructors that take an (already deserialized) @PresenceMessage@-like object (or array of such objects), and optionally a @channelOptions@, and return a @PresenceMessage@ (or array of such @PresenceMessages@) that's decoded and decrypted as specified in @RSL6@, using the cipher in the @channelOptions@ if the message is encrypted, with any residual transforms (ones that the library cannot decode or decrypt) left in the @encoding@ property per @RSL6b@. This is intended for users receiving messages other than from a REST or Realtime channel (for example, from a queue), to avoid them having to parse the @encoding@ string themselves. This behaviour is the same as in @(TM3)@.
+* @(TP5)@ The size of the @PresenceMessage@ for "TO3l8":#TO3l8 is calculated in the same way as for @Message@; see "TM6":#TM6
+
+h4. StateMessage
+
+* @(SM1)@ A @StateMessage@ represents an individual objects message to be sent or received via the Ably Realtime service
+* @(SM2)@ The attributes available in a @StateMessage@ are:
```
VeskeR, line 1420: Note that I didn't include `StateMessage.channel` property here to maintain parity with the `Message` and `PresenceMessage` definitions above.

For all these messages realtime defines a `channel` property, but spec omits it: https://github.com/ably/realtime/blob/2fcbe71939df11794e31b29536accf076acdabec/protocol/types.proto#L845

textile/features.textile
mschristensen: Yes

textile/features.textile
mschristensen: Sounds like that should be ticketed separately

SimonWoolf: > For all these messages realtime defines a channel property, but spec omits it

Correct.

There is no `channel` in a message or presencemessage in our normal Ably wire protocol. There is when using SSE, because in SSE messages don't arrive in a channel-specific context. But the spec is for how to write an Ably SDK, SSE-specific fields are irrelevant.

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+* @(SOP1)@ A @StateOperation@ describes an operation to be applied to an object on a channel
+* @(SOP2)@ @StateOperation@ @Action@ enum has the following values in order from zero: @MAP_CREATE@, @MAP_SET@, @MAP_REMOVE@, @COUNTER_CREATE@, @COUNTER_INC@, @OBJECT_DELETE@
+* @(SOP3)@ The attributes available in a @StateOperation@ are:
+** @(SOP3a)@ @action@ enum - defines the operation to be applied to the object
+** @(SOP3b)@ @objectId@ string - the object ID of the object on a channel to which the operation should be applied
+** @(SOP3c)@ @mapOp@ @MapOp@ object - the payload for the operation if it is an operation on a @Map@ object type
+** @(SOP3d)@ @counterOp@ @CounterOp@ object - the payload for the operation if it is an operation on a @Counter@ object type
+** @(SOP3e)@ @map@ @Map@ object - the payload for the operation if the operation is @MAP_CREATE@. Defines the initial value for the @Map@ object
+** @(SOP3f)@ @counter@ @Counter@ object - the payload for the operation if the operation is @COUNTER_CREATE@. Defines the initial value for the @Counter@ object
+** @(SOP3g)@ @nonce@ string - the nonce, must be present on objects create operations sent to the server
```
SimonWoolf, line 1448: > must be present on objects create operations sent to the server

bit unclear what an "objects create" operation is - maybe "must be present on `COUNTER_CREATE` and `MAP_CREATE` operations sent to the server"?

textile/features.textile
```diff
@@ -1422,6 +1556,9 @@ h4. ProtocolMessage
 ** @(TR3s)@ 18: @SUBSCRIBE@ Synonymous with @MESSAGE_SUBSCRIBE@. Retained for backward compatibility.
 ** @(TR3u)@ 18: @MESSAGE_SUBSCRIBE@ Synonymous with @SUBSCRIBE@. This variant should be preferred.
 ** @(TR3t)@ 19: @PRESENCE_SUBSCRIBE@
+** @(TR3v)@ 24: @OBJECT_SUBSCRIBE@
+** @(TR3w)@ 25: @OBJECT_PUBLISH@
+** @(TR3x)@ 26: @HAS_STATE@
```
SimonWoolf, line 1561: (this was correct when you raised this PR, but is no longer -- HAS_STATE has just been moved to position 7, since it's a flag not a mode)

textile/protocol.textile
```diff
@@ -76,12 +76,17 @@ The serial number of the message must be included in the @msgSerial@ field (in t
...
 
 - AUTH (17) := Sent by the client with a new token to reauthenticate the connection, with the connection either being closed due to incompatible token details being provided, or a @CONNECTED@ message being sent back to the client confirming the authentication succeeded. The server can request that the client authenticates by sending an @AUTH@ protocol message to the client, and the client must respond with a new token in an @AUTH@ protocol message.
 
 - ACTIVATE (18) := TODO.
 
+- STATE (19) := Indicates that the Protocol Message has a payload of one or more StateMessages associated with a single channel. @STATE@ messages may be sent in either direction. The channel associated with these messages is indicated in the channel field.<br><br>
+The serial number of the message must be included in the @msgSerial@ field (in the client &rarr; service direction) or in the @connectionSerial@ field (in the service &rarr; client direction). Further information on message serial numbering is given below.
```
SimonWoolf, line 86: > or in the @connectionSerial@ field

this is no longer true.

(I realise the other messages still have this text. this is because this document is extremely out of date, unfortunately)

SimonWoolf (approved):
I'm not super familiar with the state protocol, so just reviewing from the POV of well-formed-ness as a spec. From that POV this looks fine, nice job :+1:. couple minor nitpicks.

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+** @(SM2b)@ @clientId@ string
+** @(SM2c)@ @connectionId@ string. If an objects message received from Ably does not contain a @connectionId@, it should be set to the @connectionId@ of the encapsulating @ProtocolMessage@
+** @(SM2d)@ @extras@ JSON-encodable object, used to contain any arbitrary key value pairs which may also contain other primitive JSON types, JSON-encodable objects or JSON-encodable arrays. The @extras@ field is provided to contain message metadata and/or ancillary payloads in support of specific functionality. For 3.1 no specific functionality is specified for @extras@ in objects messages. Unless otherwise specified, the client library should not attempt to do any filtering or validation of the @extras@ field itself, but should treat it opaquely, encoding it and passing it to realtime unaltered
+** @(SM2e)@ @timestamp@ time in milliseconds since epoch. If an objects message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
+** @(SM2f)@ @operation@ @StateOperation@ object
+** @(SM2g)@ @object@ @StateObject@ object
```
mschristensen, line 1427: Might be worth flagging that only one of these (operation or object) is expected to be set on a given message, and that `object` is only set of the action is `STATE_SYNC`, and `operation` is only set if the action is `STATE`

textile/protocol.textile
```diff
@@ -76,12 +76,17 @@ The serial number of the message must be included in the @msgSerial@ field (in t
...
+
+- STATE_SYNC (20) := Currently reserved for us with objects synchronization following a channel @ATTACHED@ Protocol Message being received by the client. Once a channel becomes attached, the server will automatically send a snapshot of objects persisted on the channel to the client. Every @STATE_SYNC@ Protocol Message received will contain a channelSerial value and one or more StateMessages for each object currently persisted on the channel.  The channelSerial serves two purposes, it provides a way for the client library to resume a @STATE_SYNC@ should the transport be disconnected, and it also provides a means for the client library to know when the sync is complete.  A channelSerial that is being synced will contain an ID followed by a cursor after a colon such as "cf30e75054887:psl_7g:client:189", however the final page of @STATE_SYNC@ messages will have a serial with an empty cursor such as "cf30e75054887:".  A client can explicitly request a STATE_SYNC with an optional channelSerial; if no channelSerial is provided the server will send a complete snapshot of all objects on the channel; if a channelSerial is provided, the server will resume the @STATE_SYNC@ operation.
```
mschristensen, line 88: We should specify a valid example object ID in the cursor here:

```
map:3DYRjGoon2rfGav1VdWVruZ3pX6TQSt8UYYsmo6CqfY@1742208124981
```

Also since the object ID can include a colon `:`, we should specify that the cursor should be obtained by splitting on the _last_ colon. (Also - are we handling this correctly in the client?)

mschristensen (changes requested):
Suggested a couple of small changes & also echo Simon's comments, after that looks good

textile/protocol.textile
SimonWoolf: >  we should specify that the cursor should be obtained by splitting on the last colon.

you mean the _first_ colon, surely? the random sync id is guaranteed to not include a colon; the cursor (I assume you're using an object id as a cursor) might.

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+** @(SM2h)@ @serial@ string - the origin timeserial of this objects message
```
lawrence-forooghian, line 1428: Is "origin timeserial" a term that has a definition somewhere?

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+** @(SM2i)@ @siteCode@ string - the site code corresponding to this objects message's timeserial in the @serial@ property
```
lawrence-forooghian, line 1429: Is "site code" a term that has a definition somewhere?

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+* @(SOP2)@ @StateOperation@ @Action@ enum has the following values in order from zero: @MAP_CREATE@, @MAP_SET@, @MAP_REMOVE@, @COUNTER_CREATE@, @COUNTER_INC@, @OBJECT_DELETE@
+* @(SOP3)@ The attributes available in a @StateOperation@ are:
+** @(SOP3a)@ @action@ enum - defines the operation to be applied to the object
+** @(SOP3b)@ @objectId@ string - the object ID of the object on a channel to which the operation should be applied
+** @(SOP3c)@ @mapOp@ @MapOp@ object - the payload for the operation if it is an operation on a @Map@ object type
+** @(SOP3d)@ @counterOp@ @CounterOp@ object - the payload for the operation if it is an operation on a @Counter@ object type
+** @(SOP3e)@ @map@ @Map@ object - the payload for the operation if the operation is @MAP_CREATE@. Defines the initial value for the @Map@ object
+** @(SOP3f)@ @counter@ @Counter@ object - the payload for the operation if the operation is @COUNTER_CREATE@. Defines the initial value for the @Counter@ object
+** @(SOP3g)@ @nonce@ string - the nonce, must be present on objects create operations sent to the server
+** @(SOP3h)@ @initialValue@ binary - the initial value bytes for the object
```
lawrence-forooghian, line 1449: Do we have a spec point that defines how to encode and decode a binary property of a `ProtocolMessage`?

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+** @(ME3b)@ The size of the @data@ property is calculated per "SD3":#SD3
+** @(ME3c)@ The size of a @null@ or omitted property is zero
+
+h4. StateData
+
+* @(SD1)@ A @StateData@ represents a value in an object on a channel
+* @(SD2)@ The attributes available in a @StateData@ are:
+** @(SD2a)@ @objectId@ string - a reference to another object
+** @(SD2b)@ @encoding@ string - the encoding the client library should use to interpret the @value@
+** @(SD2c)@ @value@ string, number, boolean or buffer - a concrete value of the object
```
lawrence-forooghian, line 1536: We use the term "binary" elsewhere in this PR; and, same question about how to encode / decode over the wire

textile/features.textile
```diff
@@ -1406,11 +1412,139 @@ h4. PresenceMessage
...
+** @(ME3a)@ It is equal to the size of the @data@ property
+** @(ME3b)@ The size of the @data@ property is calculated per "SD3":#SD3
+** @(ME3c)@ The size of a @null@ or omitted property is zero
+
+h4. StateData
+
+* @(SD1)@ A @StateData@ represents a value in an object on a channel
+* @(SD2)@ The attributes available in a @StateData@ are:
+** @(SD2a)@ @objectId@ string - a reference to another object
+** @(SD2b)@ @encoding@ string - the encoding the client library should use to interpret the @value@
```
lawrence-forooghian, line 1535: How does the library use this?

textile/features.textile
VeskeR: > There is no channel in a message or presencemessage in our normal Ably wire protocol. There is when using SSE, because in SSE messages don't arrive in a channel-specific context.

Thank you, good to know

textile/features.textile
VeskeR: added suggestion in https://github.com/ably/specification/compare/92b5a16e33c154bba491a4c2ddd901212e15ce5a..f79593ce24b0877b39d7c89650a3f0ce7971fb37

textile/features.textile
VeskeR: thanks, changed in https://github.com/ably/specification/compare/92b5a16e33c154bba491a4c2ddd901212e15ce5a..f79593ce24b0877b39d7c89650a3f0ce7971fb37

textile/protocol.textile
VeskeR: removed it in https://github.com/ably/specification/compare/92b5a16e33c154bba491a4c2ddd901212e15ce5a..f79593ce24b0877b39d7c89650a3f0ce7971fb37

textile/features.textile
VeskeR: added description for this in https://github.com/ably/specification/compare/f79593ce24b0877b39d7c89650a3f0ce7971fb37..74e231937f838c5476ee8d523adf6ce1435e92f1

textile/protocol.textile
VeskeR: > We should specify a valid example object ID in the cursor here:

Didn't know object ID is used as a cursor (and now that I think about it, I've never actually seen a set cursor in the STATE_SYNC message so far, all objects have fit within a single STATE_SYNC).

Updated the cursor value and added a note about splitting at a first colon in https://github.com/ably/specification/compare/f79593ce24b0877b39d7c89650a3f0ce7971fb37..74e231937f838c5476ee8d523adf6ce1435e92f1. Wouldn't hurt to include the note here and also in the relevant spec point for the LiveObjects spec later.

textile/features.textile
VeskeR: No definition in the spec (and quite hard to find something reasonable in Confluence too).
However, current iteration of the client logic for the LiveObjects is not concerned if it's actually a "origin  timeserial" or another  timeserial (regional).
So I just removed the "origin" part in https://github.com/ably/specification/compare/74e231937f838c5476ee8d523adf6ce1435e92f1..7669d86f17d1a17d58cc236477df6084c2500e9a, it's just "timeserial" now.

textile/features.textile
VeskeR: There is no definition in the spec, and hard to find in Confluence - it’s primarily realtime knowledge of the internal structure of the timeserial. The timeserial string _contains_ a part referred to as the "site code."

However, the client SDK does not need to know that that value is a "site code" literally; rather, it simply uses this string value to update a vector of timeserials (same structure as [`StateObject.siteTimeserials`](https://github.com/ably/specification/pull/279/files#diff-d0e5a641ba4816766d38d6c0e7139df7502bed462dc676227057b4509cafad18R1464)) on LiveMaps/LiveCounters when operation is applied on that object. The exact details of the procedure are to be defined in the LiveObjects spec.

Not sure what the best approach is here regarding both `siteCode` and the [`StateObject.siteTimeserials`](https://github.com/ably/specification/pull/279/files#diff-d0e5a641ba4816766d38d6c0e7139df7502bed462dc676227057b4509cafad18R1464) field definitions.
Replacing "site code" with a generic "string" wouldn't sound right. Should we just drop the description for these altogether?

textile/features.textile
VeskeR: > We use the term "binary" elsewhere in this PR

Now this one is intereseting. Basically there are currently only two places in the spec where we use term "buffer": to describe `data` field for both `Message` and `PresenceMessage`, see [TM2d](https://sdk.ably.com/builds/ably/specification/main/features/#TM2d), and [TP3e](https://sdk.ably.com/builds/ably/specification/main/features/#TP3e).
I used "buffer" here to stay consistent with that similar definition of `data` for those messages.

In other places though, the spec currently uses term "binary". Funnily enough, [RSL4a](https://sdk.ably.com/builds/ably/specification/main/features/#RSL4a) speaks about message payloads too, but uses "binary" term.

I feel like "binary" is more generic, and it is currently mentioned 31 times in the spec compared to 2 times for "buffer".
So maybe we should just use "binary" everywhere, change it here, and in TM2d and TP3e. WDYT?

textile/features.textile
VeskeR: again , will be in the liveobjects spec, but it's the same purpose as [RSL4b](https://sdk.ably.com/builds/ably/specification/main/features/#RSL4b)

textile/protocol.textile
VeskeR: > splitting on the first colon. (Also - are we handling this correctly in the client?)

Yes, we use next regexp: `/^([\w-]+):(.*)$/`

textile/features.textile
SimonWoolf: neither originTimeserial nor timesSerial are defined in the spec because they are serverside concepts which should not be exposed to SDKs or customers, which should both treat the `serial` entirely opaquely, not caring if it's a timeserial or any other kind of identifier. (Similarly siteCode etc).

I suggest `A string uniquely identifying this StateMessage`.

textile/features.textile
```diff
@@ -1406,22 +1412,153 @@ h4. PresenceMessage
...
 ** @(TP3h)@ @memberKey@ string function that combines the @connectionId@ and @clientId@ ensuring multiple connected clients with the same clientId are uniquely identifiable
 * @(TP4)@ @fromEncoded@ and @fromEncodedArray@ are alternative constructors that take an (already deserialized) @PresenceMessage@-like object (or array of such objects), and optionally a @channelOptions@, and return a @PresenceMessage@ (or array of such @PresenceMessages@) that's decoded and decrypted as specified in @RSL6@, using the cipher in the @channelOptions@ if the message is encrypted, with any residual transforms (ones that the library cannot decode or decrypt) left in the @encoding@ property per @RSL6b@. This is intended for users receiving messages other than from a REST or Realtime channel (for example, from a queue), to avoid them having to parse the @encoding@ string themselves. This behaviour is the same as in @(TM3)@.
+* @(TP5)@ The size of the @PresenceMessage@ for "TO3l8":#TO3l8 is calculated in the same way as for @Message@; see "TM6":#TM6
+
+h4. StateMessage
+
+* @(SM1)@ A @StateMessage@ represents an individual objects message to be sent or received via the Ably Realtime service
```
SimonWoolf, line 1419: > `A @StateMessage@ represents an individual objects message`

This terminology back-and-forth between State and Objects seems like it's making things unnecessarily confusing. Are you planning to rename `StateMessage` to `ObjectsMessage`?

textile/features.textile
SimonWoolf: I'd also suggest not using the term `timeserial` in the spec or public API at all. Just call it a `serial`. `channelSerial` is now a timeserial, it used to be a different kind of serial. The public api didn't change at that time and nor did the spec; the exact implementation of serials we currently use is not part of the public api.

textile/features.textile
lawrence-forooghian: Yeah, let's switch to use "binary" everywhere 👍

textile/features.textile
VeskeR: thanks, I've updated description for `serial`, `siteCode`, `siteTimeserials` and `timeserial` fields this PR adds to be more generic and just treat them as opaque strings without exposing any serverside terminology.
see updated https://github.com/ably/specification/compare/7669d86f17d1a17d58cc236477df6084c2500e9a..92fb14098a1489e34accbcec38a63f797fdaab39

textile/features.textile
VeskeR: I've updated the description for `siteCode` in https://github.com/ably/specification/compare/7669d86f17d1a17d58cc236477df6084c2500e9a..92fb14098a1489e34accbcec38a63f797fdaab39.
Given that the exact purpose of the field (e.g. how and when the library needs to use it internally) will be covered in the LiveObjects spec, the description just gives a general idea that this field will be used as a key in some sort of a map

textile/features.textile
VeskeR: I've opened https://github.com/ably/specification/pull/284 to replace "buffer" -> "binary" for presence and message spec points, and updated this PR to use "binary" in https://github.com/ably/specification/compare/7669d86f17d1a17d58cc236477df6084c2500e9a..92fb14098a1489e34accbcec38a63f797fdaab39

textile/features.textile
VeskeR: > (also tbh the term "objects message" just looks and sounds weird, like you've missed an apostrophe and are talking about the message belonging to an object)

Yeah, I agree. Basically, I noticed that phrasing for chat and presence follows the pattern `[product] message/s` (e.g., `chat message`, `presence message`). Since the product is called "Objects", I used it too literally I think, so it ended up as `objects message`, which, as you pointed out, sounds a bit odd.

I've replaced it with just `object message/s` in https://github.com/ably/specification/compare/7669d86f17d1a17d58cc236477df6084c2500e9a..92fb14098a1489e34accbcec38a63f797fdaab39

VeskeR: > This terminology back-and-forth between State and Objects seems like it's making things unnecessarily confusing. Are you planning to rename StateMessage to ObjectsMessage?

I agree, and I would prefer that we standardize on `ObjectMessage`, the `HAS_OBJECT` flag, and the `OBJECT` and `OBJECT_SYNC` actions for the SDK spec/implementation.

A bit of background how we ended up with Objects and State:
The decision to rename "LiveObjects"/"State" to "Objects" in public APIs was made in [this DR](https://ably.atlassian.net/wiki/spaces/LOB/pages/3819896841/LODR-033+Isolating+API+naming+from+Product+naming+Renaming+state+and+liveobjects+to+objects+in+APIs).  

I believe to minimize the effort required for this transition, it was decided that internal system components (state coordinator, `state_` in C* table names, and state messages) would continue to use the term "state." As a result, internal definitions (`StateMessage` and related interfaces, the `HAS_STATE` flag, the `STATE` and `STATE_SYNC` actions) are still using "state" terminology (for example [ably-js implementation](https://github.com/ably/ably-js/blob/706d4ba64082c66af84e1ec3bf3f552e77902139/src/plugins/objects/statemessage.ts#L146) for StateMessage).

However, ssince these names are used only internally and are not exposed in realtime/SDK communication, we can easily transition to using "Object" at least in the SDK to be consistent and avoid dual terminology.

I believe it is best we make this change now, before we add things to the spec and while only ably-js SDK needs to be updated.

@mschristensen  WDYT? I remember we briefly discussed making this change on the SDK side but decided against it without giving it more thought.

mschristensen: The background is that we didn't want to do the work to rename internal implementation details, like `StateMessage` etc right now - we have only renamed things that are exposed in public APIs. I agree that we should align the client spec to use object terminology, as these just map to numeric identifiers.

textile/features.textile
```diff
@@ -1406,22 +1412,153 @@ h4. PresenceMessage
...
+** @(SM2e)@ @timestamp@ time in milliseconds since epoch. If an object message received from Ably does not contain a @timestamp@, it should be set to the @timestamp@ of the encapsulating @ProtocolMessage@
+** @(SM2f)@ @operation@ @StateOperation@ object. Mutually exclusive with the @object@ field. This field is only set on object messages if the @action@ field of the @ProtocolMessage@ encapsulating it is @STATE@
+** @(SM2g)@ @object@ @StateObject@ object. Mutually exclusive with the @operation@ field. This field is only set on object messages if the @action@ field of the @ProtocolMessage@ encapsulating it is @STATE_SYNC@
+** @(SM2h)@ @serial@ string - an opaque string that uniquely identifies this object message
+** @(SM2i)@ @siteCode@ string - an opaque string used as a key to update the map of @serial@ values on an object
```
mschristensen, line 1429: This should reference `SOB2b` `siteTimeserials` explicitly

mschristensen (approved):
Looks good, with a minor comment and pending the object renaming changes per https://github.com/ably/specification/pull/279#discussion_r2005069832

textile/features.textile
VeskeR: > I agree that we should align the client spec to use object terminology, as these just map to numeric identifiers (i.e. bit flags and numeric action)

Cool, will update it then

textile/features.textile
VeskeR: The `siteCode` won't be used to update the `StateObject.siteTimeserials` (SOB2b) field specifically.
`StateObject` from the `STATE_SYNC` message is used to populate the initial value for the live object in memory, and then  when operations are applied on the local object we update its data and a map of serial values on that object, not on the `StateObject` from the realtime message.
So I don't think referencing `SOB2b` is relevant here.

textile/features.textile
VeskeR: The only place left that is still using `state` name is the [ProtocolMessage.state](https://github.com/ably/specification/pull/279/files#diff-d0e5a641ba4816766d38d6c0e7139df7502bed462dc676227057b4509cafad18R1579) - an array of object messages on a protocol message received from Ably.

@mschristensen  Would you like to bite the bullet and rename it to `objects` on the realtime side so everything is nicely aligned, or would you prefer to leave it as is?

VeskeR: > Let's leave it as is, that's quite a big change

ok, then as far as this PR goes, all of the names have been changed to use "object" terminology

lawrence-forooghian (approved):

VeskeR:
Renamed `HAS_OBJECT` -> `HAS_OBJECTS` according to comment https://github.com/ably/ably-js/pull/1992#pullrequestreview-2729088599

PR #280, opened 2025-03-13
Add `ACTIVATE` ProtocolMessage Action

VeskeR:
[TR2](https://sdk.ably.com/builds/ably/specification/main/features/#TR2) reads as follows:

> ProtocolMessage Action enum has the following values **in order from zero**:

and then lists the enums values.
My intention is to preserve the wording of the TR2 and just add the new actions for LiveObjects: `STATE` and `STATE_SYNC` (https://github.com/ably/specification/pull/279).
And in order to do that, we need to first add the `ACTIVATE` action that is currently missing. Realtime enum: https://github.com/ably/realtime/blob/2fcbe71939df11794e31b29536accf076acdabec/protocol/common.proto#L139, and also added to ably-js even though not part of the spec: https://github.com/ably/ably-js/blob/a7d8181d62fcb2375cdaef3e86eb61c2fa349070/src/common/lib/types/protocolmessagecommon.ts#L23

@SimonWoolf @mschristensen Could you please briefly describe the `ACTIVATE` action so I can write a description for it for the [protocol page](https://github.com/ably/specification/pull/280/files#diff-970e185c37a49bd8aae9281d94475b7ba89a4699bbd07045e51642d39d264fcfR83)?

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1410,7 +1410,7 @@ h4. PresenceMessage
 h4. ProtocolMessage
 
 * @(TR1)@ A @ProtocolMessage@ represents the type used to send and receive messages over the Realtime protocol.  A ProtocolMessage always relates either to the connection or to a single channel only, but can contain multiple individual Messages or PresenceMessages.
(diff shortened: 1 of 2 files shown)


SimonWoolf:
> Could you please briefly describe the ACTIVATE action so I can write a description for it for the protocol page

it's deprecated, no-one should be implementing it now, so no reason to document. (it existed just to support ably-js transport upgrading, before v2 removed transport upgrading). I recommend "(reserved for a deprecated use)" or something

mschristensen (approved):

PR #281, opened 2025-03-14
fix: updated `RTN11` spec point

ttypic:
Link [(RTN11)](https://sdk.ably.com/builds/ably/specification/main/features/#RTN11)

- Removed the mention of `recoveryKey` as it has been removed

- Added `CLOSING` to the list of states that should trigger resetting all channels, as there should be no difference for the client whether they wait for the close event to occur or not. The sequence `client.close(); client.connect();` should be deterministic.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -525,9 +525,9 @@ h3(#realtime-connection). Connection
 * @(RTN10)@ This clause has been deleted. It was valid up to and including specification version @1.2@.
 * @(RTN11)@ @Connection#connect@ function:
 ** @(RTN11a)@ Explicitly connects to the Ably service if not already connected
-** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one.
+** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one. Additionally, the client should ensure that all channels first transition to @DETACHED@, following "@RTL3b@":#RTL3b, and then reinitialize channels per "@RTN11d@":#RTN11d.
 ** @(RTN11c)@ If the state is @DISCONNECTED@ or @SUSPENDED@, skips the ongoing wait in the retry process described in "RTN14d":#RTN14d and "RTN14e":#RTN14e, so that the next reconnection attempt happens immediately.
-** @(RTN11d)@ If the state is @CLOSED@ or @FAILED@, transitions all the channels to @INITIALIZED@ and unsets their @RealtimeChannel.errorReason@, unsets the @Connection.errorReason@, clears all connection state (including in particular @Connection.recoveryKey@), and resets the @msgSerial@ to @0@
+** @(RTN11d)@ If the state is @CLOSED@ or @FAILED@, transitions all the channels to @INITIALIZED@ and unsets their @RealtimeChannel.errorReason@, clear all internal connection data (including in particular @Connection.errorReason@) and resets the @msgSerial@ to @0@
 * @(RTN12)@ @Connection#close@ function:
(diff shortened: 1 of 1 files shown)


SimonWoolf (approved):
Looks like there's RTN11b above for the CLOSING state?

ttypic:
> Looks like there's RTN11b above for the CLOSING state?

Yes, I changed wording a little. Updated `CLOSING` branch

SimonWoolf (approved):
this part of the spec seems to be done a bit weirdly. usually the branch for each state contains the full list of actions for that state, but in this section previously RTN11a applies to all states, then the individual branches each had some additional actions. which isn't a pattern I'm a fan of. But.. shrug, I don't think anyone's going to be confused and you're not making it worse, so sure happy to approve

textile/features.textile
```diff
@@ -525,9 +525,9 @@ h3(#realtime-connection). Connection
 * @(RTN10)@ This clause has been deleted. It was valid up to and including specification version @1.2@.
 * @(RTN11)@ @Connection#connect@ function:
 ** @(RTN11a)@ Explicitly connects to the Ably service if not already connected
-** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one.
+** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one. Additionally, the client should ensure that all channels first transition to @DETACHED@, following "@RTL3b@":#RTL3b, and then reinitialize channels per "@RTN11d@":#RTN11d.
```
sacOO7, line 528:
```suggestion
** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one and then reinitialize channels per "@RTN11d@":#RTN11d.
```
Not sure if we need to make sure channels go detached. If we are clearing `recoveryKey` and resetting `msgSerial` then there shouldn't be issue about detaching channels.
The latest ably-java PR https://github.com/ably/ably-java/pull/1077/files, doesn't seem to detach channels.

ttypic: It detaches channels, there are even tests for that. The reason why I added this is because state change from attached to initialized is weird and never mentioned in the spec, we should either rework the spec and ensure we cover this branch (and now we don't cover at least for presence, we should definitely clear the presence if we closed client and reconnected) or make it explicit detach before reconnection

sacOO7: Yeah, seems like we do channel detach when connection close message is received

textile/features.textile
```diff
@@ -525,9 +525,9 @@ h3(#realtime-connection). Connection
...
 ** @(RTN11c)@ If the state is @DISCONNECTED@ or @SUSPENDED@, skips the ongoing wait in the retry process described in "RTN14d":#RTN14d and "RTN14e":#RTN14e, so that the next reconnection attempt happens immediately.
-** @(RTN11d)@ If the state is @CLOSED@ or @FAILED@, transitions all the channels to @INITIALIZED@ and unsets their @RealtimeChannel.errorReason@, unsets the @Connection.errorReason@, clears all connection state (including in particular @Connection.recoveryKey@), and resets the @msgSerial@ to @0@
+** @(RTN11d)@ If the state is @CLOSED@ or @FAILED@, transitions all the channels to @INITIALIZED@ and unsets their @RealtimeChannel.errorReason@, unsets the @Connection.errorReason@, clears all connection state and resets the @msgSerial@ to @0@
```
sacOO7, line 530: I'm not sure what it means by `clears all connection state`

sacOO7: I guess, it should go `CONNECTING` and then `CONNECTED`

textile/features.textile
sacOO7, line 528:
```suggestion
** @(RTN11b)@ If the state is @CLOSING@, the client should make a new connection with a new transport instance and remove all references to the old one. In particular, it should make sure that, when the @CLOSED@ @ProtocolMessage@ arrives for the old connection, it doesn't affect the new one. Additionally, the client in `CONNECTING` state should ensure that all channels first transition to @DETACHED@, following "@RTL3b@":#RTL3b, and then reinitialize channels per "@RTN11d@":#RTN11d.
```

sacOO7 (approved):
LGTM

PR #283, opened 2025-03-18
Restore path in `#time` spec point

lawrence-forooghian:
I guess this was unintentionally removed in 27a1696.


paddybyers (approved):
LGTM, thanks

PR #291, opened 2025-04-07
Make RTN15i only apply to connection-level `ERROR`

### LINKED ISSUE #290: RTN15i should only apply to connection-level `ERROR`
RTN15i:

> If an ERROR ProtocolMessage is received, this indicates a fatal error in the connection. The server will close the transport immediately after. The client should transition to the FAILED state triggering all attached channels to transition to the FAILED state as well. Additionally the Connection#errorReason should be set with the error received from Ably

But RTL14 says that some `ERROR`s should be handled by the channel, not the connection:

> If an ERROR ProtocolMessage is received for this channel (the channel attribute matches this channel’s name), then the channel should immediately transition to the FAILED state, and the RealtimeChannel.errorReason should be set

I think that RTN15i should have the same "with an empty `channel` attribute" as RTN14b/g.

lawrence-forooghian:
For consistency with RTL14 (which says that some `ERROR`s should be handled by the channel, not the connection), and RTN14b/g.

Resolves #290.


paddybyers (approved):
LGTM, thanks

PR #292, opened 2025-04-14
Support annotations

SimonWoolf:
Spec equivalent of https://github.com/ably/ably-js/pull/1953


sacOO7:
Wow, big PR, will take some time to review it properly 👍 
Thanks for adding !

textile/features.textile
```diff
@@ -419,6 +420,25 @@ h3(#rest-encryption). Encryption
...
+h3(#rest-annotations). RestAnnotations
+
+* @(RSAN1)@ @RestAnnotations#publish@ function:
+** @(RSAN1a)@ Takes a first argument of a @Message@ identifier in the form of the @messageSerial@ string of the message to be annotated, and a second argument representing the annotation to be published.
+*** @(RSAN1a1)@ In languages where method overloading is supported, the first argument should also accept a @Message@ object (which must contain a populated @serial@ field)
+*** @(RSAN1a2)@ The form of the second argument may accept any language-idiomatic representation (e.g. plain objects in untyped languages), but must allow the user to supply at least the @type@, @clientId@, @name@, @count@, @data@, and @extras@ fields
+*** @(RSAN1a3)@ The SDK must validate that the user supplied a @TAN5b@ @type@. (All other fields are optional)
+** @(RSAN1c)@ The SDK must POST a request body consisting of an array with one element, which is the annotation to be published
+*** @(RSAN1c1)@ The @Annotation.action@ must be set to @ANNOTATION_CREATE@
+*** @(RSAN1c2)@ If the user has supplied an @Annotation.data@, that must be encoded (and the @encoding@ set) just as it would be for a @Message@, per @RSL4@
```
lawrence-forooghian, line 432: Should we support library-generated idempotency keys for REST annotation publishing the same way we do for messages (RSL1k1)?

textile/features.textile
```diff
@@ -893,6 +915,27 @@ then the @enter@ request results in an error immediately.
...
 ** @(RTP15f)@ If the client is identified and has a valid @clientId@, and the @clientId@ argument does not match the client's @clientId@, then it should indicate an error. The connection and channel remain available for further operations
 
+
+h3(#realtime-annotations). RealtimeAnnotations
+
+* @(RTAN1)@ @RealtimeAnnotations#publish@ function:
+** @(RTAN1a)@ Must accept the same arguments and performs the same validation, @Annotation.action@ setting, and data encoding as in @RSAN1@
+** @(RTAN1b)@ Has the same connection and channel state conditions as message publishing, see @RTL6c@
+** @(RTAN1c)@ Must put the annotation to be published into an array in the @annotations@ field of a @ProtocolMessage@ whose @action@ field is set to @ANNOTATION@, and @channel@ field is set to the channel name, and send it to the connection
+** @(RTAN1d)@ Must indicate success or failure of the publish (once @ACKed@ or @NACKed@) in the same way as @RealtimeChannel#publish@
```
lawrence-forooghian, line 925: RTN7a needs updating to mention `ANNOTATION` protocol messages.

textile/features.textile
```diff
@@ -893,6 +915,27 @@ then the @enter@ request results in an error immediately.
...
+* @(RTAN2)@ @RealtimeAnnotations#delete@ function:
+** @(RTAN2a)@ Must be identical to @RTAN1@ @publish()@ except that the @Annotation.action@ is set to @ANNOTATION_DELETE@, not @ANNOTATION_CREATE@
+* @(RTAN3)@ @RealtimeAnnotations#get@ function:
+** @(RTAN3a)@ Is identical to @RestAnnotations#get@
+* @(RTAN4)@ @RealtimeAnnotations#subscribe@ function:
```
lawrence-forooghian, line 930: Can we be explicit about what triggers a call to a listener here (the receipt of an `ANNOTATION` protocol message, presumably)?

textile/features.textile
```diff
@@ -893,6 +915,27 @@ then the @enter@ request results in an error immediately.
...
+** @(RTAN4b)@ Analagously to @name@-filtering with @RealtimeChannel#subscribe@, if the user subscribes with a @type@ (or array of @type@s) the SDK must deliver only annotations whose @type@ field exactly equals the requested @type@
+** @(RTAN4c)@ Delivers decoded @Annotation@ objects to the supplied listener
+** @(RTAN4d)@ Has the same connection and channel state preconditions and return value as @RealtimeChannel#subscribe@, including implicitly attaching unless the user requests otherwise per @RTL7g@/@RTL7h@
+** @(RTAN4e)@ Has one additional channel state condition: once the channel has been attached if necessary, the @RTL4m@ channel modes are checked for the presence of the @ANNOTATION_SUBSCRIBE@ mode. (Note: the thing being checked is the mode set actually granted, according to the server, not the set of modes requested by the user per @TB2d@). If this mode is missing, the library should reject the subscription with an @ErrorInfo@ with code @93001@ and a message to the effect that the user is trying to add an annotation listener without having requested the @ANNOTATION_SUBSCRIBE@ channel mode in ChannelOptions
```
lawrence-forooghian, line 935: A few thoughts about this one:

1. Does this make adding the listener become an asynchronous thing? The current convention (at least, as far as I understand it) for the `subscribe` methods is that the listener is synchronously added, and that if you call `unsubscribe(listener)` before `subscribe` completes the implicit attach, the listener will get removed.
2. Presumably should be "if the channel is already `ATTACHED` or once the channel has been attached if necessary".
3. This check can only be a best effort, given that there might be no attach if `attachOnSubscribe` is disabled, or the implicit attach might fail.

lawrence-forooghian (commented):
Reviewed everything except for the docstrings; will look at those shortly.

textile/features.textile
```diff
@@ -2261,6 +2354,7 @@ class Message: // TM*
   refType: string? // TM2m
   operation: Object? // TM2n
   createdAt: Time? // TM2o
+  summary: Dict<string, JsonObject>? // TM2q
```
lawrence-forooghian, line 2357: I worry that we're pushing quite a lot of work onto users by using such a loose type here. It's not too bad in an untyped language or something like TypeScript where you can tell the compiler "assume that this JSONObject has this given (potentially nested) shape". But there isn't a similar mechanism in a language like Swift in which the type system cannot represent arbitrarily-shaped JSON objects; we'd need to give the user a dictionary with untyped values and they'd need to convert it into some strongly typed object they can work with (i.e. something that looks like one of the `SummaryEntry` values that you've given in the docstrings). I think that expecting to to do this for non-user-supplied data (i.e. data whose shape Ably controls) would be quite a faff for them.

The ideal would be for the SDK to provide these possible `SummaryEntry` values as part of its public API, so that one of the following can happen:

1. the SDK, using a rule for mapping from a `Message.summary` key to a `SummaryEntry` value type, creates the appropriate `SummaryEntry` value type, or
2. same thing as 1. but the mapping is performed by the user

(None of this precludes us from allowing one of these possible `SummaryEntry` values to be an arbitrary JSON object for handling unknown aggregation methods.)

textile/api-docstrings.md
```diff
@@ -616,9 +616,106 @@ Enumerates the possible types of `Message` there can be.
...
+| count | number? | TAN2e | An optional count, only relevant to certain aggregation types, see aggregation types documentation for more info. |
+| data | any? | TAN2f | An arbitrary publisher-provided payload, if provided. Not aggregated. |
+| encoding | string? | TAN2g | The encoding of the payload; typically empty as the data is decoded client-side back into the original data type. |
+| timestamp | number | TAN2h | Timestamp of when the annotation was received by Ably, as milliseconds since the Unix epoch. |
+| serial | string | TAN2i | This message's unique serial (lexicographically totally ordered). |
+| messageSerial | string | TAN2j | The serial of the message (of type `MESSAGE_CREATE`) that this annotation is annotating. |
+| type | string | TAN2k | The type of annotation it is, typically some name together with an aggregation method; for example: "emoji:distinct.v1". Handled opaquely by the SDK and validated serverside. |
```
lawrence-forooghian, line 702: When I read this (not yet knowing anything about annotations), my first thought was that the "name" in "typically some name" was related to the `name` property of the annotation. Is there some other term we can use here to avoid this confusion?

textile/features.textile
```diff
@@ -2227,9 +2316,13 @@ enum MessageAction: // TM5
   MESSAGE_CREATE // TM5
   MESSAGE_UPDATE // TM5
   MESSAGE_DELETE // TM5
-  META_OCCUPANCY // TM5
+  META // TM5
   MESSAGE_SUMMARY // TM5
 
+enum AnnotationAction: // TAN2b
```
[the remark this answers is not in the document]
lawrence-forooghian: I see that we have RTF1, which says

> In particular, deserialization of ProtocolMessages and related types, and associated enums, must be tolerant to unrecognised attributes or enum values. Such unrecognised values must be ignored.

but I don't really know what it means to "ignore" an unrecognised `AnnotationAction`, for example.

textile/features.textile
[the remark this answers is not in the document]
SimonWoolf: Good point -- not really, other than that we expect most uses of annotations to not even specify any data, so to not really be in any danger of exceeding the size limit.

But it raises the question of, it looks like presence are exempt too -- probably also overlooked. If we're going to add this we should probably do it for all message types.

I didn't implement a clientside size check for ably-js, and honestly I'm tempted to suggest leaving this for now for both annotations and presence. Or adding it as a MAY.

textile/api-docstrings.md
```diff
@@ -616,9 +616,106 @@ Enumerates the possible types of `Message` there can be.
 | MESSAGE_CREATE | TM5 | A normal message that has been published by a user. |
 | MESSAGE_UPDATE | TM5 | An update that modifies a previously-published message (referenced by `serial`). |
 | MESSAGE_DELETE | TM5 | A deletion of a previously-published message (referenced by `serial`). |
-| META_OCCUPANCY | TM5 | Channel metadata containing information on what the current occupancy of the channel is. |
+| META | TM5 | A meta-message (a message originating from ably rather than being explicitly published on a channel), containing information such as inband channel occupancy events that has been requested by channel param. |
```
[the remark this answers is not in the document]
SimonWoolf: cf https://github.com/ably/ably-js/pull/1997#discussion_r2037486695 -
> You're not wrong that this is technically a breaking change.

> But the reason I wouldn't worry about it in this case is that `Message.action` was only introduced recently as part of the annotations/materialization work, which isn't yet part of a publicly released feature yet, and isn't added to our [main documentation](https://ably.com/docs/api/realtime-sdk/types#message). So it's just very unlikely that anyone is relying on the action yet. As you say, our inband occupancy events docs still all say to use the message name, which is not changing.

> If that wasn't the case then we wouldn't have made this change.

textile/features.textile
SimonWoolf: Sure, good idea, added

textile/features.textile
SimonWoolf: sure, clarified

textile/features.textile
SimonWoolf: I did consider 1 but, like you point out, it kindof sucks because it means that when we do add support the type changes suddenly in what's supposed to be a non-breaking library change. And we really do want to be free to add new aggregations serverside without having to coordinate half a dozen SDK releases each time.

But 2 seems much more reasonable, that's not a bad idea..

textile/features.textile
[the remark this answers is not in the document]
SimonWoolf: They're not frozen, we may add new actions.

It's true that "ignore" only really makes sense for unrecognised attributes, not unknown enum values.
But really the answer for enums is, it doesn't really matter (can do something language-idiomatic -- in js maybe null or an 'unknown' string, for languages which use enums maybe an UNKNOWN enum member that's always at one position later than the highest known, whatever). The only thing the spec requires about is that the library should cope in some graceful way (not start throwing exceptions or panicing etc)

textile/api-docstrings.md
```diff
@@ -671,6 +768,36 @@ Contains an individual message that is sent to, or received from, Ably.
 | createdAt: Time? ||| TM2o | The timestamp of the very first version of a given message (will differ from `timestamp` only if the message has been updated or deleted). |
 | operation: Operation? ||| TM2n | In the case of an updated or deleted message, this will contain metadata about the update or delete operation. |
 | connectionKey: String? ||| TM2h | Allows a REST client to publish a message on behalf of a Realtime client. If you set this to the [private connection key]{@link Connection.key} of a Realtime connection when publishing a message using a [`RestClient`]{@link RestClient}, the message will be published on behalf of that Realtime client. This property is only populated by a client performing a publish, and will never be populated on an inbound message. |
+| summary: `Dict<string, SummaryEntry>`? ||| TM2q | A summary of all the annotations that have been made to the message. Will always be populated for a message with action `MESSAGE_SUMMARY`, and may be populated for any other type (in particular a message retrieved from REST history will have its latest summary included). |
```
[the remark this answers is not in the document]
SimonWoolf: I can put that there, but have to be a bit careful to keep it reasonably brief, the docstring isn't a replacement for the main documentation

textile/api-docstrings.md
```diff
@@ -671,6 +768,36 @@ Contains an individual message that is sent to, or received from, Ably.
...
+
+## Type SummaryEntry
+
+A list of some different possible values of the Message.summary map. This list is non-exhaustive as new aggregation types may be added at any time.
```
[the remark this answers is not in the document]
SimonWoolf: This is kindof a limitation of typescript. I initially had `type SummaryEntry = SummaryDistinctValues | ... | unknown`. The trouble is that `unknown` is a collapsing type; `x | y | unknown` is just equivalent to `unknown` -- which means you don't get any type hints, autocompletion etc. So I ended up removing the `| unknown`, and expecting we'd have to tell people in the docs to use `(msg.summary as any).xxx` for summary types the ably-js types don't include yet.

If you happen to think of any better suggestions for that than what I did that I'm all ears, I realise it's not great.

textile/api-docstrings.md
```diff
@@ -616,9 +616,106 @@ Enumerates the possible types of `Message` there can be.
...
+
+## Class Annotation
+
+| Property | Type | Spec | Description |
+|---|---|---|---|
+| id | string | TAN2a | An ID assigned (by the publisher or Ably) to this annotation, which is used as the idempotency key. |
+| action | AnnotationAction | TAN2b | The action, whether this is an annotation being added or removed, one of the AnnotationAction enum values. |
+| clientId | string? | TAN2c | The client ID of the publisher of this annotation (if any). |
+| name | string? | TAN2d | The name of this annotation. This is the field that most annotation aggregations will operate on. For example, using "distinct.v1" aggregation (specified in the type), the message summary will show a list of clients who have published an annotation with each distinct annotation.name. |
+| count | number? | TAN2e | An optional count, only relevant to certain aggregation types, see aggregation types documentation for more info. |
```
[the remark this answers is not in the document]
SimonWoolf: yes, docs have to follow later after sdk support -- you can't release a feature before sdks support it, which means it needs to be specified first, which is this pr. those docs do not yet exist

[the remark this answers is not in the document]
SimonWoolf: sure

textile/features.textile
SimonWoolf: I initially did want it to only add the listener after the validation, but andy persuaded me to change it, see discussion https://github.com/ably/ably-js/pull/1953#discussion_r2016352014 . I don't like that we still add the listener even if the promise is rejected, but reluctantly I figure the least bad option is to be consistent with the existing behaviour, at least until we change both at once.

That convention doesn't appear to be specified for realtimechannel#subscribe so I don't think the additional condition changes anything?

good point about 3. that clause is already getting a bit complex so I think i'll add that caveat as a sub-item

textile/features.textile
```diff
@@ -762,6 +782,7 @@ h3(#realtime-channel). RealtimeChannel
 ** @(RTL18b)@ Discard the message
 ** @(RTL18c)@ Send an @ATTACH@ @ProtocolMessage@ with the @channelSerial@ set to the previous message to the message for which "vcdiff" decoding failed to the server, transitioning the channel state to @ATTACHING@, and waiting for a confirmation @ATTACHED@, as per @RTL4c@ and @RTL4f@. @ChannelStateChange.reason@ should be set to @ErrorInfo@ object with with code 40018.
 * @(RTL19)@ The data payload of the last message on each channel must be stored at all times since it will be needed to decode any subsequent message that has a "vcdiff" encoding step. The stored value is the "base payload" of the most recent message; this is the @data@ member of the message, in string or binary form, once all application-level encoding steps have been applied. The base payload is derived initially by processing a non-delta message; the processing and bookkeeping rules are as follows:
+** @(RTL19d)@ Base payloads are only stored for messages (that is, with a @ProtocolMessage.action@ of @MESSAGE@), which have a @Message.action@ which is not @META@
```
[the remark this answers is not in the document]
lawrence-forooghian: Seems like something that it would be good for us to be able to get to work on fixing without depending on this PR being merged. Would you mind pulling it out into a separate PR? Also please could you add something to the commit message explaining that it's a bug fix?

textile/api-docstrings.md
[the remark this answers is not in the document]
lawrence-forooghian: > which means you don't get any type hints, autocompletion etc

I'm not sure I understand to what extent you _currently_ get these (i.e. even without the `unknown`)? Presumably, given a `SummaryEntry`, the only thing that autocompletion can offer you are properties which exist on the union of all possible `SummaryEntry` types, which doesn't seem hugely useful?

Perhaps the thing to do here in TypeScript is that which resembles option 2 of https://github.com/ably/specification/pull/292#discussion_r2045398430 — i.e. keep `SummaryDistinctValues` etc but remove `SummaryEntry` and change `Message.summary` to a JSON object?

textile/features.textile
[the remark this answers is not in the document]
lawrence-forooghian: I'm happy for it to be left til later — mind creating an issue for it?

textile/features.textile
lawrence-forooghian: If you're happy with 2, then I think that the thing to do is to add the TypeScript types (`SummaryDistinctValues`, `SummaryUniqueValues`, `SummaryMultipleValues`, `SummaryClientIdList`, `SummaryTotal`) into the spec, and have a spec point explaining that there should be a mechanism for a user to convert a `Message.summary` into a user-selected one of these (this mechanism might just be a type assertion e.g. in TypeScript, or a constructor that takes a `JSONObject` e.g. Swift). What do you think?

textile/features.textile
lawrence-forooghian: > That convention doesn't appear to be specified for realtimechannel#subscribe so I don't think the additional condition changes anything?

That is true, but I guess it doesn't hurt to make it explicit here?

More broadly, I'm still on the fence about whether we should have this error, given that it's primarily of value in the mental model in which there exists some relation between attaching a channel and adding an event listener for a channel event. We seem to be questioning this mental model internally (moving away from it in Chat, adding an opt-out for it in the core SDKs) and have to go out of our way in documentation to explain that subscribing and attaching are different things. And I worry that by adding this error, we're actually doubling down on this mental model, which we're elsewhere trying to move away from; this could be confusing to users.

textile/api-docstrings.md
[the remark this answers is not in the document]
SimonWoolf: Yeah, fair point.

You do get one thing with the current typings, which is you can use go-to-type to go to the relevant bit of the type definitions and look at them. Adding `| unknown` kills that: the type is collapsed into `unknown` and drops all metadata. But now I come to think more, a better solution would be `| {[k: string]: unknown}`. (Which is essentially JSONObject).

Not sure I see the argument for removing SummaryEntry as a union type, it gives you that thing and doesn't really harm anything. Either way you'd still get that option 2 through a type assertions.

textile/features.textile
SimonWoolf: >  I'm still on the fence about whether we should have this error, given that it's primarily of value in the mental model in which there exists some relation between attaching a channel and adding an event listener for a channel event.

I absolutely agree with you about implicit attach, I was arguing against it years ago.

...But I don't agree that this error is "primarily" about implicit attach. The check is equally valid if you subscribe after being attached for hours. I could even have chosen to make the check on the channel modes being requested in channel options, which is purely local and has nothing to do with your attachment state at all; I only didn't because (a) it's possible to request channel modes in a way that's opaque to the sdk and (b) this way you also catch a user who doesn't have the mode because they don't have the capabilities.

textile/api-docstrings.md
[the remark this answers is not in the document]
lawrence-forooghian: Ah, so if you go with the `{[k: string]: unknown}` then it doesn't collapse it, I see. That sounds good. Might be worth adding a comment alongside the TypeScript type definition to explain why the union type exists; I can quite easily see someone otherwise coming along and going "this type adds no value, I'll remove it".

textile/features.textile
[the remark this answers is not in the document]
lawrence-forooghian: I'm not sure that any of our SDKs have actually considered how to handle unknown enum values.

For example, to take the example of TypeScript and `Annotation.action`, in the type definitions we say that this field is always populated and that it's either `"annotation.create"` or `"annotation.delete"`, but in fact we populate it with `undefined` given an unrecognised numerical value (the compiler doesn't catch this because we don't declare our internal `Annotation` type as implementing the public `API.Annotation` interface). (Ditto for, say, `PresenceMessage.action`.)

And from a brief look at ably-cocoa it seems to be a similar situation there — we end up populating a C enum with a value that isn't one of its declared cases, which is going to lead to all sorts of weird behaviour (compiler not exhaustively checking `switch` statements and the implications thereof).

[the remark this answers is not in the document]
SimonWoolf: You make a good point, we are not consistent between sdks (or even with sdks) about this, and that's probably worth harmonising at some point. But I don't think that need be in this PR.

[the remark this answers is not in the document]
lawrence-forooghian: > But I don't think that need be in this PR.

Agreed, have created https://github.com/ably/specification/issues/304

textile/features.textile
lawrence-forooghian: My preference would be that we allow this to be a warn log message, at the discretion of the platform. I was trying to figure out how we'd implement this in Swift, which distinguishes between two types of errors:

- programmer error: the caller is using an API correctly; the expected behaviour here is to trap (i.e. crash the program)
- runtime error: not a logic error; something has gone "expectedly wrong" at runtime; the language _forces_ users to catch and handle these errors

This error doesn't really fall neatly into either of these categories:

- we can't treat it as a programmer error, because the programmer who called `subscribe()` might be relying on the correct channel mode being given via a server-issued token, and we shouldn't be crashing an app because the server misbehaved
- if we treat it as a runtime error, then we have a `subscribe()` which now throws errors for two different reasons (1. implicit attach failed, 2. channel modes are wrong), and the programmer, as mentioned above, has to catch and handle this thrown error, thus forcing them to be able to distinguish between the different kinds of errors, and to have error handling code for an error that is unlikely to ever actually happen in a properly-configured program

SimonWoolf: Sure, I'll make it discretionary.

(JS of course used to have that distinction as well -- exception for programmer errors and callback with err for runtime errors -- now that it's an async function those are both collapsed into promise rejection. Which in some ways makes things easier, but in other ways makes the subscribe function now even less intuitive: when there were two separate types of error we could document that if an exception was not thrown, the subscription has been added, even if the callback was called with an error; now that it's an async function we can no longer do that, which makes the api inconsistent)

textile/features.textile
SimonWoolf: Done, wdyt?
Slight tweaks to naming (added the version to the constructor name to match the version being in the annotation type) that I'll need to do in ably-js too.

textile/features.textile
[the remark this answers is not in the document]
SimonWoolf: Pulled it out.

SimonWoolf:
@lawrence-forooghian any further comments?

lawrence-forooghian:
Was away for a couple of weeks — will take a look now.

textile/features.textile
lawrence-forooghian: Looks good, please could you update the docstrings and IDL to match?

textile/api-docstrings.md
```diff
@@ -620,9 +620,106 @@ Enumerates the possible types of `Message` there can be.
...
+## class RestAnnotations
+
+
+Functionality for annotating messages with small pieces of data, such as emoji reactions, that the server will roll up into the message as a summary.
+
+| Method / Property | Parameter | Returns | Spec | Description |
+|---|---|---|---|---|
+| publish(message: Message, annotation: OutboundAnnotation) => io ||| RSAN1 | Publish a new annotation for a message. |
+|| `message` || RSAN1a1 | The message to annotate. |
+|| `annotation` || RSAN1a2 | The annotation to publish. (Must include at least the `type`. Assumed to be an annotation.create if no action is specified) |
```
sacOO7, line 635:
```suggestion
|| `annotation` || RSAN1a2 | The annotation to publish. Must include at least the `type` |
```
Why do we need assumption if its mandatory?

sacOO7: Or are we assuming `action` based on `type` value?

sacOO7, line 635:
```suggestion
|| `annotation` || RSAN1a2 | The annotation to publish. (Must include at least the `type`. Assumed to be an annotation.create if no annotation `action` is specified) |
```
In case of `delete` method, there's no statement for assumption. It's supposed to be annotation.delete right

textile/api-docstrings.md
```diff
@@ -620,9 +620,106 @@ Enumerates the possible types of `Message` there can be.
...
+| publish(messageSerial: string, annotation: OutboundAnnotation) => io ||| RSAN1 | Publish a new annotation for a message (alternative form where you only have the serial of the message, not a complete Message object) |
+|| `messageSerial` || RSAN1a | The serial field of the message to annotate. |
+|| `annotation` || RSAN1a2 | The annotation to publish. (Must include at least the `type`. Assumed to be an annotation.create if no action is specified) |
+| delete(message: Message, annotation: OutboundAnnotation) => io ||| RSAN2 | Delete an annotation for a message. |
+|| `message` || RSAN2a | The message to remove the annotation from. |
+|| `annotation` || RSAN2a | The annotation to delete. (Must include at least the `type`.) |
```
sacOO7, line 641:
```suggestion
|| `annotation` || RSAN2a | The annotation to delete. (Must include at least the `type`.) | Assumed to be an annotation.delete if no `action` is specified) |
```

textile/api-docstrings.md
```diff
@@ -620,9 +620,106 @@ Enumerates the possible types of `Message` there can be.
...
+| delete(messageSerial: string, annotation: OutboundAnnotation) => io ||| RSAN2 | Delete an annotation for a message (alternative form where you only have the serial of the message, not a complete Message object) |
+|| `messageSerial` || RSAN2a | The serial field of the message to remove the annotation from. |
+|| `annotation` || RSAN2a | The annotation to delete. (Must include at least the `type`.) |
```
sacOO7, line 644:
```suggestion
|| `annotation` || RSAN2a | The annotation to delete. (Must include at least the `type`.) | Assumed to be an annotation.delete if no `action` is specified) |
```

sacOO7: Do we ignore annotation action in case of `delete` method?

textile/api-docstrings.md
```diff
@@ -620,9 +620,106 @@ Enumerates the possible types of `Message` there can be.
...
+| action | AnnotationAction | TAN2b | The action, whether this is an annotation being added or removed, one of the AnnotationAction enum values. |
+| clientId | string? | TAN2c | The client ID of the publisher of this annotation (if any). |
+| name | string? | TAN2d | The name of this annotation. This is the field that most annotation aggregations will operate on. For example, using "distinct.v1" aggregation (specified in the type), the message summary will show a list of clients who have published an annotation with each distinct annotation.name. |
+| count | number? | TAN2e | An optional count, only relevant to certain aggregation methods, see aggregation methods documentation for more info. |
+| data | any? | TAN2f | An arbitrary publisher-provided payload, if provided. Not aggregated. |
+| encoding | string? | TAN2g | The encoding of the payload; typically empty as the data is decoded client-side back into the original data type. |
+| timestamp | number | TAN2h | Timestamp of when the annotation was received by Ably, as milliseconds since the Unix epoch. |
+| serial | string | TAN2i | This message's unique serial (lexicographically totally ordered). |
```
sacOO7, line 704: This can be called `messageSerial`, just `serial` isn't self-explainatory

textile/api-docstrings.md
```diff
@@ -620,9 +620,106 @@ Enumerates the possible types of `Message` there can be.
 | MESSAGE_CREATE | TM5 | A normal message that has been published by a user. |
 | MESSAGE_UPDATE | TM5 | An update that modifies a previously-published message (referenced by `serial`). |
 | MESSAGE_DELETE | TM5 | A deletion of a previously-published message (referenced by `serial`). |
-| META_OCCUPANCY | TM5 | Channel metadata containing information on what the current occupancy of the channel is. |
+| META | TM5 | A meta-message (a message originating from ably rather than being explicitly published on a channel), containing information such as inband channel occupancy events that has been requested by channel param. |
```
sacOO7, line 623: Since this is server synthesized can't we call it 
`ABLY_SYNTHESIZED_META`

textile/features.textile
```diff
@@ -2493,6 +2612,22 @@ class ObjectData // OD*, internal
...
+  id: String // TAN2a
+  action: AnnotationAction // TAN2b
+  clientId: String? // TAN2c
+  name: String? // TAN2d
+  count: Int? // TAN2e
+  data: Data? // TAN2f
+  encoding: String? // TAN2g
+  timestamp: Time // TAN2h
+  serial: String // TAN2i
+  messageSerial: String // TAN2j
```
sacOO7, line 2627:
```suggestion
  messageSerial: String // TAN2i
  seedMessageSerial: String // TAN2j
```

textile/api-docstrings.md
```diff
@@ -620,9 +620,106 @@ Enumerates the possible types of `Message` there can be.
...
+| clientId | string? | TAN2c | The client ID of the publisher of this annotation (if any). |
+| name | string? | TAN2d | The name of this annotation. This is the field that most annotation aggregations will operate on. For example, using "distinct.v1" aggregation (specified in the type), the message summary will show a list of clients who have published an annotation with each distinct annotation.name. |
+| count | number? | TAN2e | An optional count, only relevant to certain aggregation methods, see aggregation methods documentation for more info. |
+| data | any? | TAN2f | An arbitrary publisher-provided payload, if provided. Not aggregated. |
+| encoding | string? | TAN2g | The encoding of the payload; typically empty as the data is decoded client-side back into the original data type. |
+| timestamp | number | TAN2h | Timestamp of when the annotation was received by Ably, as milliseconds since the Unix epoch. |
+| serial | string | TAN2i | This message's unique serial (lexicographically totally ordered). |
+| messageSerial | string | TAN2j | The serial of the message (of type `MESSAGE_CREATE`) that this annotation is annotating. |
```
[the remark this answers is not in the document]
sacOO7: Okay, then we will update in SDK models 👍

textile/features.textile
sacOO7: will be updated in SDK models, closing as per https://github.com/ably/specification/pull/292#discussion_r2135598426

textile/api-docstrings.md
SimonWoolf: No: the `Annotation.messageSerial` is a separate field, one line further down, which is the serial of the Message that this annotation is annotating; the `Annotation.serial` is the annotation's _own_ serial.

(It should however say "This annotation's..." rather than "This message's...", I will fix)

textile/api-docstrings.md
sacOO7: Or `SERVER_COMPOSED_META`/ `ABLY_COMPOSED_META`

textile/features.textile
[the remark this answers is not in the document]
SimonWoolf: No.

textile/features.textile
```diff
@@ -910,6 +934,27 @@ h3(#objects). Objects
 
 Reserved for @Objects@ feature specification. Reserved spec points: @OBJ@, @LM@, @LC@
 
+h3(#realtime-annotations). RealtimeAnnotations
+
+* @(RTAN1)@ @RealtimeAnnotations#publish@ function:
+** @(RTAN1a)@ Must accept the same arguments and performs the same validation, field setting, and data encoding as in @RSAN1@
+** @(RTAN1b)@ Has the same connection and channel state conditions as message publishing, see @RTL6c@
+** @(RTAN1c)@ Must put the annotation to be published into an array in the @annotations@ field of a @ProtocolMessage@ whose @action@ field is set to @ANNOTATION@, and @channel@ field is set to the channel name, and send it to the connection
```
sacOO7, line 942: In this case we didin't specify as `create` => https://github.com/ably/specification/pull/292/files#r2137030102

sacOO7 (commented):
Added few typo comments. 
Other than this, do we need to mention that this introduces `mutable` messages.
I could see this change as a part of java annotations PR
[AppSpec.json](https://github.com/ably/ably-java/blob/93d9b314ce411f735ee64e099aa160ffbb6cc0b8/lib/src/test/resources/local/testAppSpec.json#L37)

lawrence-forooghian (approved):

textile/features.textile
```diff
@@ -425,6 +426,28 @@ h3(#rest-encryption). Encryption
...
+** @(RSAN1c)@ The SDK must POST a request body consisting of an array with one element, which is the annotation to be published
+*** @(RSAN1c1)@ The @Annotation.action@ must be set to @ANNOTATION_CREATE@
+*** @(RSAN1c2)@ The @Annotation.messageSerial@ must be set to the identifier specified in the first argument to @publish()@
+*** @(RSAN1c3)@ If the user has supplied an @Annotation.data@, that must be encoded (and the @encoding@ set) just as it would be for a @Message@, per @RSL4@
+*** @(RSAN1c4)@ If @idempotentRestPublishing@ (see "TO3n":#TO3n) is enabled and the annotation has an empty @id@ attribute, the SDK should generate a string by base64-encoding a sequence of at least 9 bytes obtained from a source of randomness, append @:0@, and set it as the @Annotation.id@
+*** @(RSAN1c5)@ The body must be encoded to the appropriate format per @RSC8@
+*** @(RSAN1c6)@ The body must be sent as a POST request to the endpoint @/channels/{channelName}/messages/{messageSerial}/annotations@, where @messageSerial@ is the identifier specified in the first argument to @publish()@
```
sacOO7, line 443: What happens when `POST` request fails, is there custom errorCode/ statusCode for annotations for error returned from server

textile/api-docstrings.md
sacOO7: Sure, will resolve this 👍

textile/api-docstrings.md
sacOO7: Change not needed

textile/api-docstrings.md
```diff
@@ -676,6 +788,34 @@ Contains an individual message that is sent to, or received from, Ably.
 | createdAt: Time? ||| TM2o | The timestamp of the very first version of a given message (will differ from `timestamp` only if the message has been updated or deleted). |
 | operation: Operation? ||| TM2n | In the case of an updated or deleted message, this will contain metadata about the update or delete operation. |
 | connectionKey: String? ||| TM2h | Allows a REST client to publish a message on behalf of a Realtime client. If you set this to the [private connection key]{@link Connection.key} of a Realtime connection when publishing a message using a [`RestClient`]{@link RestClient}, the message will be published on behalf of that Realtime client. This property is only populated by a client performing a publish, and will never be populated on an inbound message. |
+| summary: `Dict<string, JsonObject>`? ||| TM2q | A summary of all the annotations that have been made to the message, whose keys are the `type` fields from any annotations that it includes. For summary types the SDK knows about, static factor methods on `Message` exist to parse the values into strongly-typed objects. |
```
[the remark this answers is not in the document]
ttypic: Can we define this more loosely in the spec? Not require static factory methods on `Message`, but suggest providing some utility methods, idiomatic to the SDK, to perform these casts to strongly-typed objects. I’m just not sure that `Message` is the right place to put those static methods.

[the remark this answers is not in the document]
sacOO7: Yeah, in case of ably-java, we added them as static methods of [Summary.java](https://github.com/ably/ably-java/blob/49978fa62858ff8ecc5fe50c03f900114ac39ca2/lib/src/main/java/io/ably/lib/types/Summary.java) or can they be public utility methods wdyt

sacOO7 (changes requested):
requested few changes

textile/features.textile
SimonWoolf: > I guess this has to follow pattern where Summary is a base class and SummaryUniqueValues, SummaryMultipleValues, ... extends Summary

The pattern Lawrence proposed, and I then adopted, for languages where it is not easy to work with plain objects, is a function `SummaryDistinctV1()` that takes a `JsonObject` (the value of some key in the summary dict), and returns a structure that is specific to a distinct v1 summary. The user is expected to know what aggregation type they're using and call the correct function. In this pattern, there is no subclassing.

I think you're imagining an api where the library detects what kind of summary entry it is and does the decoding itself automatically, which would then need subclassing. That is a possible api we considered, but we decided against it.

In ably-js working with plain objects is easy and idiomatic, so the sdk just returns objects, and the user can use type assertions to get the correct type. (We could potentially also provide type assertion functions).

textile/features.textile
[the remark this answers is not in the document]
SimonWoolf: This is covered by RTAN1a, which pulls in RSAN1 and therefore RSAN1c1.

textile/features.textile
```diff
@@ -425,6 +426,28 @@ h3(#rest-encryption). Encryption
...
+*** @(RSAN1a1)@ In languages where method overloading is supported, the first argument should also accept a @Message@ object (which must contain a populated @serial@ field)
+*** @(RSAN1a2)@ The form of the second argument may accept any language-idiomatic representation (e.g. plain objects in untyped languages), but must allow the user to supply at least the @type@, @clientId@, @name@, @count@, @data@, and @extras@ fields
+*** @(RSAN1a3)@ The SDK must validate that the user supplied a @TAN5b@ @type@. (All other fields are optional)
+*** @(RSAN1a4)@ The SDK may validate that the total size of the annotation or (if publishing an array) annotations, calculated identially to message size per "TO3l8":#TO3l8, exceeds the @maxMessageSize@, then the client library should reject the publish and indicate an error with code 40009
+** @(RSAN1c)@ The SDK must POST a request body consisting of an array with one element, which is the annotation to be published
+*** @(RSAN1c1)@ The @Annotation.action@ must be set to @ANNOTATION_CREATE@
```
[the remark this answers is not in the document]
SimonWoolf: `create()` as a method name sounds like a constructor, it doesn't sound like a method that publishes something to a channel.

PR #296, opened 2025-04-16
Specify the behaviour of `#connect` when `CONNECTING`

lawrence-forooghian:
The connection states table mentioned this but it wasn't actually specified. From a brief check of a few implementations, this seems to match implementations.

Made a couple of fixes to connection states table whilst I was at it; see commit messages for details.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -526,7 +526,9 @@ h3(#realtime-connection). Connection
 ** @(RTN9c)@ Is @Null@ when the SDK is in the @CLOSED@, @CLOSING@, @FAILED@, or @SUSPENDED@ states
 * @(RTN10)@ This clause has been deleted. It was valid up to and including specification version @1.2@.
 * @(RTN11)@ @Connection#connect@ function:
-** @(RTN11a)@ Explicitly connects to the Ably service if not already connected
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -525,7 +525,8 @@ h3(#realtime-connection). Connection
 ** @(RTN9c)@ Is @Null@ when the SDK is in the @CLOSED@, @CLOSING@, @FAILED@, or @SUSPENDED@ states
 * @(RTN10)@ This clause has been deleted. It was valid up to and including specification version @1.2@.
 * @(RTN11)@ @Connection#connect@ function:
-** @(RTN11a)@ Explicitly connects to the Ably service if not already connected
+** @(RTN11a)@ This clause has been replaced by @"RTN11e@":#RTN11e. It was valid up to and including specification version @TBD@.
+** @(RTN11e)@ Explicitly connects to the Ably service if not @CONNECTING@ or @CONNECTED@
```
sacOO7, line 529:
```suggestion
** @(RTN11e)@ No-op if connection state is @CONNECTING@ or @CONNECTED@
```

lawrence-forooghian: But then we no longer have a spec point that says that calling `#connect` triggers a connection?

sacOO7: Then I guess we should add one, and let other spec points reference it.

lawrence-forooghian: Sure — have updated

sacOO7 (approved):
Lgtm

PR #213, opened 2024-10-24
Updates for ADR-119: updated ClientOptions for new domains

paddybyers:
This introduces the new endpoint ClientOption and deprecates other options as described in the ADR. This restructures the specification of primary and fallback domains to make this clearer and more consistent.

PR #301, opened 2025-04-17
Readability tweaks for new domain logic

lawrence-forooghian:
A few changes to the language in #213, for readability.


lawrence-forooghian:
Oops, forgot that (now-merged) #302 was stacked on top of this PR, so its contents are now here.

textile/features.textile
```diff
@@ -75,12 +75,12 @@ h2(#endpoint-configuration). Client library endpoint configuration
...
-** @(REC1c)@ If the deprecated @environment@ option is specified then it defines a production routing policy name @[name]@:
+*** @(REC1b2)@ If the @endpoint@ option is a hostname, determined by it containing at least one period (@.@), or containing at least one instance of the string @::@, or being equal to the string @localhost@, then the @primary domain@ is the value of the @endpoint@ option.
+*** @(REC1b3)@ Otherwise, if the @endpoint@ option is a non-production routing policy name, determined by it having the form @nonprod:[id]@, then the @primary domain@ is @[id].realtime.ably-nonprod.net@.
+*** @(REC1b4)@ Otherwise, the @endpoint@ option is a production routing policy name of the form @[id]@, and the @primary domain@ is @[id].realtime.ably.net@.
+** @(REC1c)@ If the deprecated @environment@ option is specified then it defines a production routing policy name @[id]@:
 *** @(REC1c1)@ If any one of the deprecated options @restHost@, @realtimeHost@ are also specified then the options as a set are invalid.
-*** @(REC1c2)@ Otherwise, the primary domain is @[name].realtime.ably.net@.
+*** @(REC1c2)@ Otherwise, the primary domain is @[id].realtime.ably.net@.
```
lmars, line 83: Why change this from `name` to `id`? The text refers to it as a "routing policy name", so `[name]` seems appropriate to me.

textile/features.textile
```diff
@@ -92,10 +92,10 @@ h2(#endpoint-configuration). Client library endpoint configuration
...
-*** @(REC2c5)@ Otherwise, if the @primary domain@ is determined to be a routing policy name via @(REC1c2)@ then the set of @fallback domains@ is @[name].a.fallback.ably-realtime.com@, @[name].b.fallback.ably-realtime.com@, @[name].c.fallback.ably-realtime.com@, @[name].d.fallback.ably-realtime.com@, @[name].e.fallback.ably-realtime.com@.
+*** @(REC2c2)@ Otherwise, if the @primary domain@ is determined to be a hostname via @(REC1b2)@ then the set of @fallback domains@ is empty.
+*** @(REC2c3)@ Otherwise, if the @primary domain@ is determined by a non-production routing policy name via @(REC1b3)@ then the set of @fallback domains@ is @[id].a.fallback.ably-realtime-nonprod.com@, @[id].b.fallback.ably-realtime-nonprod.com@, @[id].c.fallback.ably-realtime-nonprod.com@, @[id].d.fallback.ably-realtime-nonprod.com@, @[id].e.fallback.ably-realtime-nonprod.com@.
```
lmars, line 96: I think this should still read "determined to be a", not "determined by a"

lawrence-forooghian: I gave an explanation for this one in the commit message ("The existing language implies that the primary domain equals the routing policy name, which is not what the referenced spec points say.") Do you disagree with this?

textile/features.textile
lawrence-forooghian: I chose to use `id` consistently across all of these spec points where something is derived from a routing policy ID (even when the ID and the name are the same). Would you be happy if I changed the language of REC1c to be consistent with REC1b4; that is, to read `** @(REC1c)@ If the deprecated @environment@ option is specified then it defines a production routing policy name of the form @[id]@:`?

lawrence-forooghian: My understanding of the distinction is that in REC1b3, `nonprod:[id]` is the routing policy name, and `id` is the routing policy ID. I got this terminology from [this Slack conversation](https://ably-real-time.slack.com/archives/CURL4U2FP/p1744854105579979) (referenced in 49cdbaa2ea2eabc4dfe2bf6c7e1d97f8c7946ab8).

textile/features.textile
lawrence-forooghian: I guess that the fallback host spec points are written as they are (predicated on the logic for determining the primary domain) because then it allows us to write all of the spec points in a consistent fashion (i.e. even when the related spec point isn't written in terms of `endpoint`, such as in REC2c1 or REC2c6? If you find the usage of "primary domain" in these spec points too confusing, I'm happy to try changing them, though.

textile/features.textile
lmars: Hmm ok, apologies I missed that discussion, but I'm not sure I agree.

Internally, we have a set of things called "routing policies", each of which has a name (e.g. `main`, `sandbox`, `lmars`).

Some of those routing policies (e.g. `main`) exist in our prod environment (i.e. accessible via prod domains `ably.net`, `ably-realtime.com`), whereas others (e.g. `sandbox`, `lmars`) exist in our nonprod environment (i.e. accessible via nonprod domains `ably-nonprod.net`, `ably-realtime-nonprod.com`).

So, when setting `endpoint=main`, `main` is the name of the routing policy, and the fact that it exists in the prod environment is implicit, but to support using routing policies which exist in the nonprod environment, we decided to prefix the routing policy with `nonprod:` when used in the `endpoint` client option to indicate to the SDK that it should use the nonprod domains, so I think `nonprod:lmars` refers to "a routing policy named `lmars` in the nonprod environment".

I'm happy to attempt to agree this with the other's from that discussion 🙂

lawrence-forooghian: Thanks for the clarification. Would you mind coming to an agreement with @paddybyers and @surminus on the right terminology to use here, please? Then I'll update accordingly 🙂

surminus: > Internally, we have a set of things called "routing policies", each of which has a name (e.g. main, sandbox, lmars).

I suggested that we have sets of things called "routing policies", each of which has an "id", because that's how it's referred to in all the infrastructure code, [Terraform](https://github.com/ably/infrastructure/blob/4826a26a08fc5612222ddf19f9e6a14921fe9450/terraform/modules/ably/terraform-ably-routing-policy/variables.tf#L152-L155), and [documentation](https://ably.atlassian.net/wiki/spaces/INFRA/pages/2597126252/DNS#Primary-Records).

lmars: @surminus thanks for clarifying, so if we're switching to "id", I think we should drop the term "routing policy name" completely and always refer to it as "routing policy id" in the spec.

lawrence-forooghian: Thanks both! Will tweak the wording

lawrence-forooghian: Updated

lmars (approved):

Issue #295, opened 2025-04-16
Render a version of the spec that doesn't contain "this has been removed" etc


lawrence-forooghian:
For someone just wanting to read through the spec and understand how the library works, these add quite a lot of noise.

PR #302, opened 2025-04-17
Make `endpoint` special-case `localhost` and IP addresses

lawrence-forooghian:
I believe it's quite common (e.g. for Realtime devs) to want to point to a local Realtime. Or, for example in the SDK team, to want to point to a local reverse proxy for interception.

Turns out that it's already [special-cased in ably-go](https://github.com/ably/ably-go/blob/17f6575773b28a146ac04bfaf3b8de44ce1ff162/ably/options.go#L506-L510), so let's add to spec.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -75,7 +75,7 @@ h2(#endpoint-configuration). Client library endpoint configuration
 ** @(REC1a)@ The @primary domain@ is @main.realtime.ably.net@ unless overridden by specifying an @endpoint@ option or any of the deprecated options @environment@, @restHost@, @realtimeHost@.
 ** @(REC1b)@ If the @endpoint@ option is specified then:
 *** @(REC1b1)@ If any one of the deprecated options @environment@, @restHost@, @realtimeHost@ or @fallbackHostsUseDefault@ are also specified then the options as a set are invalid.
-*** @(REC1b2)@ If the @endpoint@ option is a domain name, determined by it containing at least one period (@.@) then the @primary domain@ is the value of the @endpoint@ option.
+*** @(REC1b2)@ If the @endpoint@ option is a hostname, determined by it containing at least one period (@.@), or containing at least one instance of the string @::@, or being equal to the string @localhost@, then the @primary domain@ is the value of the @endpoint@ option.
 *** @(REC1b3)@ Otherwise, if the @endpoint@ option is a non-production routing policy name, determined by it having the form @nonprod:[id]@, then the @primary domain@ is @[id].realtime.ably-nonprod.net@.
 *** @(REC1b4)@ Otherwise, the @endpoint@ option is a production routing policy name of the form @[id]@, and the @primary domain@ is @[id].realtime.ably.net@.
 ** @(REC1c)@ If the deprecated @environment@ option is specified then it defines a production routing policy name @[id]@:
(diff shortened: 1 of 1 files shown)


textile/features.textile
```diff
@@ -75,7 +75,8 @@ h2(#endpoint-configuration). Client library endpoint configuration
 ** @(REC1a)@ The @primary domain@ is @main.realtime.ably.net@ unless overridden by specifying an @endpoint@ option or any of the deprecated options @environment@, @restHost@, @realtimeHost@.
 ** @(REC1b)@ If the @endpoint@ option is specified then:
 *** @(REC1b1)@ If any one of the deprecated options @environment@, @restHost@, @realtimeHost@ or @fallbackHostsUseDefault@ are also specified then the options as a set are invalid.
-*** @(REC1b2)@ If the @endpoint@ option is a domain name, determined by it containing at least one period (@.@) then the @primary domain@ is the value of the @endpoint@ option.
+** @(REC1b2)@ This clause has been replaced by "@REC1b5@":#REC1b5. It was valid up to and including specification version @TBD@.
```
lmars, line 78: Why not just update the spec point?

lawrence-forooghian: I've always operated under the assumption that if we modify a spec point, even if in a purely additive way, then we should create a new one. Otherwise, if you have code comments that claim that we fully implement a given spec point (which for example we have in [ably-js](https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/CONTRIBUTING.md#tests-alignment-with-the-ably-features-specification) amongst others), these comments will cease to be true.

(I think there are times when we've bent this rule a bit, e.g. by just updating [RTN7a](https://sdk.ably.com/builds/ably/specification/main/features/#RTN7a) when we add a new type of `ProtocolMessage` that requires an `ACK`, but overall I think it's what we do.)

lmars: Hmm maybe not something to decide here, but it seems that's something that would be better handled in the code comment which references the spec point by also referencing the version of the spec the code implements, rather than the spec being littered with tombstones every time we update them?

For example, if a code comment was formatted as:

```
// Implements REC1b2 @ v1.2.3
```

then that's always going to remain true, even if `v1.2.4` of the spec rewrote the spec point.

(and of course, with this scheme, how do you know looking at a spec point with its version in the code, if it's now obsolete because a later version of the spec changed the semantics of the spec point, but that seems about the same as the current status quo of how do you know that a spec point in the code has now been superseded by a "this has been replaced by X" update to the spec point).

lawrence-forooghian: Just wanted to check whether we're coming at this from the same place — are you aware that we do already have a process (created by Quintin) in place for handling spec point modifications?

https://github.com/ably/specification/blob/9801e6e495964906b36fcdba4d604b069e9a617a/CONTRIBUTING.md?plain=1#L48

That's the process that I was following here. I interpreted your original question as being about whether we can apply some nuance when interpreting this rule, but from your second message I'm not sure.

If you're suggesting that we change our process, then I'm happy for us to think about that, but I don't think that this PR is the place to do it. But just to write down my thoughts on your suggestion, in case we do return to this:

- agree that tombstones clutter up the spec; there are other things we could also try for addressing this (e.g. in #295 I suggest having a version of the rendered output that doesn't contain them, or we could store these tombstones in some sidecar yaml file instead of in the spec)
- whilst we have a versioning and release process in place for the spec, the reality is that we haven't really been making much use of it; the last release of the spec was January 2024 (that said, I think that this is because we have not yet really understood the value of having a versioned spec; if there were a compelling reason e.g. what you've suggested above then we might be incentivised to release more often)
- there are lots of other places where we refer to spec points (e.g. Slack) where we wouldn't be able to go back and add this context of a spec version number

lawrence-forooghian: After some discussion about this on Slack, I decided that since ably-go is the only place that the `endpoint` functionality has been implemented, I would just change the spec point implementation and tweak ably-go to match (https://github.com/ably/ably-go/pull/681).

lmars: > are you aware that we do already have a process (created by Quintin) in place for handling spec point modifications?

I was semi-aware, but thanks for linking to it, I guess here I'm questioning whether that process is useful as it stands.

> If you're suggesting that we change our process, then I'm happy for us to think about that, but I don't think that this PR is the place to do it

Sure, like I said, not something to decide here, but I think worthy of discussion, which I'm happy to have elsewhere.

I think the source of the issue is that we had [a rough plan](https://ably.atlassian.net/wiki/spaces/ENG/pages/1505886246/ADR-034+SDK+versioning+policy) to have tooling to track which SDK versions have implemented which spec points from which spec version (e.g. so a tool which could answer "which SDKs have implemented spec point REC1b2 as per spec version 1.2.3), but it was not fully implemented, so we're left with lots of process but very little automation around it, leaving the process seem like busy work.

> Please can you resolve this comment if you're happy?

I'll review the Slack discussion, but yeh sure.

lmars (approved):

PR #303, opened 2025-04-22
Fix language in RTP19

lawrence-forooghian:
"published" makes it sound like we need to send something to Realtime; use "emitted" to make it clear it's a local thing.

(Yes, this spec point could probably do with a broader language overhaul but that wasn't my intention here.)


sacOO7 (approved):
small change but difficult to spot.
Not sure why the `emitted` keyword is not highlighted.
lgtm

PR #308, opened 2025-04-25
Simplify language of `queueMessages` condition

lawrence-forooghian:
Its default value is `true`; if the existing wording is trying to express anything beyond "if it's `true`" then it's not clear to me what.


SimonWoolf (approved):

textile/features.textile
```diff
@@ -830,7 +830,7 @@ h3(#realtime-presence). RealtimePresence
...
 * @(RTP16)@ Connection state conditions:
 ** @(RTP16a)@ If the channel is @ATTACHED@ then the presence messages are handled per @RTL6c2@. (That is: they should be sent to the connection, to be published immediately if the connection is @CONNECTED@, else if the connection state and @queueMessages@ option allows they may be placed in a connection-wide queue to be published once the connection becomes @CONNECTED@, else rejected)
-** @(RTP16b)@ If the channel is @ATTACHING@ or @INITIALIZED@, then if @ClientOptions.queueMessages@ has not been explicitly set to @false@, the presence messages should be queued at a channel level, to be handled per @RTP16a@ once the channel becomes @ATTACHED@, or failed if the channel state becomes @SUSPENDED@, @FAILED@, or @DETACHED@ first
+** @(RTP16b)@ If the channel is @ATTACHING@ or @INITIALIZED@, then if @ClientOptions.queueMessages@ is @true@, the presence messages should be queued at a channel level, to be handled per @RTP16a@ once the channel becomes @ATTACHED@, or failed if the channel state becomes @SUSPENDED@, @FAILED@, or @DETACHED@ first
```
sacOO7, line 833:
```suggestion
** @(RTP16b)@ If the channel is @ATTACHING@ or @INITIALIZED@, then if @ClientOptions.queueMessages@ is @true@(default is true), the presence messages should be queued at a channel level, to be handled per @RTP16a@ once the channel becomes @ATTACHED@, or failed if the channel state becomes @SUSPENDED@, @FAILED@, or @DETACHED@ first
```

textile/features.textile
```diff
@@ -703,7 +703,7 @@ h3(#realtime-channel). RealtimeChannel
...
-*** @(RTL6c2)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@; and the channel is neither @SUSPENDED@ nor @FAILED@; and @ClientOptions#queueMessages@ has not been explicitly set to false; then the message will be placed in a connection-wide message queue to be delivered as soon as the connection is @CONNECTED@. (The recommended implementation is to have the message be sent from the channel to the connection if it fulfils the channel state condition; the connection can then dispatch, queue, or reject it according to its own state and @queueMessages@)
+*** @(RTL6c2)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@; and the channel is neither @SUSPENDED@ nor @FAILED@; and @ClientOptions#queueMessages@ is @true@; then the message will be placed in a connection-wide message queue to be delivered as soon as the connection is @CONNECTED@. (The recommended implementation is to have the message be sent from the channel to the connection if it fulfils the channel state condition; the connection can then dispatch, queue, or reject it according to its own state and @queueMessages@)
```
sacOO7, line 706: I guess the statements were phrased in such a way because default value is `true`. 
But anyways, simplification looks good.

sacOO7, line 706:
```suggestion
*** @(RTL6c2)@ If the connection is @INITIALIZED@, @CONNECTING@ or @DISCONNECTED@; and the channel is neither @SUSPENDED@ nor @FAILED@; and @ClientOptions#queueMessages@ is @true@(default is true); then the message will be placed in a connection-wide message queue to be delivered as soon as the connection is @CONNECTED@. (The recommended implementation is to have the message be sent from the channel to the connection if it fulfils the channel state condition; the connection can then dispatch, queue, or reject it according to its own state and @queueMessages@)
```

lawrence-forooghian: Cool, gonna resolve this comment

lawrence-forooghian: I don't think that in general we write what the default is when we refer to client options. If we do, we should do so consistently. Gonna resolve this conversation, but if you feel there are changes we should make re mentioning default values, feel free to open an issue, thanks!

PR #311, opened 2025-04-25
Clarify when connection state should fail `ACK`-pending messages

lawrence-forooghian:
On my first reading, it wasn't clear to me whether the presence message queue is meant to be directly monitoring the connection state in order to satisfy RTN7e for queued presence messages. After looking at a few implementations (ably-cocoa, ably-js, ably-java), it seems that the intention is that the presence message queue should find out about the connection state change via the channel state change (i.e. RTL11), not by directly monitoring the connection state.


SimonWoolf (approved):

textile/features.textile
```diff
@@ -512,7 +512,7 @@ h3(#realtime-connection). Connection
...
 ** @(RTN7c)@ This clause has been replaced by "@RTN7e@":#RTN7e (it contained a mistake). It was valid up to and including specification version @TBD@.
-** @(RTN7e)@ If a connection enters the @SUSPENDED@, @CLOSED@ or @FAILED@ state, and an @ACK@ or @NACK@ has not yet been received for a message, the client should consider the delivery of those messages as failed, meaning their callback (or language equivalent) should be called with an error representing the reason for the state change, and they should be removed from any @RTN19a@ retry queue
+** @(RTN7e)@ If a connection enters the @SUSPENDED@, @CLOSED@ or @FAILED@ state, and an @ACK@ or @NACK@ has not yet been received for a message submitted to the connection via "@RTL6c1@":#RTL6c1 or "@RTL6c2@":#RTL6c2, the client should consider the delivery of those messages as failed, meaning their callback (or language equivalent) should be called with an error representing the reason for the state change, and they should be removed from any @RTN19a@ retry queue. (Note that @PRESENCE@ messages which, per "@RTP16b@":#RTP16b, have not yet been submitted to the connection are handled by "@RTL11@":#RTL11.)
```
sacOO7, line 515:
```suggestion
** @(RTN7e)@ If a connection enters the @SUSPENDED@, @CLOSED@ or @FAILED@ state, and an @ACK@ or @NACK@ has not yet been received for a message submitted to the connection via "@RTL6c1@":#RTL6c1 or "@RTL6c2@":#RTL6c2, the client should consider the delivery of those messages as failed, meaning their callback (or language equivalent) should be called with an error from the state change reason, and they should be removed from any @RTN19a@ retry queue. (Note that @PRESENCE@ messages which, per "@RTP16b@":#RTP16b, have not yet been submitted to the connection are handled by "@RTL11@":#RTL11.)
```

lawrence-forooghian: I'm sure there are plenty of these sorts of little language things that we could tweak throughout the spec, but I don't really want to get drawn into a back-and-forth on them in a PR that has a specific purpose. Gonna resolve this comment, but if you'd like to raise PRs for this sort of language tweak then they'd be very welcome 🙂

PR #312, opened 2025-04-25
Remove RTL4e

lawrence-forooghian:
It's unclear (how should the SDK determine if this has occurred?) and seems redundant.


SimonWoolf (approved):
lgtm.
(I don't think we need to include "It was valid up to and including specification version..." language for changes which are just improving clarity, not actually changing the expected behaviour)

lawrence-forooghian:
> I don't think we need to include "It was valid up to and including specification version..." language for changes which are just improving clarity, not actually changing the expected behaviour

Sure, have removed

PR #314, opened 2025-04-30
Make `deviceIdentityToken` nullable

lawrence-forooghian:
RSH8c explains when this first gets set.


textile/api-docstrings.md
```diff
@@ -829,8 +829,8 @@ Contains the device identity token and secret of a device. `LocalDevice` extends
 
 | Method / Property | Parameter | Returns | Spec | Description |
 |---|---|---|---|---|
-| deviceIdentityToken: String |||| A unique device identity token used to communicate with APNS or FCM. |
-| deviceSecret: String |||| A unique device secret generated by the Ably SDK. |
+| deviceIdentityToken: String? |||| A unique device identity token used to communicate with APNS or FCM. |
+| deviceSecret: String? |||| A unique device secret generated by the Ably SDK. |
```
ttypic, line 833: I would prefer to `deviceSecret` to be not optional, it's not very well written in the SDK, but device should always have both `id` and `deviceSecret` (they can be potentially initialized lazily, but they have to exist when we work with device)

lawrence-forooghian: Gonna tweak this based on our Zoom call

ttypic (approved):
LGTM

PR #318, opened 2025-05-01
Add side effects of more connection states for `DETACHING` channels

### LINKED ISSUE #242: Make sure there is no attempt to re-attach `DETACHING` channel after connection becomes `SUSPENDED`
See [internal conversation](https://ably-real-time.slack.com/archives/C8SPU4589/p1732272857922879?thread_ts=1732191641.458019&cid=C8SPU4589).

We think the spec does not define what should happen when the following sequence of events occurs:

1. the user calls `detach()`, putting the channel into `DETACHING`;
2. some time after, the connection becomes `SUSPENDED`;
3. some time after, the connection becomes `CONNECTED` again.

In particular, does the channel reattach and start receiving messages again? The desired behaviour is that it not do so. The way in which to make this happen is probably to make sure that the channel ends up in the `DETACHED` state when the connection becomes `SUSPENDED`.

The aims of this task are to:

- [ ] make sure that this behaviour is specified
- [ ] once specified, create SDK repo issues to implement this
- [ ] in particular, implement in JS as a priority

lawrence-forooghian:
Describes what should happen to a `DETACHING` channel when the connection becomes `FAILED`, `CLOSED`, or `SUSPENDED`.

Please take a look at the commit messages, since there are some things where I'd appreciate either confirmation that my reasoning is correct, or better understanding of why things are the way they currently are.

Resolves #242.


ttypic (approved):
LGTM

lawrence-forooghian:
Will return to this to also think about whether the connection error should be propagated to the channel (and making this clear for the other RTL4* spec points that don't make it explicit).

PR #321, opened 2025-05-21
chat: V1 API Review

AndyTWF:
Adds the specification points as a result of CHADR-101, the V1 API review.

[CHA-955]

[CHA-955]: https://ably.atlassian.net/browse/CHA-955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -579,6 +579,10 @@ to typing subscribers.
 ** @(CHA-O6a)@ @[Testable]@ If @occupancy.enableEvents@ is set to @true@, then the client shall set the @options.params.occupancy@ channel parameter to @"metrics"@ on the underlying realtime channel per @CHA-RC3@.
 ** @(CHA-O6b)@ @[Testable]@ If @occupancy.enableEvents@ is set to @false@, then the client shall not set the @options.params.occupancy@ channel parameter.
 ** @(CHA-O6c)@ @[Testable]@ The default value of @occupancy.enableEvents@ is @false@.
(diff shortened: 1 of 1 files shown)


textile/chat-features.textile
```diff
@@ -957,9 +1008,9 @@ h3(#chat-structs-typing-event-V3). Typing Event V3
 
 <pre>
   {
-    "type": "typing.set.changed",
+    "type": TypingSetEventType.SetChanged
```
sacOO7, line 1011: This is supposed to be string value right? Why is it a code instead?

sacOO7: I guess we have assigned similar enum for others too 👍

AndyTWF: Yeah - tried to make it match the enums so its clear

textile/chat-features.textile
```diff
@@ -968,10 +1019,22 @@ h3(#chat-structs-typing-event-V3). Typing Event V3
 
 h3(#chat-structs-occupancy-event). Occupancy Event
 
+<pre>
+  {
+    "type": OccupancyEventType.Updated,
```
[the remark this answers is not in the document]
AndyTWF: We don't define the other enums - but I've now mentioned it where relevant

splindsay-92 (approved):
LGTM

PR #324, opened 2025-05-29
Add tombstones for some deleted spec points

lawrence-forooghian:
These subpoints were deleted in 1ef8d1a without adding a tombstone. This is causing the ably-js spec coverage CI job to fail.

Per our contributing guidelines, we should have tombstones for these spec points, so add them.


lawrence-forooghian:
Relates to https://github.com/ably/ably-js/pull/2038

sacOO7 (approved):
lgtm

PR #327, opened 2025-06-09
Update ObjectData value properties spec to match latest implementation

VeskeR:
This was changed in ably-js PR [1]

[1] https://github.com/ably/ably-js/pull/2026


textile/features.textile
```diff
@@ -1551,13 +1551,19 @@ h4. ObjectData
 * @(OD1)@ An @ObjectData@ represents a value in an object on a channel
 * @(OD2)@ The attributes available in an @ObjectData@ are:
 ** @(OD2a)@ @objectId@ string - a reference to another object
-** @(OD2b)@ @encoding@ string - the encoding the client library should use to interpret the @value@
-** @(OD2c)@ @value@ string, number, boolean or binary - a concrete value of the object
+** @(OD2b)@ @encoding@ string - can be set by the client to indicate that value in @string@ or @bytes@ field have an encoding
+** @(OD2c)@ This clause has been deleted
```
VeskeR, line 1555: I believe there was a discussion recently about how strictly we should adhere to the rules regarding deleting/updating spec points in order to minimaze polluting the spec with "This clause has been deleted" statements. @lawrence-forooghian 
Do you think it makes sense to add this statement now in this case and for [OD3a](https://github.com/ably/specification/pull/327/files#diff-d0e5a641ba4816766d38d6c0e7139df7502bed462dc676227057b4509cafad18R1561)? The development in other SDKs just started but I assume you may already have references to those spec points

sacOO7: I think we should make a separate class for `ObjectValue` that accepts four values in constructor, that way it's easy to interpret, instead of 4 top level fields for value.

VeskeR: The `ObjectData` interface, along with other currently existing object interfaces, describes the protocol message received from realtime. Its structure cannot be easily changed without modifying the protocol version of the spec.

That said, since this is simply the structure of the protocol message received (and what needs to be sent to realtime), I'd it's up to the SDK to decide how best to work this type internally. If you'd like to parse `ObjectData` received from realtime message into a class that contains only the `value` field and use that internally, that's perfectly acceptable. You just need to ensure that, when sending data back to the realtime system, you conform to the `ObjectData` interface - storing the value in one of its typed properties.

sacOO7: But yes, usage at SDK side, we will use format 
```
export interface ObjectData {
  /** A reference to another object, used to support composable object structures. */
  objectId?: string;

  /** Can be set by the client to indicate that value in `string` or `bytes` field have an encoding. */
  encoding?: string;
  
  /** The value stored in this ObjectData. */
  value?: ObjectValue;
}
```

VeskeR: It is used internally only.

In the spec, in the "Interface Definition" section at the bottom where all interfaces are listed, the internal components of the system that should not be exposed in the public API are marketd with [an "internal" tag](https://github.com/ably/specification/blob/9f719f9866dbbc665b956ffda08651a393a0392f/textile/features.textile#L2500), for example:
> class ObjectData // OD*, internal

Currently all interfaces added for Objects are marked as "internal" and none of them will be exposed to the end user, as they only describe the protocol message to communicate with the realtime

sacOO7 (approved):
lgtm

textile/features.textile
sacOO7: I am happy it to be replaced ( true for subsequent spec ), since it's not implemented any other SDKs yet

VeskeR: >I am happy for it to be replaced ( also true for subsequent specs ), since it's not implemented in any other SDKs yet

oh, I see, thanks! You replied before I posted the comment so I didn't see it

VeskeR: will update the PR and merge it

PR #328, opened 2025-06-09
Update internal Objects interface names to have Object*/Objects* prefix

VeskeR:


textile/features.textile
```diff
@@ -1492,7 +1492,7 @@ h4. ObjectState
 ** @(OST3d)@ The size of the @createOp@ property is calculated per "OOP4":#OOP4
 ** @(OST3e)@ The size of a @null@ or omitted property is zero
 
-h4. MapOp
+h4. ObjectMapOp
 
 * @(MOP1)@ A @MapOp@ describes an operation to be applied to a @Map@ object
```
VeskeR, line 1497: whould you also like to have spec point namings updated to start with O*, so something like:
ObjectMapOp - OMO
ObjectCounterOp - OCO
ObjectMap - OMP
ObjectCounter - OCN

@sacOO7 @lawrence-forooghian ?

sacOO7: We can also make all liveobject related spec points to start with `LO*`, then it will be more specific to `LiveObjects`

VeskeR: > We can also make all liveobject related spec points to start with LO*, then it will be more specific to LiveObjects and easy to locate from codebase.

"LiveObjects" is a product name for marketing purposes, not the name of a component of the realtime system.
It was decided in [this DR](https://ably.atlassian.net/wiki/spaces/LOB/pages/3819896841/LODR-033+Isolating+API+naming+from+Product+naming+Renaming+state+and+liveobjects+to+objects+in+APIs) to isolate namings and use "LiveObjects" only in marketing materials when talking about the product. For example, term "LiveObjects" is not used at all in code in the current ably-js's Objects plugin implementation.

We got really used to calling it "LiveObjects" (and it still is, but it's the name of the product, not the API), and we should ensure we don't use it in plugin's implementation in other SDKs

lawrence-forooghian: Sure, am happy with prefixing all LiveObjects stuff with O

textile/features.textile
```diff
@@ -1492,7 +1492,7 @@ h4. ObjectState
 ** @(OST3d)@ The size of the @createOp@ property is calculated per "OOP4":#OOP4
 ** @(OST3e)@ The size of a @null@ or omitted property is zero
 
-h4. MapOp
+h4. ObjectMapOp
```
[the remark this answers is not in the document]
lawrence-forooghian: e.g. `ObjectCounter` reads a bit weird (it counts objects?) — `ObjectsCounter` is still not super clear but if you know that there's a feature called "Objects" then I think it's a bit better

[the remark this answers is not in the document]
sacOO7: In the case of Java, the outcome depends on how you import packages. Ultimately, it will recommend using `Counter` directly with a full-fledged import statement.

textile/features.textile
```diff
@@ -2407,8 +2407,8 @@ enum ObjectOperationAction: // OOP2, internal
   COUNTER_INC // OOP2
   OBJECT_DELETE // OOP2
 
-enum MapSemantics: // MAP2, internal
-  LWW // MAP2
+enum MapSemantics: // OMP2, internal
```
sacOO7, line 2410: This can also be called `ObjectMapSemantics`

sacOO7 (changes requested):
small change requested

textile/features.textile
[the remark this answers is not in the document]
VeskeR: Hm, the only thing that makes me slightly uncomfortable with using the plural `Objects` for a specific type of object is the inconsistency it would introduce between the types used in the spec. Naming everything using the plural `Objects` is also not ideal, as some types - like `ObjectsMessage` - can be misinterpreted as a message belonging to an object (i.e., read as "object's message" with an apostrophe).

So basically:

* All `Object` - sounds a bit odd when referring to a specific type of object.
* All `Objects` - can be misinterpreted as a possessive form, especially for non-specific types.
* A mix of both - inconsistency in naming.

I think I would lean slightly more towards using plural `Objects` everywhere, and just rely on the fact that the reader should first understand that they're reading something related to the feature called Objects, then it shouldn't be really a problem.
I'm not adamant on this, so if you think a mix of singualr and plural will work just fine we can go with that approach.

VeskeR:
As discussed on the standup, I've updated the names for the interfaces that are related to a specific type of an object to use plural prefix Objects* in https://github.com/ably/specification/compare/7ac745d97245dcd8a7f8630f3a09029a09ac9653..fd3ae6feff97474c6c0e9dead07d118bd0562bd5

textile/features.textile
VeskeR: updated in https://github.com/ably/specification/compare/7ac745d97245dcd8a7f8630f3a09029a09ac9653..fd3ae6feff97474c6c0e9dead07d118bd0562bd5

textile/features.textile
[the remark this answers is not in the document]
VeskeR: updated in https://github.com/ably/specification/compare/7ac745d97245dcd8a7f8630f3a09029a09ac9653..fd3ae6feff97474c6c0e9dead07d118bd0562bd5

sacOO7 (approved):
lgtm

PR #333, opened 2025-06-12
[PUB-1041] OBJECT_SYNC spec

VeskeR:
This is the spec equivalent of the object sync sequence handling in ably-js https://github.com/ably/ably-js/blob/main/src/plugins/objects/objects.ts#L215

Resolves [PUB-1041](https://ably.atlassian.net/browse/PUB-1041)

[PUB-1041]: https://ably.atlassian.net/browse/PUB-1041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/textile/objects-features.textile
+++ b/textile/objects-features.textile
@@ -20,7 +20,7 @@ h3(#realtime-objects). RealtimeObjects
 ** @(RTO1b)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 90001
 ** @(RTO1c)@ Waits for the objects sync sequence to complete and for "RTO5c":#RTO5c to finish
 ** @(RTO1d)@ Returns the object with id @root@ from the internal @ObjectsPool@ as a @LiveMap@
-* @(RTO2)@ Various object operations may require a specific channel mode to be set on a channel in order to be performed. If a specific channel mode is required by an operation, then:
+* @(RTO2)@ Certain object operations may require a specific channel mode to be set on a channel in order to be performed. If a specific channel mode is required by an operation, then:
 ** @(RTO2a)@ If the channel is in the @ATTACHED@ state, the presence of the required channel mode is checked against the set of channel modes granted by the server per "RTL4m":../features#RTL4m :
 *** @(RTO2a1)@ If the channel mode is in the set, the operation is allowed
 *** @(RTO2a2)@ If the channel mode is missing, unless otherwise specified by the operation, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 40024, indicating that the operation cannot be performed without the required channel mode
@@ -61,7 +61,7 @@ h3(#realtime-objects). RealtimeObjects
 **** @(RTO5c2a)@ The object with ID @root@ must not be removed from @ObjectsPool@, as per "RTO3b":#RTO3b
 *** @(RTO5c3)@ Clear any stored sync sequence identifiers and cursor values
 *** @(RTO5c4)@ The @SyncObjectsPool@ must be cleared
-* @(RTO6)@ When needed, a zero-value object can be created if it does not exist in the internal @ObjectsPool@ for an @objectId@, in the following way:
+* @(RTO6)@ Certain object operations may require creating a zero-value object if one does not already exist in the internal @ObjectsPool@ for the given @objectId@. This can be done as follows:
 ** @(RTO6a)@ If an object with @objectId@ exists in @ObjectsPool@, do not create a new object
 ** @(RTO6b)@ The expected type of the object can be inferred from the provided @objectId@:
 *** @(RTO6b1)@ Split the @objectId@ (formatted as @type:hash&#64;timestamp@) on the separator @:@ and parse the first part as the type string


textile/objects-features.textile
```diff
@@ -0,0 +1,137 @@
...
+*** @(OBJ5c3)@ Clear any stored sync sequence identifiers and cursor values
+*** @(OBJ5c4)@ @SyncObjectsPool@ must be emptied
+* @(OBJ6)@ When needed, a zero-value object can be created if it does not exist in the internal @ObjectsPool@ for an @objectId@ in the following way:
+** @(OBJ6a)@ If an object with @objectId@ exists in @ObjectsPool@ do not create a new object
+** @(OBJ6b)@ The expected type of the object can be inferred from the provided @objectId@:
+*** @(OBJ6b1)@ Split @objectId@ (which is in the format @type:hash&#64;timestamp@) on the separator @:@ and parse the first part as the string
+*** @(OBJ6b2)@ If parsed type is @map@ create a zero-value @LiveMap@ per "OLM4":OLM4 in @ObjectsPool@
+*** @(OBJ6b3)@ If parsed type is @counter@ create a zero-value @LiveCounter@ per "OLC4":OLC4 in @ObjectsPool@
+
+h3(#livecounter). LiveCounter
```
sacOO7, line 61: I guess this can be called `ObjectLiveCounter` to keep naming in sync with models https://github.com/ably/specification/pull/328

VeskeR: This time around `LiveCounter`, `LiveMap` are actually the types that are exposed to the end user, so they can't be changed to `ObjectLiveCounter`

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+*** @(RTLM7a1)@ If the operation cannot be applied as per "RTLM9":#RTLM9, discard the operation without taking any action
+*** @(RTLM7a2)@ Otherwise, apply the operation:
+**** @(RTLM7a2a)@ Set @MapEntry.data@ to the @ObjectData@ from the operation
+**** @(RTLM7a2b)@ Set @MapEntry.timeserial@ to the operation's serial
+**** @(RTLM7a2c)@ Set @MapEntry.tombstone@ to @false@
+** @(RTLM7b)@ If an entry does not exist in the private @data@ for the specified key:
+*** @(RTLM7b1)@ Create a new entry in @data@ for the specified key with the provided @ObjectData@ and the operation's serial
+*** @(RTLM7b2)@ Set @MapEntry.tombstone@ for the new entry to @false@
+** @(RTLM7c)@ If the operation has a non-empty @ObjectData.objectId@ attribute:
+*** @(RTLM7c1)@ Create a zero-value @LiveObject@ in the internal @ObjectsPool@ per "RTO6":#RTO6
```
lawrence-forooghian, line 129:
```suggestion
*** @(RTLM7c1)@ Create a zero-value @LiveObject@ for this @objectId@ in the internal @ObjectsPool@ per "RTO6":#RTO6
```

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+* @(RTLC1)@ The @LiveCounter@ extends @LiveObject@
+* @(RTLC2)@ Represents the counter object type for Object IDs of type @counter@
+* @(RTLC3)@ Holds a 64-bit floating-point number as a private @data@
+* @(RTLC4)@ The zero-value @LiveCounter@ is a @LiveCounter@ with @data@ set to 0
+* @(RTLC5)@ @LiveCounter#value@ function:
+** @(RTLC5a)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTLC5b)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should indicate an error with code 90001
+** @(RTLC5c)@ Returns the current @data@ value
+* @(RTLC6)@ @LiveCounter@'s internal @data@ can be overridden with the provided @ObjectState@ in the following way:
+** @(RTLC6a)@ Replace the private @siteTimeserials@ of the @LiveCounter@ with the value from @ObjectState.siteTimeserials@
```
lawrence-forooghian, line 80: The spec doesn't describe the initial values of the private properties that it instructs us to set (`siteTimeserials`, `createOperationIsMerged`, `semantics`). Should it? (I don't know how we're going to _use_ these properties yet so can't say.)

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+h2(#overview). Overview
+
+This document outlines the feature specification for the Objects feature of the Realtime system. It is currently under development and stored separately from the main specification to simplify the initial implementation of the feature in other SDKs. Once completed, it will be moved to the main "features":../features spec.
+
+Objects feature enables clients to store shared data as "objects" on a channel. When an object is updated, changes are automatically propagated to all subscribed clients in realtime, ensuring each client always sees the latest state.
+
+h3(#realtime-objects). RealtimeObjects
+
+* @(RTO1)@ @Objects#getRoot@ function:
+** @(RTO1a)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
```
sacOO7, line 19: If `OBJECT_SUBSCRIBE` mode is not granted, this should throw error right?

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+** @(RTO1b)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should indicate an error with code 90001
```
owenpearson, line 20: What if the channel is initialised, detaching, suspended, etc?

I think it would be better to say "If the channel is not ATTACHED[ or ATTACHING]...", since these are probably the only two cases where we would expect the operation to work. If we want it to work when ATTACHING we need to spec how the operation will be queued.

The other thing we need to spec is whether the operation should implicitly attach an INITIALIZED channel

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+** @(RTLM5b)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTLM5c)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should indicate an error with code 90001
+** @(RTLM5d)@ Returns the value from the current @data@ at the specified key, as follows:
+*** @(RTLM5d1)@ If no @ObjectsMapEntry@ exists at the key, return undefined/null
+*** @(RTLM5d2)@ If a @ObjectsMapEntry@ exists at the key:
+**** @(RTLM5d2a)@ If @ObjectsMapEntry.tombstone@ is @true@, return undefined/null
+**** @(RTLM5d2b)@ If @ObjectsMapEntry.data.boolean@ exists, return it
+**** @(RTLM5d2c)@ If @ObjectsMapEntry.data.bytes@ exists, return it
+**** @(RTLM5d2d)@ If @ObjectsMapEntry.data.number@ exists, return it
+**** @(RTLM5d2e)@ If @ObjectsMapEntry.data.string@ exists, return it
```
lawrence-forooghian, line 104: Per OD5b3, the contents of `string` may be a JSON object or array:

> *** @(OD5b3)@ If @ObjectData.encoding@ is set to "json", the @ObjectData.string@ content is decoded by parsing the string as JSON

But, at least in the JS type declarations and documentation, it seems like `LiveMap.get` cannot return a JSON-like object: https://github.com/ably/ably-js/blob/6f2b7e654921dbda08e178939b1bfd230fd78d4a/ably.d.ts#L2518

Should it be able to?

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+* @(RTLC2)@ Represents the counter object type for Object IDs of type @counter@
+* @(RTLC3)@ Holds a 64-bit floating-point number as a private @data@
+* @(RTLC4)@ The zero-value @LiveCounter@ is a @LiveCounter@ with @data@ set to 0
+* @(RTLC5)@ @LiveCounter#value@ function:
+** @(RTLC5a)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTLC5b)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should indicate an error with code 90001
+** @(RTLC5c)@ Returns the current @data@ value
+* @(RTLC6)@ @LiveCounter@'s internal @data@ can be overridden with the provided @ObjectState@ in the following way:
+** @(RTLC6a)@ Replace the private @siteTimeserials@ of the @LiveCounter@ with the value from @ObjectState.siteTimeserials@
+** @(RTLC6b)@ Set the private flag @createOperationIsMerged@ to @false@
```
sacOO7, line 81: Based on the SDK `createOperationIsMerged` is supposed to be thread safe is operations performed from different threads right

textile/objects-features.textile
sacOO7: Seems, it's mentioned as a part of `RTO2`

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+** @(RTO3a)@ @ObjectsPool@ is a @Dict<String, LiveObject>@ - a map of @LiveObject@s keyed by "@objectId@":../features#OST2a string
+** @(RTO3b)@ It must always contain a @LiveMap@ object with id @root@
+* @(RTO4)@ When a channel @ATTACHED@ @ProtocolMessage@ is received, the @ProtocolMessage@ may contain a @HAS_OBJECTS@ bit flag indicating that it will perform an objects sync, see "TR3":../features#TR3 . Note that this does not imply that objects are definitely present on the channel, only that there may be; the @OBJECT_SYNC@ message may be empty
+** @(RTO4a)@ If the @HAS_OBJECTS@ flag is 1, the server will shortly perform an @OBJECT_SYNC@ sequence as described in "RTO5":#RTO5
+** @(RTO4b)@ If the @HAS_OBJECTS@ flag is 0 or there is no @flags@ field, the sync sequence must be considered complete immediately, and the client library must perform the following actions in order:
+*** @(RTO4b1)@ All objects except the one with id @root@ must be removed from the internal @ObjectsPool@
+*** @(RTO4b2)@ The data for the @LiveMap@ with id @root@ must be cleared by setting it to a zero-value per "RTLM4":#RTLM4
+*** @(RTO4b3)@ The @SyncObjectsPool@ must be cleared
```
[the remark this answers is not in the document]
sacOO7: Seems like a temporary array to hold intermediate object state updates.

textile/objects-features.textile
sacOO7: `granted` doesn't feel like right keyword to use here

textile/objects-features.textile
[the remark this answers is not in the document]
sacOO7: I think it's necessary to define what type `SyncObjectsPool` / `SyncObjectsDataPool` represents. At first, I thought it was a boolean flag that should be cleared.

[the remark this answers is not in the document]
sacOO7: In ably-js it's `Map<string, AnyDataEntry>`. Can we mention somewhere that it's a map of `objectId` to `ObjectState`.

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+**** @(RTO5a2a)@ The current @SyncObjectsPool@ list must be cleared
+*** @(RTO5a3)@ If the sequence id matches the previously received sequence id, the client library should continue the sync process
+*** @(RTO5a4)@ The objects sync sequence for that sequence identifier is considered complete once the cursor is empty; that is when the @channelSerial@ looks like @<sequence id>:@
+*** @(RTO5a5)@ An @OBJECT_SYNC@ may also be sent with no @channelSerial@ attribute. In this case, the sync data is entirely contained within the @ProtocolMessage@
+** @(RTO5b)@ During the sync sequence, the @ObjectMessage.object@ values from incoming @OBJECT_SYNC@ @ProtocolMessage@s must be temporarily stored in the internal @SyncObjectsPool@ list
+** @(RTO5c)@ When the objects sync has completed, the client library must perform the following actions in order:
+*** @(RTO5c1)@ For each @ObjectState@ member in the @SyncObjectsPool@ list:
+**** @(RTO5c1a)@ If an object with @ObjectState.objectId@ exists in the internal @ObjectsPool@:
+***** @(RTO5c1a1)@ Override the internal data for the object as per "RTLC6":#RTLC6, "RTLM6":#RTLM6
```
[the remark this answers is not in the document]
sacOO7: Yeah, `replace` seems like a right word for this 
```
Replace the internal data for the object with updated data as per "RTLC6":#RTLC6, "RTLM6":#RTLM6
```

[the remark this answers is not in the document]
sacOO7: There's also technical term called `Upsert` to represent `update` or `insert` op. 
Maybe it's a right fit, since we are not always `replacing` but can be inserting entries if not present.

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+*** @(RTO6b3)@ If the parsed type is @counter@, create a zero-value @LiveCounter@ per "RTLC4":#RTLC4 in the @ObjectsPool@
+
+h3(#livecounter). LiveCounter
+
+* @(RTLC1)@ The @LiveCounter@ extends @LiveObject@
+* @(RTLC2)@ Represents the counter object type for Object IDs of type @counter@
+* @(RTLC3)@ Holds a 64-bit floating-point number as a private @data@
+* @(RTLC4)@ The zero-value @LiveCounter@ is a @LiveCounter@ with @data@ set to 0
+* @(RTLC5)@ @LiveCounter#value@ function:
+** @(RTLC5a)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
```
sacOO7, line 76:
```suggestion
** @(RTLC5a)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be validated per "RTO2":#RTO2
```

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+
+h3(#livemap). LiveMap
+
+* @(RTLM1)@ The @LiveMap@ extends @LiveObject@
+* @(RTLM2)@ Represents the map object type for Object IDs of type @map@
+* @(RTLM3)@ Holds a @Dict<String, ObjectsMapEntry>@ as a private @data@ map
+* @(RTLM4)@ The zero-value @LiveMap@ is a @LiveMap@ with @data@ set to an empty map
+* @(RTLM5)@ @LiveMap#get@ function:
+** @(RTLM5a)@ Accepts a key of type String
+** @(RTLM5b)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
```
sacOO7, line 95:
```suggestion
** @(RTLM5b)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be validated per "RTO2":#RTO2
```

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+*** @(RTO2b1)@ If the channel mode is in the set, the operation is allowed
+*** @(RTO2b2)@ If the channel mode is missing, unless otherwise specified by the operation, the library should indicate an error with code 40024 stating that the operation cannot be performed without the required channel mode
+* @(RTO3)@ An internal @ObjectsPool@ should be used to maintain the list of objects present on a channel
+** @(RTO3a)@ @ObjectsPool@ is a @Dict<String, LiveObject>@ - a map of @LiveObject@s keyed by "@objectId@":../features#OST2a string
+** @(RTO3b)@ It must always contain a @LiveMap@ object with id @root@
+* @(RTO4)@ When a channel @ATTACHED@ @ProtocolMessage@ is received, the @ProtocolMessage@ may contain a @HAS_OBJECTS@ bit flag indicating that it will perform an objects sync, see "TR3":../features#TR3 . Note that this does not imply that objects are definitely present on the channel, only that there may be; the @OBJECT_SYNC@ message may be empty
+** @(RTO4a)@ If the @HAS_OBJECTS@ flag is 1, the server will shortly perform an @OBJECT_SYNC@ sequence as described in "RTO5":#RTO5
```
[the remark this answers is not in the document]
sacOO7: In ably-js, we check for null `protocolMessage.state`  => https://github.com/ably/ably-js/blob/39e06773b45a3bb606ca5482450019e2c2f8c77b/src/common/lib/client/realtimechannel.ts#L619
Seems we should also include empty check there

sacOO7 (commented):
Can we also add spec point somewhere to [populateMissingFieldsFromParent](https://github.com/ably/ably-js/blob/39e06773b45a3bb606ca5482450019e2c2f8c77b/src/common/lib/client/realtimechannel.ts#L623) e.g. fields like `objectMsg.connectionId`, `objectMsg.timestamp` and `objectMsg.id`

textile/objects-features.textile
lawrence-forooghian: Just a general note on this one — our specs have always assumed that they're operating in a single-threaded world and just left it up to individual platforms to figure out how to work around that. It might not be ideal but I don't think we should try and address that here.

lawrence-forooghian:
> Can we also add spec point somewhere to [populateMissingFieldsFromParent](https://github.com/ably/ably-js/blob/39e06773b45a3bb606ca5482450019e2c2f8c77b/src/common/lib/client/realtimechannel.ts#L623) e.g. fields like `objectMsg.connectionId`, `objectMsg.timestamp` and `objectMsg.id`

Don't we already have that? OM2c etc

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+** @(RTO4b)@ If the @HAS_OBJECTS@ flag is 0 or there is no @flags@ field, the sync sequence must be considered complete immediately, and the client library must perform the following actions in order:
+*** @(RTO4b1)@ All objects except the one with id @root@ must be removed from the internal @ObjectsPool@
+*** @(RTO4b2)@ The data for the @LiveMap@ with id @root@ must be cleared by setting it to a zero-value per "RTLM4":#RTLM4
```
lawrence-forooghian, line 37: This one is a bit unclear — do we _replace_ the root `LiveMap` or just manipulate its internal data?

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+** @(RTO5c)@ When the objects sync has completed, the client library must perform the following actions in order:
+*** @(RTO5c1)@ For each @ObjectState@ member in the @SyncObjectsPool@ list:
+**** @(RTO5c1a)@ If an object with @ObjectState.objectId@ exists in the internal @ObjectsPool@:
+***** @(RTO5c1a1)@ Override the internal data for the object as per "RTLC6":#RTLC6, "RTLM6":#RTLM6
+**** @(RTO5c1b)@ If an object with @ObjectState.objectId@ does not exist in the internal @ObjectsPool@:
+***** @(RTO5c1b1)@ Create a new @LiveObject@ using the data from @ObjectState@ and add it to the internal @ObjectsPool@:
+****** @(RTO5c1b1a)@ If @ObjectState.counter@ is present, create a zero-value @LiveCounter@ (per "RTLC4":#RTLC4), set its private @objectId@ equal to @ObjectState.objectId@ and override its internal data using the current @ObjectState@ per "RTLC6":#RTLC6
+****** @(RTO5c1b1b)@ If @ObjectState.map@ is present, create a zero-value @LiveMap@ (per "RTLM4":#RTLM4), set its private @objectId@ equal to @ObjectState.objectId@, set its private @semantics@ equal to @ObjectState.map.semantics@ and override its internal data using the current @ObjectState@ per "RTLM6":#RTLM6
```
sacOO7, line 56: I'm not sure why we refer to them as `zeroValueObjects` when we're just creating instances of `LiveMap` and `LiveCounter`. In essence, both are just empty instances.

Calling them zeroValueObjects seems to imply some special significance, which might be misleading in terms of their actual behavior.

```suggestion
****** @(RTO5c1b1b)@ If @ObjectState.map@ is present, create a empty @LiveMap@ (per "RTLM4":#RTLM4) instance, set its private @objectId@ equal to @ObjectState.objectId@, set its private @semantics@ equal to @ObjectState.map.semantics@ and override its internal data using the current @ObjectState@ per "RTLM6":#RTLM6
```

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+****** @(RTO5c1b1c)@ Otherwise, log a warning that an unsupported object state message has been received, and discard the current @ObjectState@ without taking any action
+*** @(RTO5c2)@ Remove any objects from the internal @ObjectsPool@ for which @objectId@s were not received during the sync sequence
+**** @(RTO5c2a)@ The object with ID @root@ must not be removed from @ObjectsPool@, as per "RTO3b":#RTO3b
+*** @(RTO5c3)@ Clear any stored sync sequence identifiers and cursor values
```
sacOO7, line 60: We only need to store `syncId` and doesn't need to store `syncCursor` value.
```suggestion
*** @(RTO5c3)@ Clear stored sync sequence identifier.
```
Even in case of ably-js, [`syncCursor` value is unused ](https://github.com/ably/ably-js/blob/39e06773b45a3bb606ca5482450019e2c2f8c77b/src/plugins/objects/objects.ts#L51)

textile/objects-features.textile
sacOO7, line 34: In ably-js, we check if objects state is `INITIALIZED`, so irrespective of `hasObjects`, we always call `startNewSync` in such a case, can we include this case here?

textile/objects-features.textile
sacOO7: Just a note, after clearing `root`, it should emit respective updates, can be covered as a part of new spec.

textile/objects-features.textile
lawrence-forooghian: Seems this is addressed by https://github.com/ably/ably-js/pull/2052

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+
+h2(#overview). Overview
+
+This document outlines the feature specification for the Objects feature of the Realtime system. It is currently under development and stored separately from the main specification to simplify the initial implementation of the feature in other SDKs. Once completed, it will be moved to the main "features":../features spec.
+
+Objects feature enables clients to store shared data as "objects" on a channel. When an object is updated, changes are automatically propagated to all subscribed clients in realtime, ensuring each client always sees the latest state.
+
+h3(#realtime-objects). RealtimeObjects
+
+* @(RTO1)@ @Objects#getRoot@ function:
```
lawrence-forooghian, line 18: I have been caught out a few times by this method call hanging because I forgot to attach the channel. Does JS implement any sort of implicit attach?

sacOO7: Currently, we only throw error for `ChannelState.detached` and `ChannelState.failed`, I feel we should also check for `ChannelState.initialized` state. I guess, we do the same in case of chat-sdk

lawrence-forooghian: I see objects as a feature of the core SDK, on the same level as, say, presence. We have presence features that implicitly perform an attach (e.g. `enter`).

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+** @(RTO1a)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTO1b)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should indicate an error with code 90001
+** @(RTO1c)@ Waits for the objects sync sequence to complete and for "RTO5c":#RTO5c to finish
```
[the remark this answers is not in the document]
VeskeR: Created https://github.com/ably/specification/issues/369 for this, as it looks like we need to update `getRoot` to handle implicit attach and make it clear how it handles different states of sync sequence

textile/objects-features.textile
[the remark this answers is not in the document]
VeskeR: going to address `SyncObjectsPool` in https://github.com/ably/specification/pull/370, it needs to be renamed to more clearly indicate its purpose and will clarify its type in a dedicated spec item.
(also note that ably-js implementation with `SyncObjectsDataPool` is excessive at this point, no need to have a separate class for this and store message in a `objectId -> ObjectState` map, a simple list should suffice just like spec suggests)

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+* @(RTLM6)@ @LiveMap@ internal @data@ can be overridden with the provided @ObjectState@ in the following way:
+** @(RTLM6a)@ Replace the private @siteTimeserials@ of the @LiveMap@ with the value from @ObjectState.siteTimeserials@
+** @(RTLM6b)@ Set the private flag @createOperationIsMerged@ to @false@
+** @(RTLM6c)@ Set @data@ to @ObjectState.map.entries@, or to an empty map if it does not exist
+** @(RTLM6d)@ If @ObjectState.createOp@ is present:
+*** @(RTLM6d1)@ For each key–@MapEntry@ pair in @ObjectState.createOp.map.entries@:
+**** @(RTLM6d1a)@ If @MapEntry.tombstone@ is @false@, apply the @MAP_SET@ operation to the specified key using @MapEntry.timeserial@ and @MapEntry.data@ per "RTLM7":#RTLM7
+**** @(RTLM6d1b)@ If @MapEntry.tombstone@ is @true@, apply the @MAP_REMOVE@ operation to the specified key using @MapEntry.timeserial@ per "RTLM8":#RTLM8
+*** @(RTLM6d2)@ Set the private flag @createOperationIsMerged@ to @true@
+* @(RTLM7)@ @MAP_SET@ operation for a key can be applied to a @LiveMap@ in the following way:
```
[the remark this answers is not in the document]
VeskeR: this has been done in https://github.com/ably/specification/commit/f12d5ba6108ac320302ccb486fcb6035e63d84a5

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+** @(RTLM7a)@ If an entry exists in the private @data@ for the specified key:
+*** @(RTLM7a1)@ If the operation cannot be applied as per "RTLM9":#RTLM9, discard the operation without taking any action
+*** @(RTLM7a2)@ Otherwise, apply the operation:
```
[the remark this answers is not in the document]
VeskeR: Well, sort of — but it also includes more specific assignments, such as `RTLM7a2b`, where we set `ObjectsMapEntry.timeserial` to the operation's `serial` — which is not present on the `ObjectsMapEntry` object itself from `RTLM6d1a`. It also explicitly instructs setting `tombstone` to `false` (which may be omitted in the Realtime-received entry), and in the later spec PRs, it sets a `tombstonedAt` property as well.

Given all that, I believe it makes sense to have a dedicated `RTLM7a2` spec item to specify how we should apply the map entry from Realtime to the local state.

textile/objects-features.textile
VeskeR: described in later PRs: https://github.com/ably/specification/commit/f12d5ba6108ac320302ccb486fcb6035e63d84a5, https://github.com/ably/specification/pull/361

textile/objects-features.textile
VeskeR: > Just a note, after clearing root, it should emit respective updates, can be covered as a part of new spec.

`RTO4b2a` was added for this in https://github.com/ably/specification/commit/29633002c947838223c8ce6bc580758448af8eb7

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+*** @(RTO2a1)@ If the channel mode is in the set, the operation is allowed
+*** @(RTO2a2)@ If the channel mode is missing, unless otherwise specified by the operation, the library should indicate an error with code 40024 stating that the operation cannot be performed without the required channel mode
+** @(RTO2b)@ Otherwise, a best-effort attempt is made, and the channel mode is checked against the set of channel modes requested by the user per "TB2d":../features#TB2d :
+*** @(RTO2b1)@ If the channel mode is in the set, the operation is allowed
+*** @(RTO2b2)@ If the channel mode is missing, unless otherwise specified by the operation, the library should indicate an error with code 40024 stating that the operation cannot be performed without the required channel mode
+* @(RTO3)@ An internal @ObjectsPool@ should be used to maintain the list of objects present on a channel
+** @(RTO3a)@ @ObjectsPool@ is a @Dict<String, LiveObject>@ - a map of @LiveObject@s keyed by "@objectId@":../features#OST2a string
+** @(RTO3b)@ It must always contain a @LiveMap@ object with id @root@
```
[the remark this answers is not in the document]
VeskeR: clarified in https://github.com/ably/specification/pull/333/commits/bfcf05b9c86289d1bd3340a2d95ae136a4066b3a

textile/objects-features.textile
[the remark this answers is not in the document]
VeskeR: changed to use "replace" in https://github.com/ably/specification/commit/bfcf05b9c86289d1bd3340a2d95ae136a4066b3a

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+*** @(RTLM8a1)@ If the operation cannot be applied as per "RTLM9":#RTLM9, discard the operation without taking any action
+*** @(RTLM8a2)@ Otherwise, apply the operation:
+**** @(RTLM8a2a)@ Set @MapEntry.data@ to undefined/null
+**** @(RTLM8a2b)@ Set @MapEntry.timeserial@ to the operation's serial
+**** @(RTLM8a2c)@ Set @MapEntry.tombstone@ to @true@
+** @(RTLM8b)@ If an entry does not exist in the private @data@ for the specified key:
+*** @(RTLM8b1)@ Create a new entry in @data@ for the specified key, with @MapEntry.data@ set to undefined/null and the operation's serial
+*** @(RTLM8b2)@ Set @MapEntry.tombstone@ for the new entry to @true@
+* @(RTLM9)@ Whether a map operation can be applied to a map entry is determined as follows:
+** @(RTLM9a)@ For a @LiveMap@ using @LWW@ (Last-Write-Wins) CRDT semantics, the operation must only be applied if its serial is strictly greater ("after") than the entry's serial when compared lexicographically
```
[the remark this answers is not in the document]
VeskeR: updated phrasing in https://github.com/ably/specification/commit/bfcf05b9c86289d1bd3340a2d95ae136a4066b3a

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+**** @(RTLM7a2b)@ Set @MapEntry.timeserial@ to the operation's serial
+**** @(RTLM7a2c)@ Set @MapEntry.tombstone@ to @false@
+** @(RTLM7b)@ If an entry does not exist in the private @data@ for the specified key:
+*** @(RTLM7b1)@ Create a new entry in @data@ for the specified key with the provided @ObjectData@ and the operation's serial
+*** @(RTLM7b2)@ Set @MapEntry.tombstone@ for the new entry to @false@
+** @(RTLM7c)@ If the operation has a non-empty @ObjectData.objectId@ attribute:
+*** @(RTLM7c1)@ Create a zero-value @LiveObject@ in the internal @ObjectsPool@ per "RTO6":#RTO6
+* @(RTLM8)@ @MAP_REMOVE@ operation for a key can be applied to a @LiveMap@ in the following way:
+** @(RTLM8a)@ If an entry exists in the private @data@ for the specified key:
+*** @(RTLM8a1)@ If the operation cannot be applied as per "RTLM9":#RTLM9, discard the operation without taking any action
```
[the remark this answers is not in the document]
VeskeR: updated in https://github.com/ably/specification/commit/bfcf05b9c86289d1bd3340a2d95ae136a4066b3a

textile/objects-features.textile
VeskeR: updated in https://github.com/ably/specification/commit/bfcf05b9c86289d1bd3340a2d95ae136a4066b3a

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+**** @(RTLM8a2a)@ Set @ObjectsMapEntry.data@ to undefined/null
+**** @(RTLM8a2b)@ Set @ObjectsMapEntry.timeserial@ to the operation's serial
+**** @(RTLM8a2c)@ Set @ObjectsMapEntry.tombstone@ to @true@
+** @(RTLM8b)@ If an entry does not exist in the private @data@ for the specified key:
+*** @(RTLM8b1)@ Create a new entry in @data@ for the specified key, with @ObjectsMapEntry.data@ set to undefined/null and the operation's serial
+*** @(RTLM8b2)@ Set @ObjectsMapEntry.tombstone@ for the new entry to @true@
+* @(RTLM9)@ Whether a map operation can be applied to a map entry is determined as follows:
+** @(RTLM9a)@ For a @LiveMap@ using @LWW@ (Last-Write-Wins) CRDT semantics, the operation must only be applied if its serial is strictly greater ("after") than the entry's serial when compared lexicographically
+** @(RTLM9b)@ If both the entry serial and the operation serial are null or empty strings, they are treated as the "earliest possible" serials and considered "equal", so the operation must not be applied
+** @(RTLM9c)@ If only the entry serial exists, the missing operation serial is considered lower than the existing entry serial, so the operation must not be applied
```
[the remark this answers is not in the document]
VeskeR: should also be "not an empty string", updated in https://github.com/ably/specification/commit/bfcf05b9c86289d1bd3340a2d95ae136a4066b3a

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+
+* @(RTO1)@ @Objects#getRoot@ function:
+** @(RTO1a)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTO1b)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should indicate an error with code 90001
+** @(RTO1c)@ Waits for the objects sync sequence to complete and for "RTO5c":#RTO5c to finish
+** @(RTO1d)@ Returns the object with id @root@ from the internal @ObjectsPool@ as a @LiveMap@
+* @(RTO2)@ Various object operations may require a specific channel mode to be set on a channel in order to be performed. If a specific channel mode is required by an operation, then:
+** @(RTO2a)@ If the channel is in the @ATTACHED@ state, the presence of the required channel mode is checked against the set of channel modes granted by the server per "RTL4m":../features#RTL4m :
+*** @(RTO2a1)@ If the channel mode is in the set, the operation is allowed
+*** @(RTO2a2)@ If the channel mode is missing, unless otherwise specified by the operation, the library should indicate an error with code 40024 stating that the operation cannot be performed without the required channel mode
```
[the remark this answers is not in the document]
VeskeR: Might affect multiple spec items, and I also think that those restrictions you had to put for a `subscribe()` call in annotations were the consequence of `subscribe()` doing an implicit attach. Created a separate https://github.com/ably/specification/issues/371 to think more about this, and will need to change ably-js implementation as the result of that too

textile/objects-features.textile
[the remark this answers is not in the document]
lawrence-forooghian: OK, feel free to resolve this one

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+*** @(RTO5a1)@ The @channelSerial@ is used as the sync cursor and is a two-part identifier: @<sequence id>:<cursor value>@
+*** @(RTO5a2)@ If a new sequence id is sent from Ably, the client library must treat it as the start of a new objects sync sequence, and any previous in-flight sync must be discarded:
+**** @(RTO5a2a)@ The current @SyncObjectsPool@ list must be cleared
+*** @(RTO5a3)@ If the sequence id matches the previously received sequence id, the client library should continue the sync process
+*** @(RTO5a4)@ The objects sync sequence for that sequence identifier is considered complete once the cursor is empty; that is when the @channelSerial@ looks like @<sequence id>:@
+*** @(RTO5a5)@ An @OBJECT_SYNC@ may also be sent with no @channelSerial@ attribute. In this case, the sync data is entirely contained within the @ProtocolMessage@
+** @(RTO5b)@ During the sync sequence, the @ObjectMessage.object@ values from incoming @OBJECT_SYNC@ @ProtocolMessage@s must be temporarily stored in the internal @SyncObjectsPool@ list
```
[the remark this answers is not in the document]
lawrence-forooghian: Resolving

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+*** @(RTLM6d1)@ For each key–@MapEntry@ pair in @ObjectState.createOp.map.entries@:
+**** @(RTLM6d1a)@ If @MapEntry.tombstone@ is @false@, apply the @MAP_SET@ operation to the specified key using @MapEntry.timeserial@ and @MapEntry.data@ per "RTLM7":#RTLM7
+**** @(RTLM6d1b)@ If @MapEntry.tombstone@ is @true@, apply the @MAP_REMOVE@ operation to the specified key using @MapEntry.timeserial@ per "RTLM8":#RTLM8
+*** @(RTLM6d2)@ Set the private flag @createOperationIsMerged@ to @true@
+* @(RTLM7)@ @MAP_SET@ operation for a key can be applied to a @LiveMap@ in the following way:
+** @(RTLM7a)@ If an entry exists in the private @data@ for the specified key:
+*** @(RTLM7a1)@ If the operation cannot be applied as per "RTLM9":#RTLM9, discard the operation without taking any action
+*** @(RTLM7a2)@ Otherwise, apply the operation:
+**** @(RTLM7a2a)@ Set @MapEntry.data@ to the @ObjectData@ from the operation
+**** @(RTLM7a2b)@ Set @MapEntry.timeserial@ to the operation's serial
```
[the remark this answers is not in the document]
lawrence-forooghian: The thing is that, to take one example: as it currently reads, RTLM6d1a mentions passing `ObjectsMapEntry.timeserial` and RTLM7a2b mentions the "operation's serial" but it's not obvious to the reader that these are the same thing (like, you can kind of inituit that they are because the `timeserial` value has `serial` in its name so they're probably the same thing, but it's not super clear)

textile/objects-features.textile
lawrence-forooghian: OK, feel free to resolve this one

textile/objects-features.textile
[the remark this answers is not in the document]
VeskeR: no, it's part of the https://github.com/ably/specification/pull/343

textile/objects-features.textile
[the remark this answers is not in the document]
VeskeR: I see.
I think the best approach here would be to refer explicitly to the `@serial@` param the `RTLM7` accepts, which is more clearly defined in https://github.com/ably/specification/pull/343 PR here: https://github.com/ably/specification/pull/343/files#diff-c598156b2e1e7159ee7addc9b73d820e220a41135e64bbde772aa5b838340f25R214-R217.

So `RTLM7a2b` will read:
```
**** @(RTLM7a2b)@ Set @MapEntry.timeserial@ to the @serial@
```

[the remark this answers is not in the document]
VeskeR: uhh, I guess given your comment below in https://github.com/ably/specification/pull/333#discussion_r2152354953, I'll just move arguments for those methods into this PR so it's all in one place

textile/objects-features.textile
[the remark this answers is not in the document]
VeskeR: I have ported better phrasing (and arguments list) from https://github.com/ably/specification/pull/343 into this PR, see https://github.com/ably/specification/pull/333/commits/027ec46be34c883bba743c5f0f612e790597b36c.

It should be now clear that `RTLM7` and `RTLM8` use the `@serial@` value that was provided to them as an argument, no matter where the value came from.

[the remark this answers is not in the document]
lawrence-forooghian: Cool, looks good!

lawrence-forooghian (approved):

textile/objects-features.textile
VeskeR: see https://github.com/ably/specification/pull/333#discussion_r2325013765

textile/objects-features.textile
VeskeR: see https://github.com/ably/specification/pull/333#discussion_r2325013765

textile/objects-features.textile
VeskeR: The check in ably-js you mentioned is only required so that library emits "syncing" -> "synced" events (in that order) for lifecycle events even if `HAS_OBJECTS` flag was false.
The lifecycle events are not part of this spec and will be added in the later PR, which will also address this spec point. Will work on the lifecycle events PR as soon as we finish addressing the outstanding comments on existing PRs.

In relation to this PR, it's fine for the spec to only check `HAS_OBJECTS` flag

textile/objects-features.textile
```diff
@@ -0,0 +1,145 @@
...
+****** @(RTO5c1b1b)@ If @ObjectState.map@ is present, create a zero-value @LiveMap@ (per "RTLM4":#RTLM4), set its private @objectId@ equal to @ObjectState.objectId@, set its private @semantics@ equal to @ObjectState.map.semantics@ and override its internal data using the current @ObjectState@ per "RTLM6":#RTLM6
+****** @(RTO5c1b1c)@ Otherwise, log a warning that an unsupported object state message has been received, and discard the current @ObjectState@ without taking any action
+*** @(RTO5c2)@ Remove any objects from the internal @ObjectsPool@ for which @objectId@s were not received during the sync sequence
+**** @(RTO5c2a)@ The object with ID @root@ must not be removed from @ObjectsPool@, as per "RTO3b":#RTO3b
+*** @(RTO5c3)@ Clear any stored sync sequence identifiers and cursor values
+*** @(RTO5c4)@ The @SyncObjectsPool@ must be cleared
+* @(RTO6)@ When needed, a zero-value object can be created if it does not exist in the internal @ObjectsPool@ for an @objectId@, in the following way:
```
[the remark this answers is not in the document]
VeskeR: I see it was resolved already, but I thought the spec item could be made more clear for the first time reader, thanks for pointing this out. See the change in: https://github.com/ably/specification/pull/333/commits/a70dc1a5634bca7ee219cf51dee499afc25d8914

sacOO7 (approved):
LGTM ( You can recheck all review comments in case )

PR #335, opened 2025-06-13
Add spec for `ObjectMessage` encoding and decoding

VeskeR:


textile/features.textile
```diff
@@ -1518,6 +1518,8 @@ h4. ObjectMessage
 ** @(OM3d)@ The size of the @extras@ property is the string length of its JSON representation
 ** @(OM3f)@ The size of the @clientId@ property is its string length
 ** @(OM3e)@ The size of a @null@ or omitted property is zero
+* @(OM4)@ For @ObjectMessage@ encoding see @ObjectData@ "OD4":#OD4 and @ObjectOperation@ "OOP5":#OOP5 encoding
```
sacOO7, line 1521:
```suggestion
* @(OM4)@ For @ObjectMessage@ encoding see @ObjectData@ "OD4":#OD4 and @ObjectOperation@ encoding see "OOP5":#OOP5
```
Or can simply be made like @ObjectMessage@ decoding - OM5
```suggestion
* @(OM4)@ For @ObjectMessage@ encoding see @ObjectData@ "OD4":#OD4
```

textile/features.textile
```diff
@@ -1530,16 +1532,23 @@ h4. ObjectOperation
 ** @(OOP3d)@ @counterOp@ @CounterOp@ object - the payload for the operation if it is an operation on a @Counter@ object type
 ** @(OOP3e)@ @map@ @Map@ object - the payload for the operation if the operation is @MAP_CREATE@. Defines the initial value for the @Map@ object
 ** @(OOP3f)@ @counter@ @Counter@ object - the payload for the operation if the operation is @COUNTER_CREATE@. Defines the initial value for the @Counter@ object
-** @(OOP3g)@ @nonce@ string - the nonce, must be present on @COUNTER_CREATE@ and @MAP_CREATE@ operations sent to the server
-** @(OOP3h)@ @initialValue@ binary - the initial value bytes for the object
-** @(OOP3i)@ @initialValueEncoding@ string - defines how the @initialValue@ should be interpreted. Should be @msgpack@ or @json@
+** @(OOP3g)@ @nonce@ string - the nonce. Must be present on @COUNTER_CREATE@ and @MAP_CREATE@ operations sent to the server. Should not be accessed by the client library if received from the server
```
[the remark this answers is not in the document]
sacOO7: same goes for @initialValue@ and @initialValueEncoding@

textile/features.textile
```diff
@@ -1530,16 +1532,23 @@ h4. ObjectOperation
...
 ** @(OOP4d)@ The size of the @map@ property is calculated per "MAP4":#MAP4
 ** @(OOP4e)@ The size of the @counter@ property is calculated per "CNT3":#CNT3
 ** @(OOP4f)@ The size of a @null@ or omitted property is zero
+* @(OOP5)@ @ObjectOperation@ encoding:
+** @(OOP5a)@ When the MessagePack protocol is used:
+*** @(OOP5a1)@ A binary @ObjectOperation.initialValue@ is encoded as a MessagePack binary type
+*** @(OOP5a2)@ Set @ObjectOperation.initialValueEncoding@ to @msgpack@
+** @(OOP5b)@ When the JSON protocol is used:
+*** @(OOP5b1)@ A binary @ObjectOperation.initialValue@ is Base64-encoded and represented as a JSON string
+*** @(OOP5b2)@ Set @ObjectOperation.initialValueEncoding@ to @json@
```
sacOO7, line 1552: Don't we have @ObjectOperation@ decoding for any of the fields?

textile/features.textile
```diff
@@ -1530,16 +1532,23 @@ h4. ObjectOperation
...
+** @(OOP3i)@ @initialValueEncoding@ string - defines how the @initialValue@ should be interpreted by the server. Must be @msgpack@ or @json@. Must be present on @COUNTER_CREATE@ and @MAP_CREATE@ operations sent to the server. Should not be accessed by the client library if received from the server
 * @(OOP4)@ The size of the @ObjectOperation@ is calculated as follows:
 ** @(OOP4a)@ The size is the sum of the sizes of the @mapOp@, @counterOp@, @map@, and @counter@ properties
 ** @(OOP4b)@ The size of the @mapOp@ property is calculated per "MOP3":#MOP3
 ** @(OOP4c)@ The size of the @counterOp@ property is calculated per "COP3":#COP3
 ** @(OOP4d)@ The size of the @map@ property is calculated per "MAP4":#MAP4
 ** @(OOP4e)@ The size of the @counter@ property is calculated per "CNT3":#CNT3
 ** @(OOP4f)@ The size of a @null@ or omitted property is zero
+* @(OOP5)@ @ObjectOperation@ encoding:
+** @(OOP5a)@ When the MessagePack protocol is used:
```
sacOO7, line 1546: I have a couple of questions about this:

1. Typically, the client specifies the encoding format (msgpack/json) through HTTP headers or query parameters before the WebSocket handshake. In that case, why do we need a separate initialValueEncoding field? Wouldn't its value always match what was already defined in the header or parameters?

2. Given that custom encoding is supported, is it valid to send initialValue as base64-encoded data while setting initialValueEncoding to json, even if the top-level encoding is msgpack?

sacOO7: The second use case would be beneficial, as we're primarily implementing JSON serializers/deserializers and using utility functions to convert the data into msgpack format.

textile/features.textile
VeskeR: I'm not sure what you're trying to achieve with this change.
ObjectMessage encoding is a bit more complicated than other types of messages, as it doesn't have a single top-level `data` field. Instead different parts of the object message must be encoded and in different ways. This is why there are separate encoding specs for ObjectOperation encoding, and ObjectData encoding.
`OM4` in this case simply instructs you to go and see the encoding spec for those two interfaces

textile/features.textile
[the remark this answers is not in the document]
VeskeR: No, it's straight up should not be used client-side in a case if it for some reason has been sent to the client. Those three fields are for the server and are set by the client library when sending COUNTER_CREATE and MAP_CREATE messages

textile/features.textile
VeskeR: client library should not access the `initialValue` field, so there is not reason to decode it

textile/features.textile
```diff
@@ -1630,6 +1639,28 @@ h4. ObjectData
 ** @(OD3d)@ If set, the size of a @number@ property is 8
 ** @(OD3e)@ If set, the size of a @string@ property is its length
 ** @(OD3f)@ The size of a @null@ or omitted property is zero
+* @(OD4)@ @ObjectData@ encoding:
+** @(OD4a)@ Payloads must be booleans, binary, numbers, strings, or objects capable of JSON representation. Any other data type must not be permitted and result in an error with code 40013.
+** @(OD4b)@ When the MessagePack protocol is used:
+*** @(OD4b1)@ A boolean payload is encoded as a MessagePack boolean type, and the result is set on the @ObjectData.boolean@ attribute.
```
sacOO7, line 1645:
```suggestion
*** @(OD4b1)@ A boolean payload is encoded as a MessagePack boolean type, and the result is set on the @ObjectData.boolean@ attribute. The @ObjectData.encoding@ attribute is then set to "msgpack"
```
or you can add separate spec point for the same

textile/features.textile
VeskeR: > Typically, the client specifies the encoding format (msgpack/json) through HTTP headers or query parameters before the WebSocket handshake. In that case, why do we need a separate initialValueEncoding field? Wouldn't its value always match what was already defined in the header or parameters?

True, but this is the way realtime system operates end expects the client library to send the `initialValue` field.

> Given that custom encoding is supported, is it valid to send initialValue as base64-encoded data while setting initialValueEncoding value to json, even if the top-level encoding header/params is msgpack?

Yes, technically this would work! Not that we're going to do it in SDK implementation to save on headaches from mixing protocols, but it it does work.

textile/features.textile
```diff
@@ -1630,6 +1639,28 @@ h4. ObjectData
...
+*** @(OD4b2)@ A binary payload is encoded as a MessagePack binary type, and the result is set on the @ObjectData.bytes@ attribute.
+*** @(OD4b3)@ A number payload is encoded as a MessagePack number type, and the result is set on the @ObjectData.number@ attribute.
+*** @(OD4b4)@ A string payload is encoded as a MessagePack string type, and the result is set on the @ObjectData.string@ attribute.
+*** @(OD4b5)@ A payload consisting of an object capable of JSON representation is stringified as either a JSON object or array, encoded as a MessagePack string type, and the result is set on the @ObjectData.string@ attribute. The @ObjectData.encoding@ attribute is then set to "json".
+** @(OD4c)@ When the JSON protocol is used:
+*** @(OD4c1)@ A boolean payload is represented as a JSON boolean and set on the @ObjectData.boolean@ attribute.
```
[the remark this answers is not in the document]
VeskeR: see above

textile/features.textile
sacOO7: I just meant to change the spec statement to 
```
* @(OM4)@ For @ObjectMessage@ encoding see @ObjectData@ "OD4":#OD4

File: textile/features.textile (hunk 4 above)
Comment (VeskeR) [reply textile/features.textile:1546]: > The second use case would be beneficial in ably-java, as we're primarily implementing JSON serializers/deserializers and using utility functions to convert the data into msgpack format.

if ably-java supports msgpack protocol connections I'd suggest not to mix encodings and follow Msgpack encoding steps for `initialValue`. Doing the opposite wil also be against the current spec

File: textile/features.textile (hunk 2 above)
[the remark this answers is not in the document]
Comment (sacOO7) [reply textile/features.textile:1535]: This also means, we will never receive values for those from the server.

File: textile/features.textile [hunk 7]
```diff
@@ -1518,6 +1518,8 @@ h4. ObjectMessage
 ** @(OM3d)@ The size of the @extras@ property is the string length of its JSON representation
 ** @(OM3f)@ The size of the @clientId@ property is its string length
 ** @(OM3e)@ The size of a @null@ or omitted property is zero
+* @(OM4)@ For @ObjectMessage@ encoding see @ObjectData@ "OD4":#OD4 and @ObjectOperation@ "OOP5":#OOP5 encoding
+* @(OM5)@ For @ObjectMessage@ decoding see @ObjectData@ "OD5":#OD5 decoding
```
sacOO7, line 1522:
```suggestion
* @(OM5)@ For @ObjectMessage@ decoding see @ObjectData@ "OD5":#OD5 decoding (@ObjectOperation@ doesn't need decoding)
```

sacOO7: This is optional, but makes decoding explicit for @ObjectOperation@

textile/features.textile
sacOO7: Yeah, we have kept both options available. Will be implementing as per spec 👍

textile/features.textile
VeskeR: > Okay, then we should explicitly specify that it should not be set

I don't think it's reasonable to list all the lings the client library _should not_ do. ObjectData encoding/decoding spec should not be compared to encoding for ObjectOperation

textile/features.textile
VeskeR: > and is set to null

This is misleading, as there is nothing you should set to `null` in the context of ObjectOperation.
If it would make things clearear, I can add the spec item for ObjectOperation decoding and just write that there is nothing to decode

VeskeR: > Also, makes statements Should not be accessed by the client library if received from the server obsolete, as they already are as per https://github.com/ably/specification/pull/335/files#r2144719377

Not true. Not having the decoding in place for a field doesn't mean that the client library won't decide to access the field. It just wouldn't understand the contents

textile/features.textile
sacOO7: Of course, it shouldn't be compared directly. However, from a reader's perspective, it's reasonable to expect that the encoding field would have a specific value. It's always better to define things explicitly in the spec to avoid any implicit misunderstandings. In this case, the spec doesn't clearly state whether a value should or shouldn't be set—it's rather ambiguous

textile/features.textile
lmars: I actually don't think it was necessary to add `initialValueEncoding`, I think `initialValue` should always be JSON, there's no need for this optionality, I think we should remove it and document `initialValue` is JSON encoded. I will raise this internally.

sacOO7: Okay, thanks for the more clarity 👍

textile/features.textile
```diff
@@ -1618,18 +1627,40 @@ h4. ObjectData
...
 ** @(OD3c)@ If set, the size of a @bytes@ property is its size in bytes (of the actual binary, not the base64 representation, regardless of whether the binary protocol is in use)
 ** @(OD3d)@ If set, the size of a @number@ property is 8
 ** @(OD3e)@ If set, the size of a @string@ property is its length
 ** @(OD3f)@ The size of a @null@ or omitted property is zero
+* @(OD4)@ @ObjectData@ encoding:
+** @(OD4a)@ Payloads must be booleans, binary, numbers, strings, or objects capable of JSON representation. Any other data type must not be permitted and result in an error with code 40013.
+** @(OD4b)@ When the MessagePack protocol is used:
+*** @(OD4b1)@ A boolean payload is encoded as a MessagePack boolean type, and the result is set on the @ObjectData.boolean@ attribute.
+*** @(OD4b2)@ A binary payload is encoded as a MessagePack binary type, and the result is set on the @ObjectData.bytes@ attribute.
+*** @(OD4b3)@ A number payload is encoded as a MessagePack number type, and the result is set on the @ObjectData.number@ attribute.
```
lawrence-forooghian, line 1647: Are we prescriptive about _which_ MessagePack number type (`Float` or `Integer`)? Probably best that we give implementations the freedom to choose, so that it can be decoded into the most appropriate type on the other end in languages which support such a distinction.

textile/features.textile
lmars: @sacOO7 apologies for not following up, there was a discussion about this on an internal Slack channel, and a decision was made to always use JSON encoding, although it has not been decided who will see the change through (I have asked on the Slack channel).

lmars: > So initialValue will always be base64 encoded string right?

And no, I'm suggesting the `initialValue` NOT be base64 encoded, I'm suggesting it is just JSON encoded.

textile/features.textile
```diff
@@ -1618,18 +1627,40 @@ h4. ObjectData
...
+*** @(OD4b4)@ A string payload is encoded as a MessagePack string type, and the result is set on the @ObjectData.string@ attribute.
+*** @(OD4b5)@ A payload consisting of an object capable of JSON representation is stringified as either a JSON object or array, encoded as a MessagePack string type, and the result is set on the @ObjectData.string@ attribute. The @ObjectData.encoding@ attribute is then set to "json".
+** @(OD4c)@ When the JSON protocol is used:
+*** @(OD4c1)@ A boolean payload is represented as a JSON boolean and set on the @ObjectData.boolean@ attribute.
+*** @(OD4c2)@ A binary payload is Base64-encoded and represented as a JSON string; the result is set on the @ObjectData.bytes@ attribute.
+*** @(OD4c3)@ A number payload is represented as a JSON number and set on the @ObjectData.number@ attribute.
+*** @(OD4c4)@ A string payload is represented as a JSON string and set on the @ObjectData.string@ attribute.
```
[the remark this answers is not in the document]
sacOO7: `ObjectData.encoding` in itself feels misleading, when we do `msgpack` encoding but value can be set to `json` or `none`

textile/features.textile
VeskeR: The current encoding spec for ObjectMessage was written in the same format as the encoding spec for regular messages in [RSL4](https://sdk.ably.com/builds/ably/specification/main/features/#RSL4), where a spec only specifies when _to set the `encoding` field_.
That being said, I don't see a huge issue with adding a single spec point that states that `encoding` field is left unset by default, unless specified by the encoding procedure. Added it in https://github.com/ably/specification/compare/826e4bddf506d3923dc746edc893591ce78d245d..7f30177752945dae89891c9146109fbfa1dcea08

> we describe how the value is encoded—for instance, "A string payload is encoded as a MessagePack string type"—but we neither state that the encoding field must be set, nor that it should be left unset.

I should note that `ObjectData.encoding` (just like `Message.encoding` from [RSL4](https://sdk.ably.com/builds/ably/specification/main/features/#RSL4)) is not meant to tell the client library that the value at the `ObjectData.string` attribute is string-encoded. It exists to inform the client library that the value represents something else and should be further processed. For example, if client received `ObjectData.string` with `ObjectData.encoding` set to `json`, then the string is actually a JSON string - so if you JSON-parse it, you'll get the original object.

textile/features.textile
VeskeR: Realtime treats `ObjectData.number` as a double https://github.com/ably/realtime/blob/80451c37b690d8a5503a32841681d406f71742d9/protocol/state.proto#L18, so I think it's safer for each implementation to explicitly treat it as Float64. I don't think languages with Float and Integer support would like to **not know** what types they will get.
Changed it to `float64` in https://github.com/ably/specification/compare/7f30177752945dae89891c9146109fbfa1dcea08..f4cff454b46de2f450f6641426eb223fe1fc0965

textile/features.textile
```diff
@@ -1618,18 +1627,40 @@ h4. ObjectData
...
+** @(OD2f)@ @string@ string - a primitive string leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - must be set at a time
 * @(OD3)@ The size of the @ObjectData@ is calculated as follows:
 ** @(OD3a)@ The size is the sum of the sizes of the @boolean@, @bytes@, @number@, and @string@ properties
 ** @(OD3b)@ If set, the size of a @boolean@ property is 1
 ** @(OD3c)@ If set, the size of a @bytes@ property is its size in bytes (of the actual binary, not the base64 representation, regardless of whether the binary protocol is in use)
 ** @(OD3d)@ If set, the size of a @number@ property is 8
 ** @(OD3e)@ If set, the size of a @string@ property is its length
 ** @(OD3f)@ The size of a @null@ or omitted property is zero
+* @(OD4)@ @ObjectData@ encoding:
+** @(OD4a)@ Payloads must be booleans, binary, numbers, strings, or objects capable of JSON representation. Any other data type must not be permitted and result in an error with code 40013.
```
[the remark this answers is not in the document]
VeskeR: This spec item uses the same language as existing [RSL4](https://sdk.ably.com/builds/ably/specification/main/features/#RSL4a) for regular messages.
I can see that IDL header in `JSONObject` and `JSONArray` definition also states:
> In the specification text above, values of these types are collectively referred to as “JSON-encodable”.

Would you like me to change the "objects capable of JSON representation" to "JSON-encodable object or array"?

textile/features.textile
sacOO7: Okay, got it 👍

textile/features.textile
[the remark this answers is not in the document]
sacOO7: @lmars, I guess this can be addressed as a part of getting rid of `initialValueEncoding`.

textile/features.textile
```diff
@@ -1618,18 +1627,41 @@ h4. ObjectData
...
-** @(OD2c)@ @boolean@ boolean - a primitive boolean leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - can be set at a time
-** @(OD2d)@ @bytes@ binary - a primitive binary leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - can be set at a time
-** @(OD2e)@ @number@ number - a primitive number leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - can be set at a time
-** @(OD2f)@ @string@ string - a primitive string leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - can be set at a time
+** @(OD2b)@ @encoding@ string - may be set by the client library to indicate that value in @string@ field have an encoding
+** @(OD2c)@ @boolean@ boolean - a primitive boolean leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - must be set at a time
+** @(OD2d)@ @bytes@ binary - a primitive binary leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - must be set at a time
```
sacOO7, line 1632:
```suggestion
** @(OD2d)@ @bytes@ binary | base64 string - a primitive binary leaf value in the object graph. Only one of the value fields - @boolean@, @bytes@, @number@ or @string@ - must be set at a time
```
In case of `Json` encoding, we set `binary` field as base64 encoded `string`.

textile/features.textile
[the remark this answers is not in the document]
VeskeR: Updated phrasing in https://github.com/ably/specification/compare/f4cff454b46de2f450f6641426eb223fe1fc0965..2619ee260ad6f4b039c66fd271d7a81cb8e8d41a

textile/features.textile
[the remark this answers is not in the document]
VeskeR: updated in https://github.com/ably/specification/compare/21dda5224756a9a910ae35c54492c4481aff227a..2e975cb8076dce211733d95e5581ab88a56a4d56

textile/features.textile
lawrence-forooghian: Not sure I understand what problem you're describing here nor what solution you're suggesting? The spec point as currently written (i.e. always encode as a float64) is implementable in Swift. I'm not sure how we'd implement it in JS though — can you hint to the MessagePack encoder which MessagePack type it should use for encoding a JS `number`?

Probably worth getting some input from the Realtime team on how prescriptive we need to be here.

lawrence-forooghian:
> Instead it should use code starting from 0

Why?

lawrence-forooghian:
Sorry, I don't think I know what you're referring to. What property are you talking about exactly, and what is the enum that you're referring to?

textile/features.textile
```diff
@@ -1618,18 +1627,41 @@ h4. ObjectData
...
+** @(OD4a)@ Payloads must be booleans, binary, numbers, strings, or JSON-encodable objects or arrays. Any other data type must not be permitted and result in an error with code 40013
+** @(OD4b)@ @ObjectData.encoding@ must be left unset unless specified otherwise by the payload encoding procedure in "OD4c":#OD4c and "OD4d":#OD4d
+** @(OD4c)@ When the MessagePack protocol is used:
+*** @(OD4c1)@ A boolean payload is encoded as a MessagePack boolean type, and the result is set on the @ObjectData.boolean@ attribute
+*** @(OD4c2)@ A binary payload is encoded as a MessagePack binary type, and the result is set on the @ObjectData.bytes@ attribute
+*** @(OD4c3)@ A number payload is encoded as a MessagePack float64 type, and the result is set on the @ObjectData.number@ attribute
+*** @(OD4c4)@ A string payload is encoded as a MessagePack string type, and the result is set on the @ObjectData.string@ attribute
+*** @(OD4c5)@ A payload consisting of a JSON-encodable object or array is stringified as a JSON object or array, encoded as a MessagePack string type, and the result is set on the @ObjectData.string@ attribute. The @ObjectData.encoding@ attribute is then set to "json"
```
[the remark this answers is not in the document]
sacOO7: e.g. JsonObject with json string size 65kb can be packed to 42-44kb. Since, all SDKs use `msgpack` as the default encoding, it will be helpful.

lawrence-forooghian:
> Sorry, I don't think I know what you're referring to. What property are you talking about exactly, and what is the enum that you're referring to?

Asked about this in standup — @sacOO7 just meant that the spec should be explicit that when sending an enum value (e.g. a `MapSemantics` then the client should encode it as the numerical enum value). I don't think we need to do this here — we've been treating that as implicit elsewhere in the spec (e.g. when sending a `PresenceAction` or `ProtocolMessageAction`). Might not be good to be explicit about it in general but I don't think you need to let it distract you on this PR.

lmars (dismissed):

textile/features.textile
```diff
@@ -1530,16 +1532,23 @@ h4. ObjectOperation
...
-** @(OOP3g)@ @nonce@ string - the nonce, must be present on @COUNTER_CREATE@ and @MAP_CREATE@ operations sent to the server
-** @(OOP3h)@ @initialValue@ binary - the initial value bytes for the object
-** @(OOP3i)@ @initialValueEncoding@ string - defines how the @initialValue@ should be interpreted. Should be @msgpack@ or @json@
+** @(OOP3g)@ @nonce@ string - the nonce. Must be present on @COUNTER_CREATE@ and @MAP_CREATE@ operations sent to the server. Should not be accessed by the client library if received from the server
+** @(OOP3h)@ @initialValue@ binary - the initial value bytes for the object. Must be present on @COUNTER_CREATE@ and @MAP_CREATE@ operations sent to the server. Should not be accessed by the client library if received from the server
+** @(OOP3i)@ @initialValueEncoding@ string - defines how the @initialValue@ should be interpreted by the server. Must be @msgpack@ or @json@. Must be present on @COUNTER_CREATE@ and @MAP_CREATE@ operations sent to the server. Should not be accessed by the client library if received from the server
```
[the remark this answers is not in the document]
sacOO7: Thanks for the update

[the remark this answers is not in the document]
sacOO7: Btw, does this mean, type of `initialValue` will be `JsonObject` ?

textile/features.textile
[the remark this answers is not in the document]
VeskeR: `ObjectData.encoding` and `.json` will be changed in https://github.com/ably/specification/issues/360

textile/features.textile
lawrence-forooghian: I'm happy for you to address this one as you see fit — feel free to resolve and / or create another issue if you want to get this merged

VeskeR: > It sounds like the client SDKs should have freedom to use whichever numerical type they want in the MessagePack encoding, but at the public API level they should only be able to submit a float64

Nicely said, the spec should point out that SDK should stick to float64 representation in the public API and decode realtime values as float64, but maybe it can be more lax when encoding the value to be sent to the realtime to allow for msgpack optimization - it makes sense msgpack to encode small numbers as uint8 for example to save on bandwidth.
Created https://github.com/ably/specification/issues/373 for this to address the wording later.

PR #341, opened 2025-07-04
Add Objects Access API spec

VeskeR:
Adds spec for `LiveMap.size`, `LiveMap.entries`, `LiveMap.keys` and `LiveMap.values`

### DIFF
--- a/textile/objects-features.textile
+++ b/textile/objects-features.textile
@@ -106,6 +106,27 @@ h3(#livemap). LiveMap
 ***** @(RTLM5d2f1)@ If an object with id @objectId@ does not exist, return undefined/null
 ***** @(RTLM5d2f2)@ If an object with id @objectId@ exists, return it
 **** @(RTLM5d2g)@ Otherwise, return undefined/null
+* @(RTLM10)@ @LiveMap#size@:
+** @(RTLM10a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a property similar to @Map.size@ for the native @Map@ class
+** @(RTLM10b)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTLM10c)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 90001
+** @(RTLM10d)@ Returns the number of non-tombstoned entries (per "RTLM14":#RTLM14) in the internal @data@ map
+* @(RTLM11)@ @LiveMap#entries@:
+** @(RTLM11a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.entries()@ for the native @Map@ class
+** @(RTLM11b)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTLM11c)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 90001
+** @(RTLM11d)@ Returns key-value pairs from the internal @data@ map:
+*** @(RTLM11d1)@ Pairs with tombstoned entries (per "RTLM14":#RTLM14) are not returned
+*** @(RTLM11d2)@ The return type is idiomatic for the platform's analogous Map/Dictionary interface operation. For example, in JavaScript, it returns a map iterator object like the one returned by @Map.entries()@ method for the native @Map@ class
+* @(RTLM12)@ @LiveMap#keys@:
+** @(RTLM12a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.keys()@ for the native @Map@ class
+** @(RTLM12b)@ Identical to @LiveMap#entries@, except that it returns only the keys from the internal @data@ map
+* @(RTLM13)@ @LiveMap#values@:
+** @(RTLM13a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.values()@ for the native @Map@ class
+** @(RTLM13b)@ Identical to @LiveMap#entries@, except that it returns only the values from the internal @data@ map
+* @(RTLM14)@ An @ObjectsMapEntry@ in the internal @data@ map can be checked for being tombstoned using the convenience method:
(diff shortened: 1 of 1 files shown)


textile/objects-features.textile
```diff
@@ -106,6 +106,27 @@ h3(#livemap). LiveMap
...
+* @(RTLM10)@ @LiveMap#size@:
+** @(RTLM10a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a property similar to @Map.size@ for the native @Map@ class
+** @(RTLM10b)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTLM10c)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 90001
+** @(RTLM10d)@ Returns the number of non-tombstoned entries (per "RTLM14":#RTLM14) in the internal @data@ map
+* @(RTLM11)@ @LiveMap#entries@:
+** @(RTLM11a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.entries()@ for the native @Map@ class
+** @(RTLM11b)@ Requires the @OBJECT_SUBSCRIBE@ channel mode to be granted per "RTO2":#RTO2
+** @(RTLM11c)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 90001
+** @(RTLM11d)@ Returns key-value pairs from the internal @data@ map:
```
lawrence-forooghian, line 118: Given that this presumably doesn't return `ObjectsMapEntry` values, this presumably needs to apply the same mapping as RTLM5d2?

lawrence-forooghian: And then also, how does `entries` handle the following cases?

- referenced object doesn't exist (RTLM5d2f1)
- entry contains no data we know how to handle (RTLM5d2g)

— can `entries` represent an "undefined/null" value in its output, or does it drop such entries?

VeskeR: created a separate ticket https://github.com/ably/specification/issues/372 to discuss how we want to represent fields with **"there is an entry but we don't know how to represent it"** - suggested sentinel pattern seems interesting, should discuss if would make sense across all SDKs

textile/objects-features.textile
```diff
@@ -111,6 +111,29 @@ h3(#livemap). LiveMap
...
+* @(RTLM12)@ @LiveMap#keys@:
+** @(RTLM12a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.keys()@ for the native @Map@ class
+** @(RTLM12b)@ Identical to @LiveMap#entries@, except that it returns only the keys from the internal @data@ map
+* @(RTLM13)@ @LiveMap#values@:
+** @(RTLM13a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.values()@ for the native @Map@ class
+** @(RTLM13b)@ Identical to @LiveMap#entries@, except that it returns only the values from the internal @data@ map
+* @(RTLM14)@ An @ObjectsMapEntry@ in the internal @data@ map can be checked for being tombstoned using the convenience method:
+** @(RTLM14a)@ The method returns true if @ObjectsMapEntry.tombstone@ is true
```
sacOO7, line 135:
```suggestion
** @(RTLM14a)@ The method returns true if @ObjectsMapEntry.tombstone@ is true or reference to `LiveMap` or `LiveCounter` as per "RTLM5d2f ":#RTLM5d2f is tombstoned. 
```

sacOO7: We also need to update `RTLM5d2f` to return `null` if given `object.isTombstoned` is set to `true`. 
That's how it's written in ably-js.

textile/objects-features.textile
```diff
@@ -111,6 +111,29 @@ h3(#livemap). LiveMap
...
+** @(RTLM11d)@ Returns key-value pairs from the internal @data@ map:
+*** @(RTLM11d1)@ Pairs with tombstoned entries (per "RTLM14":#RTLM14) are not returned
+*** @(RTLM11d3)@ @ObjectsMapEntry@ values are mapped to user-facing values following the same procedure as in "RTLM5d2":#RTLM5d2
+**** @(RTLM11d3a)@ Note that if "RTLM5d2":#RTLM5d2 results in an @ObjectsMapEntry@ being mapped to an undefined/null value, the corresponding key-value pair is still returned by this @LiveMap#entries@ call
+*** @(RTLM11d2)@ The return type is idiomatic for the platform's analogous Map/Dictionary interface operation. For example, in JavaScript, it returns a map iterator object like the one returned by @Map.entries()@ method for the native @Map@ class
+* @(RTLM12)@ @LiveMap#keys@:
+** @(RTLM12a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.keys()@ for the native @Map@ class
+** @(RTLM12b)@ Identical to @LiveMap#entries@, except that it returns only the keys from the internal @data@ map
```
sacOO7, line 130:
```suggestion
** @(RTLM12b)@ Implementation is identical to @LiveMap#entries@, except that it returns only the keys from the internal @data@ map
```

sacOO7 (approved):
Except from the minor comments ( makes spec explicit and more clear to understand )
LGTM

textile/objects-features.textile
```diff
@@ -111,6 +111,29 @@ h3(#livemap). LiveMap
...
+* @(RTLM13)@ @LiveMap#values@:
+** @(RTLM13a)@ A method or property, depending on what is more idiomatic for the platform to use for a Map/Dictionary interface. For example, in JavaScript, this is a method similar to @Map.values()@ for the native @Map@ class
+** @(RTLM13b)@ Identical to @LiveMap#entries@, except that it returns only the values from the internal @data@ map
```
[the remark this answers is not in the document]
VeskeR: updated in https://github.com/ably/specification/pull/341/commits/0bff2667b13d5265974071b10059d668a0e65d34,
didn't enclose "values" in @@ as we usually do this only for explicit variable/property/class names and "values" here is used just as a generic word (just like `RTLM11d` has just "key-value pairs")

textile/objects-features.textile
VeskeR: updated in https://github.com/ably/specification/pull/341/commits/0bff2667b13d5265974071b10059d668a0e65d34,
didn't enclose "keys" in @@ as we usually do this only for explicit variable/property/class names and "keys" here is used just as a generic word (just like `RTLM11d` has just "key-value pairs")

PR #346, opened 2025-07-11
[PUB-1827] Add Objects spec for subscriptions

VeskeR:
Resolves [PUB-1827](https://ably.atlassian.net/browse/PUB-1827)

[PUB-1827]: https://ably.atlassian.net/browse/PUB-1827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/textile/objects-features.textile
+++ b/textile/objects-features.textile
@@ -36,6 +36,7 @@ h3(#realtime-objects). RealtimeObjects
 ** @(RTO4b)@ If the @HAS_OBJECTS@ flag is 0 or there is no @flags@ field, the sync sequence must be considered complete immediately, and the client library must perform the following actions in order:
 *** @(RTO4b1)@ All objects except the one with id @root@ must be removed from the internal @ObjectsPool@
 *** @(RTO4b2)@ The data for the @LiveMap@ with id @root@ must be cleared by setting it to a zero-value per "RTLM4":#RTLM4. Note that the client SDK must not create a new @LiveMap@ instance with id @root@; it must only clear the internal data of the existing @LiveMap@ with id @root@
+**** @(RTO4b2a)@ Emit a @LiveMapUpdate@ object for the @LiveMap@ with ID @root@, with @LiveMapUpdate.update@ consisting of entries for the keys that were removed, each set to @removed@
 *** @(RTO4b3)@ The @SyncObjectsPool@ list must be cleared
 *** @(RTO4b5)@ The @BufferedObjectOperations@ list must be cleared
 *** @(RTO4b4)@ Perform the actions for objects sync completion as described in "RTO5c":#RTO5c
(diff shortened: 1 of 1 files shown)


textile/objects-features.textile
```diff
@@ -52,13 +53,15 @@ h3(#realtime-objects). RealtimeObjects
 *** @(RTO5c1)@ For each @ObjectState@ member in the @SyncObjectsPool@ list:
 **** @(RTO5c1a)@ If an object with @ObjectState.objectId@ exists in the internal @ObjectsPool@:
 ***** @(RTO5c1a1)@ Override the internal data for the object as per "RTLC6":#RTLC6, "RTLM6":#RTLM6
+***** @(RTO5c1a2)@ Store the @LiveObjectUpdate@ object returned by the operation, along with a reference to the updated object
```
lawrence-forooghian, line 56: RTLC6 and RTLM6 don't currently define a "return value" — sounds like they should?

textile/objects-features.textile
```diff
@@ -265,14 +318,30 @@ class LiveObject: // RTLO*
   siteTimeserials: Dict<String, String> // RTLO3b, internal
   createOperationIsMerged: Boolean // RTLO3c, internal
   canApplyOperation(ObjectMessage) -> Boolean // RTLO4a, internal
+  subscribe((LiveObjectUpdate) ->) -> LiveObjectSubscription // RTLO4b
+  unsubscribe((LiveObjectUpdate) ->) // RTLO4c
+  unsubscribeAll() // RTLO4d
+
+interface LiveObjectSubscription: // RTLO4b5
```
sacOO7, line 325:
```suggestion
interface ObjectsSubscription: // RTLO4b5
```
As per new naming conventions

textile/objects-features.textile
```diff
@@ -265,14 +318,30 @@ class LiveObject: // RTLO*
   siteTimeserials: Dict<String, String> // RTLO3b, internal
   createOperationIsMerged: Boolean // RTLO3c, internal
   canApplyOperation(ObjectMessage) -> Boolean // RTLO4a, internal
+  subscribe((LiveObjectUpdate) ->) -> LiveObjectSubscription // RTLO4b
```
sacOO7, line 321: Since LiveObject is an internal type, exposing subscribe methods on it doesn't make much sense. While this approach works well in JavaScript, in Java, classes implementing the LiveMap and LiveCounter interfaces should implement those methods.

textile/objects-features.textile
```diff
@@ -265,14 +318,30 @@ class LiveObject: // RTLO*
...
+
+interface LiveObjectUpdate: // RTLO4b4
+  update: Object // RTLO4b4a
+  noop: Boolean // RTLO4b4b, internal
 
 class LiveCounter extends LiveObject: // RTLC*, RTLC1
   value() -> Number // RTLC5
 
+interface LiveCounterUpdate extends LiveObjectUpdate: // RTLC11, RTLC11a
+  update: { amount: Number } // RTLC11b, RTLC11b1
```
sacOO7, line 336: Why can't it simply be 
```suggestion
  update: Number  // RTLC11b, RTLC11b1
```
This will make it same as `LiveMapUpdate`, where `update` is at top level instead of nesting.

sacOO7: In Java or similar languages, we often have to create redundant classes with accessor methods just to retrieve a single value. Having two levels of accessors to access one piece of data feels inefficient. Also, for the `LiveCounter`, it seems unlikely that any further updates will be needed.

textile/objects-features.textile
VeskeR: Will clarify the returned update object in a separate PR as it will add quite a few spec points: https://github.com/ably/specification/pull/363 (WIP draft)

textile/objects-features.textile
VeskeR: > Athough, we have a LiveObject class, it's internal and we chose not to use this naming convention right?

The `LiveObject`  interface is exposed in ably-js implementation and the naming convention is correct in this case. This is a "live object" - an object whose data is synced in realtime - and its name only coincidentally resembles the name of the **LiveObjects** feature.
The `LiveObjectSubscription` in this case represents a subscription for a `LiveObject` instance, so I think the name should stay as it is.

> I think we can have a common interface that implements unsubscribe for all type of subscribe or on methods

The current API uses the subscribe/unsubscribe and on/off naming pairs, so for `.on` methods the returned object should expose `.off()` method I think.

textile/objects-features.textile
VeskeR: The `LiveObject` interface is exposed in the ably-js implementation. However, if you feel that in the Java implementation it’s better and more convenient to keep it as an internal type, you can certainly declare the `subscribe` methods directly on the `LiveMap` and `LiveCounter` interfaces.

textile/objects-features.textile
lawrence-forooghian: I'll resolve this one in that case and leave #363 to you

lawrence-forooghian (approved):

PR #351, opened 2025-07-17
chore: update `msgSerial` spec point

ttypic:
Clarify that `msgSerial` should reset to `0` on clean connection attempts, including reconnects after `SUSPENDED`.

There is nothing in the spec about going back from `SUSPENDED`, and we have different implementation at least in js and java. js resets `msgSerial` after `suspended`, java doesn't. `msgSerial` reset is mentioned in resume attempts and manual `connect()` after terminal states.

### DIFF
--- a/textile/features.textile
+++ b/textile/features.textile
@@ -1690,7 +1690,7 @@ h4. ProtocolMessage
 ** @(TR4g)@ @count@ integer
 ** @(TR4h)@ @error@ @ErrorInfo@ object
 ** @(TR4i)@ @flags@ integer. Contains one or more of the bit flags specified in @TR3@
-** @(TR4j)@ @msgSerial@ long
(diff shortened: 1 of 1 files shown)


SimonWoolf (changes requested):
This sort of requirement does not belong in TRxx, which is just listing the fields of a `ProtocolMessage`. Should be in the RTN section.

RTN15g already states:

> `** @(RTN15g)@ Connection state is only maintained server-side for a brief period, given by the @connectionStateTtl@ in the @connectionDetails@, see "CD2f":#CD2f. If a client has been disconnected for longer than the @connectionStateTtl@, it should not attempt to resume. Instead, it should clear the local connection state, and any connection attempts should be made as for a fresh connection`

"clear the local connection state" technically does already imply setting the msgSerial to 0. But it's implicit; worth rewording that item to make it more explicit exactly what should be cleared (connectionId, connectionKey, msgSerial) if people are confused.

lawrence-forooghian:
I was doing something similar in #351 just now but realised that it's basically the same as this — I'm happy to take this one over if you want @ttypic?

PR #352, opened 2025-07-20
chat: add disposal spec

AndyTWF:
Adds a specification for disposing of the Chat Client and its rooms instance.

[CHA-1054]

[CHA-1054]: https://ably.atlassian.net/browse/CHA-1054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -302,11 +302,11 @@ h2(#roos-disposal). Rooms Disposal
 
 Rooms in the Chat SDK are a complex type that register many different listeners and consume many different events. Therefore, there must be methods available to clean up any listeners to allow for garbage collection.
 
-** @(CHA-RD1)@ @[Testable]@ The @Rooms@ map / manager must offer an internal @dispose@ method that must prepare the rooms manager for garbage collection and release all managed rooms.
-*** @(CHA-RD1a)@ @[Testable]@ When @dispose@ is called, it must mark the instance as disposed to prevent future room operations.
+** @(CHA-RD1)@ @[Testable]@ The @Rooms@ map / manager must offer an internal API for disposal, that must prepare the rooms manager for garbage collection and release all managed rooms.
+*** @(CHA-RD1a)@ @[Testable]@ When disposal is called, it must mark the instance as disposed to prevent future room operations.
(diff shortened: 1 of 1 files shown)


textile/chat-features.textile
```diff
@@ -289,6 +298,16 @@ Each chat room can be configured individually, allowing options to be passed as
...
+h2(#roos-disposal). Rooms Disposal
+
+Rooms in the Chat SDK are a complex type that register many different listeners and consume many different events. Therefore, there must be methods available to clean up any listeners to allow for garbage collection.
+
+** @(CHA-RD1)@ @[Testable]@ The @Rooms@ map / manager must offer an internal @dispose@ method that must prepare the rooms manager for garbage collection and release all managed rooms.
+*** @(CHA-RD1a)@ @[Testable]@ When @dispose@ is called, it must mark the instance as disposed to prevent future room operations.
+*** @(CHA-RD1b)@ @[Testable]@ If no rooms exist when @dispose@ is called, the method must complete successfully without error.
+*** @(CHA-RD1c)@ @[Testable]@ If rooms exist in the map, the @dispose@ method must release all rooms currently in the rooms map concurrently.
+*** @(CHA-RD1d)@ @[Testable]@ The @dispose@ method must wait for all room release operations to complete before resolving.
+
```
vladvelici, line 310: I thought the point of the spec generally is public API. These points describe how things work internally. Do we want these in the spec?

vladvelici (commented):
There's quite a bit of internal API in this. Shouldn't the spec mostly focus on the public API and behaviour we expect from the SDK when interacting with this public API?

textile/chat-features.textile
AndyTWF: Whilst we've tried to focus it on the public API where possible, there are occasions with complex internal behaviour that needs specifying to avoid implementation divergence (e.g. RTL3d specifies performing an internal attach operation where necessary). We did the same with the room lifecycle manager.

vladvelici (approved):

PR #353, opened 2025-07-23
Add object-level write API spec for RealtimeObjects

VeskeR:
Adds spec for:
- `RealtimeObjects.createMap`
- `RealtimeObjects.createCounter`

Resolves PUB-1829


textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
...
+***** @(RTO11f4c2)@ Add a new entry to @ObjectOperation.map.entries@ with the current key and the created @ObjectsMapEntry@ as the value
+*** @(RTO11f5)@ Create an initial value JSON string as described in "RTO13":#RTO13, passing in the partial @ObjectOperation@ from "RTO11f4":#RTO11f4
+*** @(RTO11f6)@ Create a unique nonce as a random string
+*** @(RTO11f7)@ Get the current server time as described in "RTO16":#RTO16
+*** @(RTO11f8)@ Create an @objectId@ for the new @LiveMap@ object as described in "RTO14":#RTO14, passing in @map@ string as the @type@, the initial value JSON string from "RTO11f5":#RTO11f5, the nonce from "RTO11f6":#RTO11f6, and the server time from "RTO11f7":#RTO11f7
+*** @(RTO11f9)@ Set @ObjectMessage.operation.action@ to @ObjectOperationAction.MAP_CREATE@
+*** @(RTO11f10)@ Set @ObjectMessage.operation.objectId@ to the @objectId@ created in "RTO11f8":#RTO11f8
+*** @(RTO11f11)@ Set @ObjectMessage.operation.nonce@ to the nonce value created in "RTO11f6":#RTO11f6
+*** @(RTO11f12)@ Set @ObjectMessage.operation.initialValue@ to the JSON string created in "RTO11f5":#RTO11f5
```
lawrence-forooghian, line 53: I didn't realise until now that this PR doesn't have #335 as an ancestor, but I think that somewhere we need to update OOP5 to remove `initialValueEncoding` and make `initialValue` a string?

textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
...
+** @(RTO12e)@ If "@echoMessages@":../features#TO3h client option is @false@, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 40000, indicating that @echoMessages@ must be enabled for this operation
+** @(RTO12f)@ Creates an @ObjectMessage@ for a @COUNTER_CREATE@ action in the following way:
+*** @(RTO12f1)@ If @count@ is null, not of type @Number@, or not a finite number, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 40003, indicating that @count@ must be a valid number. Note that @count@ is an optional argument, and if omitted, this error must not be thrown
+*** @(RTO12f2)@ Create a partial @ObjectOperation@ with the initial value for the new @LiveCounter@:
+**** @(RTO12f2a)@ Set @ObjectOperation.counter.count@ to 0 if @count@ is omitted
+**** @(RTO12f2b)@ Otherwise, set @ObjectOperation.counter.count@ to the provided @count@ value
+*** @(RTO12f3)@ Create an initial value JSON string as described in "RTO13":#RTO13, passing in the partial @ObjectOperation@ from "RTO12f2":#RTO12f2
+*** @(RTO12f4)@ Create a unique nonce as a random string
```
lawrence-forooghian, line 77: Any guidance on this? e.g. would a UUID suffice?

textile/objects-features.textile
```diff
@@ -396,6 +548,8 @@ interface LiveObjectUpdate: // RTLO4b4
 
 class LiveCounter extends LiveObject: // RTLC*, RTLC1
   value() -> Number // RTLC5
+  increment(Number amount) => io // RTLC12
```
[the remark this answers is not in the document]
sacOO7: Same can be updated for `value` and `decrement` methods.

textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
...
 ** @(RTO1c)@ Waits for the objects sync sequence to complete and for "RTO5c":#RTO5c to finish
 ** @(RTO1d)@ Returns the object with id @root@ from the internal @ObjectsPool@ as a @LiveMap@
+* @(RTO11)@ @RealtimeObjects#createMap@ function:
+** @(RTO11a)@ Expects the following arguments:
+*** @(RTO11a1)@ @entries@ @Dict<String, Boolean | Binary | Number | String | JsonArray | JsonObject | LiveCounter | LiveMap>@ (optional) - the initial entries for the new @LiveMap@ object
+** @(RTO11b)@ The return type is a @LiveMap@, which is returned once the required I/O has successfully completed
+** @(RTO11c)@ Requires the @OBJECT_PUBLISH@ channel mode to be granted per "RTO2":#RTO2
+** @(RTO11d)@ If the channel is in the @DETACHED@, @FAILED@ or @SUSPENDED@ state, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 90001
+** @(RTO11e)@ If "@echoMessages@":../features#TO3h client option is @false@, the library should throw an @ErrorInfo@ error with @statusCode@ 400 and @code@ 40000, indicating that @echoMessages@ must be enabled for this operation
```
sacOO7, line 29: Since the connection can be shared across PubSub, Chat, and Objects, this check might limit the ability to do so. We need to assess the potential impact across various product offerings

textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
...
+****** @(RTO11f4c1c)@ If the value is of type @String@, set @ObjectsMapEntry.data.string@ to that value
+****** @(RTO11f4c1d)@ If the value is of type @Number@, set @ObjectsMapEntry.data.number@ to that value
+****** @(RTO11f4c1e)@ If the value is of type @Boolean@, set @ObjectsMapEntry.data.boolean@ to that value
+****** @(RTO11f4c1f)@ If the value is of type @Binary@, set @ObjectsMapEntry.data.bytes@ to that value
+***** @(RTO11f4c2)@ Add a new entry to @ObjectOperation.map.entries@ with the current key and the created @ObjectsMapEntry@ as the value
+*** @(RTO11f5)@ Create an initial value JSON string as described in "RTO13":#RTO13, passing in the partial @ObjectOperation@ from "RTO11f4":#RTO11f4
+*** @(RTO11f6)@ Create a unique nonce as a random string
+*** @(RTO11f7)@ Get the current server time as described in "RTO16":#RTO16
+*** @(RTO11f8)@ Create an @objectId@ for the new @LiveMap@ object as described in "RTO14":#RTO14, passing in @map@ string as the @type@, the initial value JSON string from "RTO11f5":#RTO11f5, the nonce from "RTO11f6":#RTO11f6, and the server time from "RTO11f7":#RTO11f7
```
sacOO7, line 49: In the case of REST APIs, `objectId` is optional and is automatically generated by the server.
Applying the same approach for realtime could reduce round-trip time—something particularly valuable when customers are using RealtimeObjects.
So, why don’t we implement it this way for realtime as well?

textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
...
+*** @(RTO11f9)@ Set @ObjectMessage.operation.action@ to @ObjectOperationAction.MAP_CREATE@
+*** @(RTO11f10)@ Set @ObjectMessage.operation.objectId@ to the @objectId@ created in "RTO11f8":#RTO11f8
+*** @(RTO11f11)@ Set @ObjectMessage.operation.nonce@ to the nonce value created in "RTO11f6":#RTO11f6
+*** @(RTO11f12)@ Set @ObjectMessage.operation.initialValue@ to the JSON string created in "RTO11f5":#RTO11f5
+*** @(RTO11f13)@ Merge values from the partial @ObjectOperation@ created in "RTO11f4":#RTO11f4 into @ObjectMessage.operation@
```
sacOO7, line 54: Is this necessary? My integration test passes without merging those values—tbe, without setting the `operation.map` field. I believe our `initialValueJSONString` already contains the required data, so setting map might be redundant. What do you think?

textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
...
+**** @(RTO12f2b)@ Otherwise, set @ObjectOperation.counter.count@ to the provided @count@ value
+*** @(RTO12f3)@ Create an initial value JSON string as described in "RTO13":#RTO13, passing in the partial @ObjectOperation@ from "RTO12f2":#RTO12f2
+*** @(RTO12f4)@ Create a unique nonce as a random string
+*** @(RTO12f5)@ Get the current server time as described in "RTO16":#RTO16
+*** @(RTO12f6)@ Create an @objectId@ for the new @LiveCounter@ object as described in "RTO14":#RTO14, passing in @counter@ string as the @type@, the initial value JSON string from "RTO12f3":#RTO12f3, the nonce from "RTO12f4":#RTO12f4, and the server time from "RTO12f5":#RTO12f5
+*** @(RTO12f7)@ Set @ObjectMessage.operation.action@ to @ObjectOperationAction.COUNTER_CREATE@
+*** @(RTO12f8)@ Set @ObjectMessage.operation.objectId@ to the @objectId@ created in "RTO12f6":#RTO12f6
+*** @(RTO12f9)@ Set @ObjectMessage.operation.nonce@ to the nonce value created in "RTO12f4":#RTO12f4
+*** @(RTO12f10)@ Set @ObjectMessage.operation.initialValue@ to the JSON string created in "RTO12f3":#RTO12f3
+*** @(RTO12f11)@ Merge values from the partial @ObjectOperation@ created in "RTO12f2":#RTO12f2 into @ObjectMessage.operation@
```
sacOO7, line 84: Same as https://github.com/ably/specification/pull/353/files#r2244829677.
`CounterCreate` works fine without merging values.

textile/objects-features.textile
```diff
@@ -374,6 +523,9 @@ Types and their properties/methods are public and exposed to users by default. A
 <pre>
 class RealtimeObjects: // RTO*
   getRoot() => io LiveMap // RTO1
+  createMap(Dict<String, Boolean | Binary | Number | String | JsonArray | JsonObject | LiveCounter | LiveMap> entries?) => io LiveMap // RTO11
```
sacOO7, line 526:
```suggestion
  createMap(Dict<String, Boolean | Binary | Number | JsonArray | JsonObject | LiveCounter | LiveMap> entries?) => io LiveMap // RTO11
```

textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
...
+****** @(RTO11f4c1e)@ If the value is of type @Boolean@, set @ObjectsMapEntry.data.boolean@ to that value
+****** @(RTO11f4c1f)@ If the value is of type @Binary@, set @ObjectsMapEntry.data.bytes@ to that value
+***** @(RTO11f4c2)@ Add a new entry to @ObjectOperation.map.entries@ with the current key and the created @ObjectsMapEntry@ as the value
+*** @(RTO11f5)@ Create an initial value JSON string as described in "RTO13":#RTO13, passing in the partial @ObjectOperation@ from "RTO11f4":#RTO11f4
+*** @(RTO11f6)@ Create a unique nonce as a random string
+*** @(RTO11f7)@ Get the current server time as described in "RTO16":#RTO16
+*** @(RTO11f8)@ Create an @objectId@ for the new @LiveMap@ object as described in "RTO14":#RTO14, passing in @map@ string as the @type@, the initial value JSON string from "RTO11f5":#RTO11f5, the nonce from "RTO11f6":#RTO11f6, and the server time from "RTO11f7":#RTO11f7
+*** @(RTO11f9)@ Set @ObjectMessage.operation.action@ to @ObjectOperationAction.MAP_CREATE@
+*** @(RTO11f10)@ Set @ObjectMessage.operation.objectId@ to the @objectId@ created in "RTO11f8":#RTO11f8
```
sacOO7, line 51: We can make `RTO11f10`, `RTO11f11` as subpoints of `RTO11f9`, since we are constructing `ObjectMessage` for `MapCreateOperation` right

textile/objects-features.textile
VeskeR: `initialValue` and `initialValueEncoding` will be updated in https://github.com/ably/specification/issues/359 to not retroactively change the spec points in https://github.com/ably/specification/pull/335

textile/objects-features.textile
VeskeR: Any reasonably random string to guarantee that two clients won't end up generating the same value at the same time will suffice, yes.
UUID is good, I think I'll use the `16+ character nonce` rule from https://sdk.ably.com/builds/ably/specification/main/features/#RSA9c and update `RTO12f4`

textile/objects-features.textile
VeskeR: I've opened a separate issue for this https://github.com/ably/specification/issues/365, and this can be discussed in future LiveObjects plugin discussions.
As of now, LiveObjects does not support applying operations locally first and instead relies on receiving echoed operations.

textile/objects-features.textile
VeskeR: In Realtime, developers can obtain a reference to an object **before** the corresponding `*_CREATE` operation has been processed by the server and an ID has been generated for it. In such cases, it is crucial that the Realtime client is still able to send mutation operations for that object or reference it within other collection types. Without knowing the object ID, this becomes impossible and would require implementing a queuing mechanism, which would quickly become complex and difficult to manage.

The need for client-generated object IDs is further emphasized by the new path-based API and the proposed object creation changes [here](https://ably.atlassian.net/wiki/spaces/LOB/pages/4235722804/LODR-042+WIP+Realtime+Client+API+Improvements#Creating-objects). Object creation is no longer an async method but instead a simple call like `LiveCounter.create()`, which returns a value type describing the structure of the object.

Furthermore, the end goal of the LiveObjects plugin is to provide a local-first, offline-first API, where changes can be made entirely offline on the client side and later synced via Realtime across all other clients. This is only possible if the client is able to generate object IDs.

For the REST API, all of the above is not relevant, so we can rely on the server to generate object IDs.

textile/objects-features.textile
VeskeR: Interesting, it might be the relic from before we started using `initialValue` for object creations. I've created https://github.com/ably/specification/issues/366 for this as I don't want to spend time right now figuring out if realtime does indeed ignore `operation.map`/`operation.counter` values for create ops.

textile/objects-features.textile
VeskeR: Created https://github.com/ably/specification/issues/366 for this, see https://github.com/ably/specification/pull/353#discussion_r2319486381 comment

textile/objects-features.textile
```diff
@@ -20,6 +20,77 @@ h3(#realtime-objects). RealtimeObjects
 ** @(RTO1b)@ If the channel is in the @DETACHED@ or @FAILED@ state, the library should indicate an error with code 90001
 ** @(RTO1c)@ Waits for the objects sync sequence to complete and for "RTO5c":#RTO5c to finish
 ** @(RTO1d)@ Returns the object with id @root@ from the internal @ObjectsPool@ as a @LiveMap@
+* @(RTO11)@ @RealtimeObjects#createMap@ function:
+** @(RTO11a)@ Expects the following arguments:
+*** @(RTO11a1)@ @entries@ @Dict<String, Boolean | Binary | Number | String | JsonArray | JsonObject | LiveCounter | LiveMap>@ (optional) - the initial entries for the new @LiveMap@ object
```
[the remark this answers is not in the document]
VeskeR: No, the current spec is correct:
`entries` is a dictionary `Dict`, with `String` keys: `@Dict<String, VALUE_TYPE>`, and the value type can be one of `Boolean | Binary | Number | String | JsonArray | JsonObject | LiveCounter | LiveMap` - which must also include `String`

textile/objects-features.textile
VeskeR: current spec is correct, see similar comment https://github.com/ably/specification/pull/353#discussion_r2319491318

textile/objects-features.textile
VeskeR: I think the current structure is correct. The parent `RTO11f` specifies that an `ObjectMessage` for a `MAP_CREATE` action can be created by following its sub-items. You then simply follow these sub-items in sequence to construct the correct `ObjectMessage`.
It makes sense for `RTO11f10` and the following items to be at the same level as `RTO11f9` and the other `RTO11f*` entries, as they all describe the steps required to create the `ObjectMessage`.

sacOO7 (approved):
lgtm

PR #355, opened 2025-08-04
Fix nullability of `ObjectsMapEntry.data`

lawrence-forooghian:
I have observed it being absent for a tombstoned map entry received in an `OBJECT_SYNC`. Also RTLM8a2a of #333 at dd25dca tells us to set it to null.


VeskeR (approved):
LGTM

PR #357, opened 2025-08-14
Chat/move to pv4

splindsay-92:
## Context

- Updated the chat message protocol from V3 to V4, including REST API endpoints (sending, updating, deleting, history).
- Added a new structure for message objects in V4, detailing additional fields like version and reactions.
- Added new response formats for sending, updating, deleting, and history in REST API V4.
- We now also don't nest messages in the response from API calls for send/delete/update.


vladvelici (approved):
👍

PR #363, opened 2025-09-03
[WIP] Clarify the update object returned after replacing the internal data for LiveMap or LiveCounter in RTLC6 and RTLM6

VeskeR:


sacOO7:
Current PR is outdated, we already have explicit spec point for `RTLC14` and `RTLM22`.

PR #370, opened 2025-09-04
[WIP] Rename `SyncObjectsPool` to `SyncObjectMessages` and clarify its type

VeskeR:
The name `SyncObjectsPool` incorrectly suggests it has something to do with the `ObjectsPool` used to store actual live objects, while it's only a temporary array of object sync messages received during sync sequence.

The current [ably-js implementation](https://github.com/ably/ably-js/blob/af2ce0eef23068b19cccd2677453386cc470372d/src/plugins/objects/syncobjectsdatapool.ts#L25-L28) is also too complicated as there is no need anymore to store object sync messages in any kind of map of `objectId -> objectMessage`, and it can be replaced via simple list like the current spec suggests.

Also for spec items points `RTLC6f` and `RTLM6f` which refer to `the outer @ObjectMessage@ for the @ObjectState@` it would make more sense for the `SyncObjectsPool` to directly hold a list of `ObjectMessage` messages, not the underlying `ObjectState`, hence the renaming to `SyncObjectMessages`

### DIFF
--- a/textile/objects-features.textile
+++ b/textile/objects-features.textile
@@ -108,7 +108,7 @@ h3(#realtime-objects). RealtimeObjects
 *** @(RTO4b1)@ All objects except the one with id @root@ must be removed from the internal @ObjectsPool@
 *** @(RTO4b2)@ The data for the @LiveMap@ with id @root@ must be cleared by setting it to a zero-value per "RTLM4":#RTLM4. Note that the client SDK must not create a new @LiveMap@ instance with id @root@; it must only clear the internal data of the existing @LiveMap@ with id @root@
 **** @(RTO4b2a)@ Emit a @LiveMapUpdate@ object for the @LiveMap@ with ID @root@, with @LiveMapUpdate.update@ consisting of entries for the keys that were removed, each set to @removed@
-*** @(RTO4b3)@ The @SyncObjectsPool@ list must be cleared
+*** @(RTO4b3)@ The @SyncObjectMessages@ list must be cleared
 *** @(RTO4b5)@ The @BufferedObjectOperations@ list must be cleared
 *** @(RTO4b4)@ Perform the actions for objects sync completion as described in "RTO5c":#RTO5c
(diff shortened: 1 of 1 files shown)


textile/objects-features.textile
```diff
@@ -107,21 +107,22 @@ h3(#realtime-objects). RealtimeObjects
...
-**** @(RTO5a2a)@ The @SyncObjectsPool@ list must be cleared
+**** @(RTO5a2a)@ The @SyncObjectMessages@ list must be cleared
 **** @(RTO5a2b)@ The @BufferedObjectOperations@ list must be cleared
 *** @(RTO5a3)@ If the sequence id matches the previously received sequence id, the client library should continue the sync process
 *** @(RTO5a4)@ The objects sync sequence for that sequence identifier is considered complete once the cursor is empty; that is when the @channelSerial@ looks like @<sequence id>:@
 *** @(RTO5a5)@ An @OBJECT_SYNC@ may also be sent with no @channelSerial@ attribute. In this case, the sync data is entirely contained within the @ProtocolMessage@
-** @(RTO5b)@ During the sync sequence, the @ObjectMessage.object@ values from incoming @OBJECT_SYNC@ @ProtocolMessages@ must be temporarily stored in the internal @SyncObjectsPool@ list
+** @(RTO5b)@ During the sync sequence, the @ObjectMessage.object@ values from incoming @OBJECT_SYNC@ @ProtocolMessages@ must be temporarily stored in the internal @SyncObjectMessages@ list
```
lawrence-forooghian, line 122: Given what you said in the PR description (about needing info from the `ObjectMessage`) I think that this and its subpoint need changing?

lawrence-forooghian:
@VeskeR are you still wishing to do this rename?

sacOO7:
Current PR is outdated !
`SyncObjectsPool` is actually a map, with key as `objectId` and value as `ObjectMessage`, so it doesn't make sense to rename it to `SyncObjectMessages`. 
Currently, the spec on `main` has been updated from `SyncObjectsPool list` to `SyncObjectsPool`. I don't think there's point in renaming this, since there's meaningful alterations already made to the spec

PR #386, opened 2025-10-02
chat: specify encoding of URL components

AndyTWF:
Specify that timeserials and roomnames should be url-encoded when sent to the Chat HTTP API.

### DIFF
--- a/textile/chat-features.textile
+++ b/textile/chat-features.textile
@@ -658,6 +658,8 @@ h3(#rest-general). General
 * @(CHA-RST1)@ REST API requests shall be made via the @request()@ method on the underling Ably SDK.
 * @(CHA-RST2)@ REST API requests shall use the API Protocol Version of the underlying core SDK.
 * @(CHA-RST3)@ @[Testable]@ REST API requests must be supported using @JSON@ and @msgpack@ as a @Content-Type@ (@useBinaryProtocol@ on the underlying Ably SDK)
+* @(CHA-RST4)@ @[Testable]@ Where a URL path contains a @roomName@ variable, that value shall be URL-encoded.
+* @(CHA-RST5)@ @[Testable]@ Where a URL path contains a @timeserial@ variable, that value shall be URL-encoded.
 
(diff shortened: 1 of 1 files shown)


textile/chat-features.textile
```diff
@@ -658,6 +658,8 @@ h3(#rest-general). General
 * @(CHA-RST1)@ REST API requests shall be made via the @request()@ method on the underling Ably SDK.
 * @(CHA-RST2)@ REST API requests shall use the API Protocol Version of the underlying core SDK.
 * @(CHA-RST3)@ @[Testable]@ REST API requests must be supported using @JSON@ and @msgpack@ as a @Content-Type@ (@useBinaryProtocol@ on the underlying Ably SDK)
+* @(CHA-RST4)@ @[Testable]@ Where a URL path contains a @roomName@ variable, that value shall be URL-encoded.
+* @(CHA-RST5)@ @[Testable]@ Where a URL path contains a @timeserial@ variable, that value shall be URL-encoded.
```
ttypic, line 662: do we really need to URL encode `timeserial`?

AndyTWF: There are some characters in the timeserial that aren't particularly URL-safe, e.g. `@` and `:`- so we felt it best to have them encoded.

ttypic: my bad, it should be

ttypic (approved):

textile/chat-features.textile
maratal: Is it even `timeserial`? I thought this name was obsolete and became just `serial`.

AndyTWF: The field in question on the `message` is called `serial`, but the string in that particular format is still a "timeserial string", so we could argue either-or - I went with `timeserial` as we'd want the same to apply to any timeserial, whether it's in `serial`, `messageSerial`, `fooSerial`

maratal: BTW when using `CharacterSet.urlPathAllowed` it doesn't escapes `@`, so I guess it's fine to use it in path.

PR #429, opened 2026-02-26
Delete redundant object type check when applying `SyncObjectsPool`

lawrence-forooghian:
RTO5f3 (added in 9f4d7de) rejects unsupported object types before they enter the `SyncObjectsPool`, so the RTO5c1b1c check during pool application can never be reached. Furthermore, testing RTO5c1b1c would require putting the SDK into an internal state that is inconsistent with the rest of the spec — a `SyncObjectsPool` containing an unsupported object type — which should not be possible.


VeskeR:
Sorry, my bad 🤦‍♂️ . I must've misread the PR yesterday, it was a looong day of PR feedback reviews/addressing.
LGTM

VeskeR (approved):

PR #432, opened 2026-03-04
[AIT-466] Spec for MAP_CLEAR operation

VeskeR:
Spec for a MAP_CLEAR operation described in DR https://ably.atlassian.net/wiki/x/DABECAE. Realtime implementation in https://github.com/ably/realtime/pull/8074.

ably-js implementation https://github.com/ably/ably-js/pull/2176

Resolves [AIT-466](https://ably.atlassian.net/browse/AIT-466)

[AIT-466]: https://ably.atlassian.net/browse/AIT-466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specifications/objects-features.md
+++ b/specifications/objects-features.md
@@ -346,15 +346,19 @@ Objects feature enables clients to store shared data as "objects" on a channel. 
     - `(RTLO4e1)` Expects the following arguments:
       - `(RTLO4e1a)` `ObjectMessage`
     - `(RTLO4e2)` Set `LiveObject.isTombstone` to `true`
-    - `(RTLO4e3)` Set `LiveObject.tombstonedAt` as follows:
-      - `(RTLO4e3a)` Set it equal to `ObjectMessage.serialTimestamp` if it exists
-      - `(RTLO4e3b)` Otherwise, set it to the current time using the local clock
-        - `(RTLO4e3b1)` Log a debug or trace message indicating that `serialTimestamp` was not found in the message and the local clock is being used instead for the tombstone timestamp
+    - `(RTLO4e3)` Set `LiveObject.tombstonedAt` to the value calculated per [RTLO6](#RTLO6), using `ObjectMessage.serialTimestamp`
+      - `(RTLO4e3a)` This clause has been replaced by [RTLO6a](#RTLO6a)
+      - `(RTLO4e3b)` This clause has been replaced by [RTLO6b](#RTLO6b)
+        - `(RTLO4e3b1)` This clause has been replaced by [RTLO6b1](#RTLO6b1)
     - `(RTLO4e4)` Set the data for the `LiveObject` to a zero-value, as described in [RTLC4](#RTLC4) or [RTLM4](#RTLM4) depending on the object type
 - `(RTLO5)` An `OBJECT_DELETE` operation can be applied to a `LiveObject` in the following way:
   - `(RTLO5a)` Expects the following arguments:
     - `(RTLO5a1)` `ObjectMessage`
   - `(RTLO5b)` Tombstone the current `LiveObject` using [`LiveObject.tombstone`](#RTLO4e), passing in the `ObjectMessage`
+- `(RTLO6)` A `tombstonedAt` value can be calculated from a provided `serialTimestamp` as follows:
+  - `(RTLO6a)` It is equal to `serialTimestamp` if it exists
+  - `(RTLO6b)` Otherwise, it is equal to the current time using the local clock
+    - `(RTLO6b1)` Log a debug or trace message indicating that `serialTimestamp` was not provided and the local clock is being used instead for the tombstone timestamp
 
 ### LiveCounter
 
@@ -567,10 +571,10 @@ Objects feature enables clients to store shared data as "objects" on a channel. 
   - `(RTLM6g)` Store the current `data` value as `previousData` for use in [RTLM6h](#RTLM6h)
   - `(RTLM6b)` Set the private flag `createOperationIsMerged` to `false`
   - `(RTLM6c)` Set `data` to `ObjectState.map.entries`, or to an empty map if it does not exist
-    - `(RTLM6c1)` For each `ObjectsMapEntry` with `ObjectsMapEntry.tombstone` equal to `true`, additionally set the `ObjectsMapEntry.tombstonedAt` field as follows:
-      - `(RTLM6c1a)` Set it equal to `ObjectsMapEntry.serialTimestamp` if it exists
-      - `(RTLM6c1b)` Otherwise, set it to the current time using the local clock
-        - `(RTLM6c1b1)` Log a debug or trace message indicating that `ObjectsMapEntry.serialTimestamp` was not provided and the local clock is being used instead for the tombstone timestamp
+    - `(RTLM6c1)` For each `ObjectsMapEntry` with `ObjectsMapEntry.tombstone` equal to `true`, additionally set the `ObjectsMapEntry.tombstonedAt` field to the value calculated per [RTLO6](#RTLO6), using `ObjectsMapEntry.serialTimestamp`
+      - `(RTLM6c1a)` This clause has been replaced by [RTLO6a](#RTLO6a)
+      - `(RTLM6c1b)` This clause has been replaced by [RTLO6b](#RTLO6b)
+        - `(RTLM6c1b1)` This clause has been replaced by [RTLO6b1](#RTLO6b1)
   - `(RTLM6d)` If `ObjectState.createOp` is present, merge the initial value into the `LiveMap` as described in [RTLM23](#RTLM23), passing in the `ObjectState.createOp` instance. Discard the `LiveMapUpdate` object returned by the merge operation
     - `(RTLM6d1)` This clause has been replaced by [RTLM17a](#RTLM17a)
       - `(RTLM6d1a)` This clause has been replaced by [RTLM17a1](#RTLM17a1)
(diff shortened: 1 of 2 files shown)


textile/objects-features.textile
```diff
@@ -640,16 +650,31 @@ h3(#livemap). LiveMap
...
+* @(RTLM24)@ A @MAP_CLEAR@ operation can be applied to a @LiveMap@ in the following way:
+** @(RTLM24a)@ Expects the following arguments:
+*** @(RTLM24a1)@ @serial@ string - the operation's serial value
+*** @(RTLM24a2)@ @serialTimestamp@ Time - the operation's serial timestamp value
+** @(RTLM24b)@ The return type is a @LiveMapUpdate@ object, which indicates the data update for this @LiveMap@
+** @(RTLM24c)@ If the private @clearTimeserial@ is non-null and is lexicographically greater than or equal to the provided @serial@, discard the operation without taking any action. Return a @LiveMapUpdate@ object with @LiveMapUpdate.noop@ set to @true@, indicating that no update was made to the object
+** @(RTLM24d)@ Set the private @clearTimeserial@ to the provided @serial@
+** @(RTLM24e)@ For each @ObjectsMapEntry@ in the internal @data@:
+*** @(RTLM24e1)@ If @ObjectsMapEntry.timeserial@ is null or empty, or the @serial@ is lexicographically greater than or equal to @ObjectsMapEntry.timeserial@:
+**** @(RTLM24e1a)@ Set @ObjectsMapEntry.data@ to undefined/null
```
lawrence-forooghian: I'm not sure I follow why we need to tombstone these entries, instead of just removing them from the `data`? The RTLM7h check prevents a `MAP_SET` from resurrecting them.

lawrence-forooghian: (A `MAP_REMOVE` for a key that didn't exist at the time of applying the `MAP_CLEAR` would still create an unnecessary tombstone, but that'd be addressed if we implemented https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1772807443777899?thread_ts=1772801262.669639&cid=C09SY1AQGK0)

lawrence-forooghian: Actually I think this is similar to Paddy's question here: https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1772807518911119?thread_ts=1772801262.669639&cid=C09SY1AQGK0

> Why aren't all entries that predated the clear just removed? I don't see why we would keep tombstones for them

textile/objects-features.textile
```diff
@@ -609,6 +618,7 @@ h3(#livemap). LiveMap
 *** @(RTLM7d3)@ @MapSet@
 *** @(RTLM7d2)@ @serial@ string - operation's serial value
 ** @(RTLM7e)@ The return type is a @LiveMapUpdate@ object, which indicates the data update for this @LiveMap@
+** @(RTLM7h)@ If the private @clearTimeserial@ is non-null, and the provided @serial@ is null or the @clearTimeserial@ is lexicographically greater than or equal to @serial@, discard the operation without taking any action. Return a @LiveMapUpdate@ object with @LiveMapUpdate.noop@ set to @true@, indicating that no update was made to the object
```
lawrence-forooghian: As discussed in https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1772801262669639 I think we want a similar check for `MAP_REMOVE`.

textile/features.textile
```diff
@@ -1756,7 +1763,7 @@ h4. ObjectsMapEntry
 * @(OME1)@ An @ObjectsMapEntry@ represents the value at a given key in an @ObjectsMap@ object
 * @(OME2)@ The attributes available in an @ObjectsMapEntry@ are:
 ** @(OME2a)@ @tombstone@ boolean - indicates whether the map entry has been removed
-** @(OME2b)@ @timeserial@ string - the @serial@#OM2h value of the last operation that was applied to the map entry
+** @(OME2b)@ @timeserial@ string - the "serial"#OM2h value of the last operation that was applied to the map entry
```
lawrence-forooghian: Needs a colon to be a link

textile/objects-features.textile
```diff
@@ -462,6 +462,7 @@ h3(#livemap). LiveMap
 * @(RTLM3)@ Holds a @Dict<String, ObjectsMapEntry>@ as a private @data@ map
 ** @(RTLM3a)@ @ObjectsMapEntry@ entries in a @LiveMap@ have the following attributes in addition to those defined in "OME2":../features#OME2:
 *** @(RTLM3a1)@ @tombstonedAt@ (optional) Time - a timestamp indicating when this map entry was tombstoned. This property is nullable, and specification points that manipulate this value maintain the invariant that it is non-null if and only if the corresponding @ObjectsMapEntry.tombstone@ is @true@
+* @(RTLM25)@ Holds a nullable private @clearTimeserial@ string, initially @null@. Set to the "serial":../features#OM2h of the last @MAP_CLEAR@ operation applied to this @LiveMap@
```
lawrence-forooghian: Reads like an instruction but actually this is an invariant maintained by other specification points

textile/objects-features.textile
paddybyers: If it helps, this was my attempt at a spec for `MAP_CLEAR`: https://github.com/ably/uts/commit/30dbbf7c971f91381bee40ab53c7b067399249d0

specifications/objects-features.md
```diff
@@ -665,6 +671,21 @@ Objects feature enables clients to store shared data as "objects" on a channel.
...
   - `(RTLM8e)` Return a `LiveMapUpdate` object with a `LiveMapUpdate.update` map containing the key used in this operation set to `removed`
+- `(RTLM24)` A `MAP_CLEAR` operation can be applied to a `LiveMap` in the following way:
+  - `(RTLM24a)` Expects the following arguments:
+    - `(RTLM24a1)` `serial` string - the operation's serial value
+    - `(RTLM24a2)` `serialTimestamp` Time - the operation's serial timestamp value
+  - `(RTLM24b)` The return type is a `LiveMapUpdate` object, which indicates the data update for this `LiveMap`
+  - `(RTLM24c)` If the private `clearTimeserial` is non-null and is lexicographically greater than or equal to the provided `serial`, discard the operation without taking any action. Return a `LiveMapUpdate` object with `LiveMapUpdate.noop` set to `true`, indicating that no update was made to the object
+  - `(RTLM24d)` Set the private `clearTimeserial` to the provided `serial`
+  - `(RTLM24e)` For each `ObjectsMapEntry` in the internal `data`:
+    - `(RTLM24e1)` If `ObjectsMapEntry.timeserial` is null or empty, or the `serial` is lexicographically greater than or equal to `ObjectsMapEntry.timeserial`:
```
lawrence-forooghian: What does "empty" mean here? Empty string? Is that a thing that can actually happen? Claude started writing a bunch of checks and tests for this case and I'm not convinced we need them

specifications/objects-features.md
```diff
@@ -651,16 +661,31 @@ Objects feature enables clients to store shared data as "objects" on a channel.
...
```
lawrence-forooghian: I think you asked this on Slack — the "or equal to" should be dropped here right?

lawrence-forooghian:
For the WIP Swift implementation I've been using this branch which incorporates the discussed changes: https://github.com/ably/specification/tree/AIT-466/map-clear-Lawrence-modfications-for-swift

lawrence-forooghian:
In the `ATTACH` with `!HAS_OBJECTS` case (RTO4b), what is the `clearTimeserial` of the root map meant to be set to? Can we ever receive `!HAS_OBJECTS` after a clear operation has happened to the root map?

(It would help if I knew the semantics of `HAS_OBJECTS`: does it mean "there are no objects on the channel" — which is what its name implies, or does it mean "there is only a root map on the channel, and maybe this root map used to contain other objects"?)

Update: I think a similar question for `siteTimeserials`, so might not be something we need to sort out as part of this — asked in https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1773168367642749

textile/objects-features.textile
VeskeR: changed in https://github.com/ably/specification/pull/432/changes/68295d8f0c83fe6decd0f600df0b8088e7c08123. see RTLM24e1a, we now remove entries without tombstoning them

textile/objects-features.textile
VeskeR: Added RTLM8g in https://github.com/ably/specification/pull/432/changes/68295d8f0c83fe6decd0f600df0b8088e7c08123

textile/features.textile
VeskeR: fixed in https://github.com/ably/specification/pull/432/changes/68295d8f0c83fe6decd0f600df0b8088e7c08123

textile/objects-features.textile
VeskeR: fixed in https://github.com/ably/specification/pull/432/commits/26bda6ffabfe648c9784215ee79e668201c441b8

specifications/objects-features.md
VeskeR: Sorry, meant "omitted", fixed in https://github.com/ably/specification/pull/432/commits/26bda6ffabfe648c9784215ee79e668201c441b8

specifications/objects-features.md
VeskeR: changed to "greater than" only in https://github.com/ably/specification/pull/432/changes/68295d8f0c83fe6decd0f600df0b8088e7c08123

VeskeR:
> (It would help if I knew the semantics of HAS_OBJECTS: does it mean "there are no objects on the channel" — which is what its name implies, or does it mean "there is only a root map on the channel, and maybe this root map used to contain other objects"?)

Yeah, I think we just need to clarify what exactly HAS_OBJECTS mean, specifically HAS_OBJECTS=false (honestly I have never seen it set to false, maybe only when you attach without object modes?).

I suspect the correct approach would be to just set clear timeserial to the null value when we reset a map to a zero-value (including when we clear root as a result of HAS_OBJECTS=false), so for now I've updated RTLM4 in https://github.com/ably/specification/pull/432/changes/68295d8f0c83fe6decd0f600df0b8088e7c08123. This also matches Paddy's implementation in https://github.com/ably/uts/commit/30dbbf7c971f91381bee40ab53c7b067399249d0#diff-5b1d0405c9ae22960faca135f58646eb6a7a8bc938c62cf926e2629522bff465R544

lawrence-forooghian:
LGTM — happy to approve once squashed, but it would be good to understand the correct RTLM4 behaviour

VeskeR:
Summary of the outcome for our HAS_OBJECTS flag discussion in [slack](https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1773168367642749):

A client subscribed to a channel with `OBJECT_SUBSCRIBE` mode is guaranteed to always receive `HAS_OBJECTS=true`. This means: if you want objects state, the server will always send you a sync for it.

`MAP_CLEAR` and `clearTimeserial` implication:

- `HAS_OBJECTS=false` only occurs when `OBJECT_SUBSCRIBE` mode is **not** set - meaning the client doesn't have read access to the objects API anyway.
- Therefore, on `HAS_OBJECTS=false`, it's safe to clear the root map and reset its `clearTimeserial` to null/empty.
- On mode changes, if `OBJECT_SUBSCRIBE` is set, the server will send `HAS_OBJECTS=true` followed by a sync with populated `clearTimeserial`.

In short: `HAS_OBJECTS=false` handling is safe to implement as a reset of a root map to its zero value (RTLM4, which now includes setting `clearTimeserial` to null), because that condition is only reachable when the client has no objects access anyway - and the server will drive a fresh sync whenever access is (re)established with `OBJECT_SUBSCRIBE`.

No changes are needed to this PR.

cc @lawrence-forooghian @ttypic

VeskeR:
@lawrence-forooghian squashed and rebased on main, ready for approval

lawrence-forooghian (approved):

PR #434, opened 2026-03-06
feat!: textile to md migration

ttypic:
Replace textile with Markdown with preserved git history, for better LLM usage

Also migrate to Hugo for rendering:

- Replace `meta.yaml` with `hugo.yaml` for site configuration.
- Add Hugo layouts and templates.
- Update TailwindCSS configuration and build output paths.
- Convert build script to prepare Hugo content.
- Modernize workflows and dependencies.
- Rewrite duplicate-spec detection script in Bash.


SimonWoolf (commented):
surprised at the decision to wrap lines, I'd have thought we'd want to have one line per spec item? Easier for LLMs and cleaner diffs without rewrapping noise. (And if we are going to have wrapping we should have an autoformatter and a max_line_length specified in the editorconfig)

ttypic:
@SimonWoolf unfortunately it wasn't my decision, pandoc, that I used for transformation, does wrapping unconditionally. Not sure line breaks change how LLM process input (although it's additional tokens and with our spec size it can be easy couple of thousands of additional tokens). I'll look into if we can preserve original line breaks

paddybyers:
Surely a quick grep/sed can fix it. No need to fix the initial transformation; just make sure that the final result is in the correct form

lawrence-forooghian:
> surprised at the decision to wrap lines, I'd have thought we'd want to have one line per spec item?

Yeah, without an autoformatter nobody is going to preserve this wrapping as we add / edit stuff in the future — I don't think we need wrapping

> pandoc, that I used for transformation, does wrapping unconditionally

From memory having used pandoc in the past, I'm pretty sure there are options for controlling this

> No need to fix the initial transformation; just make sure that the final result is in the correct form

I think if it's a big faff to re-do the history again then let's keep it wrapped in the history, but if it's literally just re-run a script that does the conversion again for the whole history then why not, it will mean that `git blame` continues to be useful (given that Evgenii has put effort into preserving the commit history then I assume this is a goal) (Update: actually `git blame` might be a bad example here, but `git show --word-diff` is useful when looking at history and doesn't play well with wrapping)

ttypic:
Fixed line breaks @lawrence-forooghian @paddybyers @SimonWoolf

paddybyers (approved):
LGTM.

Feature-prioritisation is out of date, but we can address that as a separate PR

PR #438, opened 2026-03-11
chore: fix typographical errors and improve consistency in `features.md`

ttypic:
- Corrected formatting issues (`annotations`, `'@'` characters, etc.).
- Improved consistency in spec point references and removed redundant/unnecessary characters.

PR #439, opened 2026-03-11
fix spec point missed in markdown conversion

owenpearson:


ttypic (commented):
created fix minute a ago: https://github.com/ably/specification/pull/438

with couple of more fixes

PR #441, opened 2026-03-23
[WIP] address ably-cocoa push notification token storage issues

lawrence-forooghian:
Partially addresses the issue of ably-cocoa storing device secret in keychain and not wiping the device token when it fails to load the secret. WIP, and largely Claude-gnerated. There are still things to address here and need to think about whether we want to try and fetch a new token instead of just validating one (see commit messages for why haven't done this yet). Also we need to decide whether we're happy migrating to a storage for token and secret that's always available. Will finish add references to existing conversations once back from illness; we should discuss


lawrence-forooghian:
replaced by #450

PR #442, opened 2026-03-24
ADD RTL3d1 requiring channel state transitions to happen before the connected event

SimonWoolf:
This ensures that when a connection listener observes the `CONNECTED` state, channels that were previously `ATTACHED` are already in the `ATTACHING` state. Without this ordering guarantee, a connection listener calling `attach()` on a channel per `RTL4a` would see the channel as `ATTACHED` and receive an immediate success callback without a server round-trip, potentially leading to operations (such as presence enter) being performed on a channel that in reality is not attached


ttypic (approved):
LGTM

PR #445, opened 2026-03-30
Universal test suite

paddybyers:
# Universal test suite

This PR proposes is a common test suite for Ably SDKs. This supersedes the approach in https://github.com/ably-labs/unified-test-suite.

This approach is based on the idea that:
- we need every SDK to have a comprehensive unit test suite implemented using the standard test framework for that language;
- a single, portable definition of those unit tests should exist, with an automated way of translating tests to the target language and framework;
- the best language for that purpose is natural language (ie English) pseudocode;
- LLMs are now powerful enough to be able to perform the translation with a high level of reliability, so long as the natural language definition is sufficiently precise.

This PR is an implementation of this approach. This is a set of test specs for the majority of the pubsub SDK feature spec.

The tests make certain assumptions about there being features in the target SDKs’ API in order for it to be effectively unit-testable. These are:
- the HTTP client is mockable; the mock must be able to capture the requests that the library initiates, and respond in a way that is specified by the test;
- similarly, the websocket client must be mockable;
- timers and clock need to be mockable; under control of the test, it must be possible to advance time (ie advance the clock, and cause timers to fire that would have fired in that interval, and process all resulting events in the library).

I'm not asking for this PR to be reviewed line-by-line. I have individually reviewed every test spec as it has been created, and these specs have been exercised by generating a [Dart library](https://github.com/ably/ably-dart/pull/2) and a [Rust library](https://github.com/ably/ably-rust/pull/35). Feedback from each of those has been incorporated as bugfixes into this branch.

I propose that the next validation step on this set of test specs is to use them to generate tests for an existing library known to have good conformance with the spec. We will need to modify that target library so that the HTTP client etc are mockable in the way expected by these tests. Having done that, we would run the resulting tests against the library and, for any test failures, decide whether the problem is with the target library, or the test spec, or elsewhere.


paddybyers:
Closing in favour of https://github.com/ably/specification/pull/456

PR #454, opened 2026-04-21
[AIT-530] Add `CLAUDE.md`

lawrence-forooghian:
**Note: This is a draft and contains some Claude-generated research materials that were used for generating the file, which I intend to remove once the PR is approved. There are also a couple of TODOs to address still, but I wanted to get some initial opinions on the overall structure before going on holiday for a couple of weeks; I'll address feedback once back.**

This is based on having asked Claude to go through comments left by Simon and I in all the PRs on the specification repo; hopefully this has extracted a good amount of the team's unwritten opinions and knowledge about how the spec should be structured. I've then tweaked it a fair bit myself, but it's still very much Claude output that we should refine and expand over time.

The context for writing this was that Matt O got Claude to write a spec PR (https://github.com/ably/specification/pull/433) for a new channel param, and was promptly told that it didn't belong in the spec nor should the spec be specifying tests. This led him to rightly not be very impressed by learning about the existence of a bunch of rules that had never been written down anywhere, and suggested that we write them down; thus this commit.

It also led to a broader conversation about feature discoverability and testing; see [this Slack thread](https://ably-real-time.slack.com/archives/CURL4U2FP/p1776169126491229). The relevant parts of that conversation ("some stuff doesn't belong in the spec but should be tracked for inclusion in these upcoming things") have been included in the `CLAUDE.md` to guide people towards the right place for ensuring that a new feature gets included in these upcoming things even if it turns out it doesn't belong in the spec.

I've focussed this `CLAUDE.md` on how to write specification points, as opposed to describing the build system for this repo, which most people who make spec changes never touch.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,92 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What This Is
+
+The Ably features specification — the authoritative, language-agnostic description of how Ably SDKs must behave. Consumed by SDK developers across all Ably SDK repositories (pub/sub, chat, objects).
+
+## Files You'll Edit
+
+The spec lives in `specifications/`:
+
+- `features.md` — REST/Realtime spec; the main file.
+- `chat-features.md` — Chat SDK spec.
+- `objects-features.md` — LiveObjects spec.
+- `api-docstrings.md` — language-agnostic docstrings that SDKs lightly edit and add to their code in order to support autocomplete and generate HTML documentation.
+- `protocol.md`, `encryption.md`, `test-api.md`, `feature-prioritisation.md`, `index.md` — supporting docs, rarely edited.
+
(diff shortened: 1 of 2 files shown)


SimonWoolf:
Thanks for doing this. Rather than making lots of small rewording/addition comments and suggestions I might just make my proposed changes in a branch and pr against this branch

Edit: https://github.com/ably/specification/pull/455

lawrence-forooghian:
@SimonWoolf I've cherry-picked your commit from #455. 

I've also dropped the earlier commit that contained the research data (have left it on `AIT-530-claude-md-research-data` just in case we want to come back to it although I think it's served its purpose).

If you're happy with this now Simon then it'd be good to get this one merged and draw a line under it; my attitude to files like this is that something is better than nothing and you only properly find out what it needs when you start trying to build stuff with it. But I think it captures a bunch of valuable stuff.

SimonWoolf (approved):

PR #457, opened 2026-04-30
feat[uts]: add test for fallback host expiration logic (RSC15f)

ttypic:
`RSC15f` states that the client should store successful fallback host for `ClientOptions.fallbackRetryTimeout`. A naive implementation stores fallback on every successful response without checking whether this fallback already expired and cleared by concurrent call. This means a long in-flight request that started while a fallback was preferred, but completes after the timeout has elapsed, silently re-pins the expired host with a fresh timer.


paddybyers:
Thanks. I've brought it across into https://github.com/ably/specification/pull/462/changes/fcfb8a512c89df3b2a6f80a6d070f255ea7f9fd7, so closing this PR

PR #469, opened 2026-05-07
Don't send `echo=true` connection param

### LINKED ISSUE #465: RTC1a needlessly requires sending the `echo` param with a default value
See https://github.com/ably/ably-js/issues/2206.

The spec says:

> RTC1a: The echoMessages option (default true) should be sent as an echo=true query parameter on the WebSocket connection URL.

There is no need to send `echo=true` since this is the default behaviour

lawrence-forooghian:
The server's default behaviour is equivalent to `echo=true`.

Resolves #465.

### DIFF
--- a/specifications/features.md
+++ b/specifications/features.md
@@ -460,7 +460,7 @@ The threading and/or asynchronous model for each realtime library will vary by l
 
 - `(RTC12)` Has the same constructors as `RestClient`, as defined in [RSC1](#RSC1)
 - `(RTC1)` Supports all the same `ClientOptions` as the `RestClient` in addition to:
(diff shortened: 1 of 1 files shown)


SimonWoolf (approved):
not sure I see the need to deprecate RTN2b in favour of a new RTN2h given this is just restoring RTN2b to what it said originally, but shrug

lawrence-forooghian:
> not sure I see the need to deprecate RTN2b in favour of a new RTN2h given this is just restoring RTN2b to what it said originally, but shrug

"originally" when? because as mentioned in https://github.com/ably/specification/pull/198#issuecomment-4397830909, the pre-that-PR-version of that spec point also suggested you were meant to always send the param

PR #472, opened 2026-05-12
Implicit attach on `channel.objects.get()` call

lawrence-forooghian:
As already implemented in JS in https://github.com/ably/ably-js/commit/9bde15.

Split out of #427 since it's a standalone behaviour.

### DIFF
--- a/specifications/features.md
+++ b/specifications/features.md
@@ -820,6 +820,11 @@ The threading and/or asynchronous model for each realtime library will vary by l
   - `(RTL32e)` Any params provided in the third argument must be sent in the `TR4q` `ProtocolMessage.params` field, as a `Dict<String, String>` (that is, with the values of the provided dict stringified per `RTC1f`)
   - `(RTL32c)` The SDK must not mutate the user-supplied `Message` object.
   - `(RTL32d)` On success, returns an `UpdateDeleteResult` object containing the version serial of the published update, obtained from the first element of the `serials` array of the [TR4s](#TR4s) `res` field of the `ACK`. Indicates an error if the operation was not successful.
+- `(RTL33)` *Ensure-active-channel* internal procedure. When invoked, the SDK must inspect the current state of the `RealtimeChannel` (see [RTL2](#RTL2)) and proceed as follows:
+  - `(RTL33a)` If the channel is in the `ATTACHED` or `SUSPENDED` state, the procedure completes immediately without performing any attach.
+  - `(RTL33b)` If the channel is in the `INITIALIZED`, `DETACHED`, `DETACHING`, or `ATTACHING` state, perform an implicit attach per [RTL4](#RTL4) and wait for it to complete
+    - `(RTL33b1)` If the implicit attach fails, the procedure must reject with the same `ErrorInfo` that caused the attach to fail
+  - `(RTL33c)` If the channel is in the `FAILED` state, the procedure must throw an `ErrorInfo` with `statusCode` 400 and `code` 90001, indicating that the channel operation failed due to the current channel state
 
 ### RealtimePresence {#realtime-presence}
 
@@ -913,8 +918,9 @@ The threading and/or asynchronous model for each realtime library will vary by l
   - `(RTP10e)` In all other ways, this method is identical to `RealtimePresence#enter` and should have matching tests
 - `(RTP11)` `RealtimePresence#get` function:
   - `(RTP11a)` Returns the list of current members on the channel in a callback. By default, will wait for the `SYNC` to be completed, see [RTP11c1](#RTP11c1)
-  - `(RTP11b)` Implicitly attaches the `RealtimeChannel` if the channel is in the `INITIALIZED` state. However, if the channel is in or enters the `DETACHED` or `FAILED` state before the operation succeeds, it will result in an error
+  - `(RTP11b)` This clause has been replaced by [RTP11e](#RTP11e)
   - `(RTP11d)` If the `RealtimeChannel` is in the `SUSPENDED` state then the `get` function will by default, or if `waitForSync` is set to `true`, result in an error with `code` `91005` and a `message` stating that the presence state is out of sync due to the channel being in a `SUSPENDED` state. If however the `get` function is called with `waitForSync` set to `false`, then it immediately returns the members currently stored in the `PresenceMap` giving developers access to the members that were present at the time the channel became `SUSPENDED`
+  - `(RTP11e)` If the channel is in any state other than `SUSPENDED` (which is handled by [RTP11d](#RTP11d)), perform the *ensure-active-channel* procedure ([RTL33](#RTL33)) on the underlying `RealtimeChannel`. If the procedure fails, the `get` function must reject with the same `ErrorInfo` that caused the procedure to fail
   - `(RTP11c)` An optional set of params can be provided:
     - `(RTP11c1)` `waitForSync` (default `true`). When `true`, method will wait until `SYNC` is complete before returning a list of members. When `false`, known set of presence members is returned immediately, which may be incomplete if the `SYNC` is not finished
     - `(RTP11c2)` `clientId` filters members by the provided `clientId`
--- a/specifications/objects-features.md
+++ b/specifications/objects-features.md
@@ -17,7 +17,8 @@ Objects feature enables clients to store shared data as "objects" on a channel. 
   - `(RTO1d)` This clause has been replaced by [RTO23d](#RTO23d).
 - `(RTO23)` `RealtimeObject#get` function:
   - `(RTO23a)` Requires the `OBJECT_SUBSCRIBE` channel mode to be granted per [RTO2](#RTO2)
-  - `(RTO23b)` If the channel is in the `DETACHED` or `FAILED` state, the library should throw an `ErrorInfo` error with `statusCode` 400 and `code` 90001
+  - `(RTO23b)` This clause has been replaced by [RTO23e](#RTO23e)
+  - `(RTO23e)` Perform the *ensure-active-channel* procedure ([RTL33](../features#RTL33)) on the underlying `RealtimeChannel`. If the procedure fails, the `get` function must reject with the same `ErrorInfo` that caused the procedure to fail
   - `(RTO23c)` If the [RTO17](#RTO17) sync state is not `SYNCED`, waits for the sync state to transition to `SYNCED`
   - `(RTO23d)` Returns a new `PathObject` ([RTPO1](#RTPO1)) with `path` ([RTPO2a](#RTPO2a)) set to an empty list and `root` ([RTPO2b](#RTPO2b)) set to the `LiveMap` with id `root` from the internal `ObjectsPool`
 - `(RTO11)` This clause has been replaced by [RTLMV3](#RTLMV3).


sacOO7:

I've cross-checked this PR against ably-js [`realtimechannel.ts:1135-1150`](https://github.com/ably/ably-js/blob/3deeee8e/src/common/lib/client/realtimechannel.ts#L1135-L1150) (`ensureAttached`) and [`realtimeobject.ts:88-101`](https://github.com/ably/ably-js/blob/3deeee8e/src/plugins/liveobjects/realtimeobject.ts#L88-L101) (`get()`). The new RTO1e/f line up with the implementation for the four states they enumerate, but there are a couple of gaps worth tightening before merge.

## 🟠 `SUSPENDED` state behaviour not addressed

The new RTO1e/f cover `INITIALIZED`, `DETACHED`, `DETACHING`, `ATTACHING`, and `FAILED`, but say nothing about `SUSPENDED`. The implementation's `ensureAttached` ([source](https://github.com/ably/ably-js/blob/3deeee8e/src/common/lib/client/realtimechannel.ts#L1135-L1150)) explicitly passes `SUSPENDED` through without re-attaching:

```ts
async ensureAttached(): Promise<void> {
  switch (this.state) {
    case 'attached':
    case 'suspended':
      break;
    case 'initialized':
    case 'detached':
    case 'detaching':
    case 'attaching':
      await this.attach();
      break;
    case 'failed':
    default:
      throw ErrorInfo.fromValues(this.invalidStateError());
  }
}
```

That's deliberate — a `SUSPENDED` channel may have local object data that the user can still read while the connection is being re-established. The current RTO1 wording is ambiguous about this case (a reader might assume `SUSPENDED` falls into "or any other non-attached state" and so gets re-attached, which is wrong).

Suggest adding:

```
- (RTO1g) If the channel is in the `ATTACHED` or `SUSPENDED` state, proceed without attaching.
```

…and renumbering / making the relationship between RTO1e/f/g explicit. The simplest framing matches the impl's switch statement:

| State | Behaviour |
|---|---|
| `ATTACHED`, `SUSPENDED` | Pass through; do not re-attach. |
| `INITIALIZED`, `DETACHED`, `DETACHING`, `ATTACHING` | Implicit attach and wait. |
| `FAILED` | Throw `ErrorInfo` 400/90001. |

## 🟠 Behaviour during pending implicit-attach if the channel transitions to `FAILED`

RTO1e says "implicitly attach the `RealtimeChannel` and wait for the attach to complete" but doesn't define what happens if the attach attempt itself fails, or if the channel transitions to `FAILED` during the wait. The impl simply propagates the `attach()` rejection ([`realtimechannel.ts:1144`](https://github.com/ably/ably-js/blob/3deeee8e/src/common/lib/client/realtimechannel.ts#L1144) → standard `attach()` failure semantics).

Worth being explicit:

```
- (RTO1e1) If the implicit attach fails (e.g. the channel transitions to `FAILED` during the wait), the returned promise MUST reject with an `ErrorInfo` 400/90001.
```

## 🟡 RFC 2119 keyword precision

`(RTO1f)` says "the library **should** throw" — should this be MUST? The pre-existing `(RTO1b)` used the same "should" wording, but per CLAUDE.md's writing principles ("Use RFC 2119 keywords precisely"), throwing for a `FAILED` channel is a hard requirement. Suggest "MUST throw".

## ⚪ Minor: cross-reference to `RTL27` lifecycle

For readers landing on this clause without context, a one-line cross-reference to channel-state transitions ([`RTL2`](https://github.com/ably/specification/blob/main/specifications/features.md#RTL2)) and the implicit-attach pattern used elsewhere (e.g. [`RTP11b`](https://github.com/ably/specification/blob/main/specifications/features.md#RTP11b) for `presence.get()`, mentioned in the PR description) would help with consistency framing.

---

These observations correspond to issue B-3 in `LIVEOBJECTS_SPEC_REVIEW.md` (cross-branch review) and Task B-6 in `LIVEOBJECTS_UPDATE_SPEC_ACTIONABLE_PLAN.md`.

sacOO7 (approved):
LGTM.
Addressed changes [as per review comment](https://github.com/ably/specification/pull/472#issuecomment-4477350788) and now behaviour is exactly similar to JS code.

specifications/objects-features.md
```diff
@@ -12,7 +12,12 @@ Objects feature enables clients to store shared data as "objects" on a channel.
 
 - `(RTO1)` `RealtimeObjects#getRoot` function:
   - `(RTO1a)` Requires the `OBJECT_SUBSCRIBE` channel mode to be granted per [RTO2](#RTO2)
-  - `(RTO1b)` If the channel is in the `DETACHED` or `FAILED` state, the library should throw an `ErrorInfo` error with `statusCode` 400 and `code` 90001
+  - `(RTO1b)` This clause has been replaced by [RTO1e](#RTO1e).
+  - `(RTO1e)` Performs the *ensure-attached* procedure to make sure the underlying `RealtimeChannel` is in the `ATTACHED` state before proceeding.
```
sacOO7: As per [ably-js#ensureAttached](https://github.com/ably/ably-js/blob/3deeee8e/src/common/lib/client/realtimechannel.ts#L1135-L1150) method comment description, 
```
   * This method is intended for use by features like Presence or Objects that need to
   * implicitly attach the channel when an operation is called (e.g., `presence.get()` per RTP11b,
   * or `objects.get()`). This guarantees that the corresponding sync sequence will start and
   * that the operation will resolve for callers even if they did not explicitly attach beforehand.
   */
```
Currently, I have limited scope to `objects` spec, otherwise `RTO1e`  is supposed to be shared by both `presence.get` and `objects.get`. Currently, [the spec `RTP11b` is incomplete](https://github.com/ably/specification/blob/main/specifications/features.md#:~:text=completed%2C%20see%20RTP11c1-,(RTP11b)%20Implicitly,-attaches%20the%20RealtimeChannel) for `presence.get` : (

So, maybe we can extract `RTO1e` and share it with both `presence.get` and `objects.get`
wdyt @lawrence-forooghian @ttypic

lawrence-forooghian: > to make sure the underlying `RealtimeChannel` is in the `ATTACHED` state before proceeding

This isn't quite true since the procedure indicates success when `SUSPENDED`, too. Ditto calling it "ensure attached" (I know the name is taken from ably-js but I think it's wrong there too)

And yes, to your other question I think we should refactor the spec to share the same logic between presence and objects (but this shared procedure should go in the main spec, not objects spec)

specifications/objects-features.md
```diff
@@ -12,7 +12,12 @@ Objects feature enables clients to store shared data as "objects" on a channel.
...
+    - `(RTO1e1)` If the channel is in the `INITIALIZED`, `DETACHED`, `DETACHING`, or `ATTACHING` state, implicitly attach the `RealtimeChannel` and wait for the attach to complete
```
lawrence-forooghian: Can this be more explicit i.e. literally just defer to calling RTL4 `RealtimeChannel#attach`, and just state that we fail if it fails (and then we don't need the "for example" in RTO1e1a)

sacOO7: Fixed https://github.com/ably/specification/pull/472/commits/0d680cac49df46ed82b526d3abbb3d1c8a0ec14e

sacOO7: Actually, when performing `callback-style` operations in some languages ( like java [channel#attach](https://github.com/ably/ably-java/blob/28222727f17e48a56a66ca6128150ab381aec006/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L214)), it's easy to miss out on error callback where it can be silently omitted, so we need explicit spec point to make it clear.

specifications/objects-features.md
sacOO7: Updated https://github.com/ably/specification/pull/472/commits/0d680cac49df46ed82b526d3abbb3d1c8a0ec14e

sacOO7: Updated https://github.com/ably/specification/pull/472/commits/0d680cac49df46ed82b526d3abbb3d1c8a0ec14e

specifications/features.md
```diff
@@ -820,6 +820,11 @@ The threading and/or asynchronous model for each realtime library will vary by l
   - `(RTL32e)` Any params provided in the third argument must be sent in the `TR4q` `ProtocolMessage.params` field, as a `Dict<String, String>` (that is, with the values of the provided dict stringified per `RTC1f`)
   - `(RTL32c)` The SDK must not mutate the user-supplied `Message` object.
   - `(RTL32d)` On success, returns an `UpdateDeleteResult` object containing the version serial of the published update, obtained from the first element of the `serials` array of the [TR4s](#TR4s) `res` field of the `ACK`. Indicates an error if the operation was not successful.
+- `(RTL33)` *Ensure-active-channel* internal procedure. When invoked, the SDK MUST inspect the current state of the `RealtimeChannel` (see [RTL2](#RTL2)) and proceed as follows:
```
lawrence-forooghian: nit (throughout): we don't use capitalised MUST

specifications/features.md
```diff
@@ -820,6 +820,11 @@ The threading and/or asynchronous model for each realtime library will vary by l
...
+  - `(RTL33a)` If the channel is in the `ATTACHED` or `SUSPENDED` state, the procedure completes immediately without performing any attach.
+  - `(RTL33b)` If the channel is in the `INITIALIZED`, `DETACHED`, `DETACHING`, or `ATTACHING` state, perform an implicit attach per [RTL4](#RTL4) and wait for it to complete
```
lawrence-forooghian, line 825: This is a behavioural change compared to RTP11b i.e. a `DETACHED` channel now gets implicit attach, instead of throwing an error. Our docstrings aren't specific about the circumstances in which an implicit attach gets performed so I think we can get away with it, and I don't see any obvious reason why we wouldn't want to do it, but I thought it worth pointing out.

specifications/features.md
```diff
@@ -913,8 +918,9 @@ The threading and/or asynchronous model for each realtime library will vary by l
...
-  - `(RTP11b)` Implicitly attaches the `RealtimeChannel` if the channel is in the `INITIALIZED` state. However, if the channel is in or enters the `DETACHED` or `FAILED` state before the operation succeeds, it will result in an error
+  - `(RTP11b)` This clause has been replaced by [RTP11e](#RTP11e)
   - `(RTP11d)` If the `RealtimeChannel` is in the `SUSPENDED` state then the `get` function will by default, or if `waitForSync` is set to `true`, result in an error with `code` `91005` and a `message` stating that the presence state is out of sync due to the channel being in a `SUSPENDED` state. If however the `get` function is called with `waitForSync` set to `false`, then it immediately returns the members currently stored in the `PresenceMap` giving developers access to the members that were present at the time the channel became `SUSPENDED`
+  - `(RTP11e)` Perform the *ensure-active-channel* procedure ([RTL33](#RTL33)) on the underlying `RealtimeChannel`. Note that [RTP11d](#RTP11d) takes precedence over current spec point [RTP11e](#RTP11e)
```
lawrence-forooghian: I think we should make it clear that it rethrows its error (or however you want to word it). Ditto in RTO23e

specifications/objects-features.md
lawrence-forooghian: I am happy for us to say that if the implicit attach fails then this procedure should fail; my point was that we don't need the "for example" that explains reasons why the attach might fail.

specifications/features.md
lawrence-forooghian: I think we can be clearer about the circumstances in which RTP11e does not apply:

```suggestion
  - `(RTP11e)` If the channel is in any state other than `SUSPENDED` (which is handled by [RTP11d](#RTP11d)), perform the *ensure-active-channel* procedure ([RTL33](#RTL33)) on the underlying `RealtimeChannel`.
```

specifications/features.md
sacOO7: This behaviour follows current ably-js implementation -> https://github.com/ably/specification/pull/472#issuecomment-4477350788

specifications/objects-features.md
sacOO7: https://github.com/ably/specification/pull/472/changes/c5627f843728c5f182b913e8f8b57ab7ce5a4a36

specifications/features.md
sacOO7: https://github.com/ably/specification/pull/472/changes/c5627f843728c5f182b913e8f8b57ab7ce5a4a36

specifications/features.md
sacOO7: https://github.com/ably/specification/pull/472/changes/c5627f843728c5f182b913e8f8b57ab7ce5a4a36

specifications/features.md
sacOO7: https://github.com/ably/specification/pull/472/changes/c5627f843728c5f182b913e8f8b57ab7ce5a4a36

sacOO7: https://github.com/ably/specification/pull/472/changes/c5627f843728c5f182b913e8f8b57ab7ce5a4a36

specifications/features.md
lawrence-forooghian: > This behaviour follows current ably-js implementation

Sure, I get that; I'm saying that the ably-js implementation is different to the spec and it's not obvious that the ably-js behaviour is the one that we want (but that it's minor)

lawrence-forooghian:
LGTM, I can't approve because it's my PR, but approved

PR #474, opened 2026-05-13
Clear `deviceIdentityToken` when we fail to load `id` or `deviceSecret`

lawrence-forooghian:
A couple of minor bits and then a replacement of RSH8j (which is not implemented in most of our SDKs AFAIK — the important one in which it's not implemented is ably-cocoa since that's the one in which load failure is likely to happen, tracked in https://github.com/ably/ably-cocoa/issues/1257). Intention is to make sure that you don't end up with a `deviceIdentityToken` that belongs to a previous `id` and to make it clearer how to implement "become `NotActivated`".

See commit messages for more details.


maratal:
There is a lot of back and forth between RSH8a, RSH8b, RSH3h, RSH8k2, RSH3a2b which makes it very hard to follow. Can we keep such a thing as "load details or generate if it doesn't exist or loading failed" in at least two places?

Earlier this year I've tried to fit specs with current cocoa behavior and reducing amount of jumping between spec points, wdyt @lawrence-forooghian? - https://github.com/ably/specification/pull/181

Regarding loading failure - it's not like unrecoverable failure, right? Why generate new details if you can try to load a bit later when user unlocks device?

lawrence-forooghian:
Will answer your other points @maratal, but re

> Regarding loading failure - it's not like unrecoverable failure, right? Why generate new details if you can try to load a bit later when user unlocks device?

It's an option that I'd considered but when I discussed with Paddy he said he didn't think we should do this and instead should just make sure it's stored in a way that's always available. Bear in mind also that in ably-cocoa `ARTRealtime.device()` loads the device synchronously and is expected to return a fully populated `LocalDevice`.

lawrence-forooghian:
BTW I've added 5ccc1d129d2e37d48b2eaabd376c0a114fd92d7d to remove a statement I didn't realise existed which seems quite redundant

lawrence-forooghian:
> There is a lot of back and forth between RSH8a, RSH8b, RSH3h, RSH8k2, RSH3a2b which makes it very hard to follow. Can we keep such a thing as "load details or generate if it doesn't exist or loading failed" in at least two places?

What are some examples of jumping-around that you think we could remove?

- RSH3h is the authoritative spec point for when to load the push activation state machine data
- RSH8a is the authoritative spec point for when to load the `LocalDevice` data, and its child RSH8a1 is the authoritative spec point for what to do if this loading fails

> Earlier this year I've tried to fit specs with current cocoa behavior and reducing amount of jumping between spec points, wdyt @lawrence-forooghian? - https://github.com/ably/specification/pull/181

It seems like that PR is largely orthogonal to the current one; that's about when the device ID and secret should be regenerated. But if you think that merging that one would allow us to simplify this one, I'm happy to hear how. My worry about merging that one is, as described in 9871e20fe304bb09fe35ee922e469ee0824cce46, we'd be replacing a spec point that describes how it's currently implemented in ably-java with one that describes how it's currently implemented in ably-cocoa and ably-js; I'd prefer that if we're going to do that then we also align the implementations.

maratal:
> BTW I've added [5ccc1d1](https://github.com/ably/specification/commit/5ccc1d129d2e37d48b2eaabd376c0a114fd92d7d) to remove a statement I didn't realise existed which seems quite redundant

I bet you mean 64fdb7f1e50f1dc8a952108095541a7ea68ccb05

lawrence-forooghian:
> I bet you mean [64fdb7f](https://github.com/ably/specification/commit/64fdb7f1e50f1dc8a952108095541a7ea68ccb05)

yes, sorry

specifications/features.md
```diff
@@ -1169,8 +1170,9 @@ The core SDK provides an API for wrapper SDKs to supply Ably with analytics info
...
...eing non-nullable is not actually consistent with [`RSH3a2b`](#RSH3a2b); that spec point implies that `id` and `deviceSecret` both start off unset and are only set upon a `CalledActivate` event. However, since `deviceSecret` needs to have the same nullability as `id` --- since per [`RSH3a2b`](#RSH3a2b) either both or neither should be set --- to reflect the behaviour described in the spec we would have to make `LocalDevice#id` nullable, but this is incompatible with the superclass `DeviceDetails`. In reality, ably-cocoa and ably-js actually generate `id` and `deviceSecret` when the device is fetched, i.e. not following [`RSH3a2b`](#RSH3a2b). ably-java follows [`RSH3a2b`](#RSH3a2b), allowing `id` to be `null` (this doesn't affect the public API because nullability is not a concept in Java's type system). What we *should* do is either make `LocalDevice` stop inheriting from `DeviceDetails`, or change the specified behaviour for when to generate `id` and `deviceSecret` to match our implementations, or both. See spec issues [#180](https://github.com/ably/specification/issues/180) and [#25](https://github.com/ably/specification/issues/25). For now, this note exists to reduce confusion.)
```
maratal, line 1173: Isn't this PR a good moment to actually specify the behaviour for when to generate `id` and `deviceSecret`? Why keep this huge note. WDYT @lawrence-forooghian

lawrence-forooghian: My preference is to not increase the scope of the spec changes any more than is needed to fix the issue we're trying to fix. Do you feel strongly we should be changing this here and now?

maratal (approved):
LGTM

PR #478, opened 2026-05-19
Clarify the default values for newly created LiveMap and LiveCounter

sacOO7:
- Based on https://github.com/ably/specification/pull/361
- Remove the term "zero-value" object as it is an unnecessary detail that simply describes a newly created object according to RTLM4 and RTLC4

### DIFF
--- a/specifications/objects-features.md
+++ b/specifications/objects-features.md
@@ -140,14 +140,14 @@ Objects feature enables clients to store shared data as "objects" on a channel. 
 - `(RTO3)` An internal `ObjectsPool` should be used to maintain the list of objects present on a channel
   - `(RTO3a)` `ObjectsPool` is a `Dict<String, LiveObject>` - a map of `LiveObject`s keyed by [`objectId`](../features#OST2a) string
   - `(RTO3b)` It must always contain a `LiveMap` object with id `root`
-    - `(RTO3b1)` Upon initialization of the `ObjectsPool`, create a new `LiveMap` (per [RTLM4](#RTLM4)) with `objectId` set to `root` and add it to the `ObjectsPool`
+    - `(RTO3b1)` Upon initialization of the `ObjectsPool`, create a new `LiveMap` per [RTLM4](#RTLM4) with `objectId` set to `root` and add it to the `ObjectsPool`
 - `(RTO4)` When a channel `ATTACHED` `ProtocolMessage` is received, the client library must perform the following actions in order. The `ProtocolMessage` may contain a `HAS_OBJECTS` bit flag (see [TR3](../features#TR3)); note that some of the following actions are conditional on this flag.
   - `(RTO4c)` The [RTO17](#RTO17) sync state must transition to `SYNCING` if not already `SYNCING`
   - `(RTO4d)` The `bufferedObjectOperations` list must be cleared without applying any buffered operations
   - `(RTO4a)` If the `HAS_OBJECTS` flag is 1, the server will shortly perform an `OBJECT_SYNC` sequence as described in [RTO5](#RTO5). Note that this does not imply that objects are definitely present on the channel, only that there may be; the `OBJECT_SYNC` message may be empty
   - `(RTO4b)` If the `HAS_OBJECTS` flag is 0 or there is no `flags` field, the sync sequence must be considered complete immediately, and the client library must perform the following actions in order:
     - `(RTO4b1)` All objects except the one with id `root` must be removed from the internal `ObjectsPool`
-    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be cleared by setting it to a zero-value per [RTLM4](#RTLM4). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
+    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be set to the value described in [RTLM4c](#RTLM4c). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
       - `(RTO4b2a)` Emit a `LiveMapUpdate` object for the `LiveMap` with ID `root`, with `LiveMapUpdate.update` consisting of entries for the keys that were removed, each set to `removed`
     - `(RTO4b3)` The `SyncObjectsPool` must be cleared
     - `(RTO4b5)` This clause has been replaced by [RTO4d](#RTO4d)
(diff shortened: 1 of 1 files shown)


sacOO7:
I had a question
1. We already have `zero-value` references in the codebase ( both ably-js and ably-java AFAIK), so do we want to keep existing terminology or adapt to new one?
2. If we adapt to new one ( `default value` or a `new object` ), do we also need to update relevant references in all codebases. If we keep implementation spec-agnostic, then it will differ across SDKs for `zero-value` objects
3. Maybe replacing `zero-value object` with `empty object` would make more sense?

sacOO7:
> I had a question
> 
> 1. We already have `zero-value` references in the codebase ( both ably-js and ably-java AFAIK), so do we want to keep existing terminology or adapt to new one?
> 2. If we adapt to new one ( `default value` or a `new object` ), do we also need to update relevant references in all codebases. If we keep implementation spec-agnostic, then it will differ across SDKs for `zero-value` objects
> 3. Maybe replacing `zero-value object` with `empty object` would make more sense?

Validated [all public facing liveobject docs](https://ably.com/docs/liveobjects/) for `zero-value`, `zero value` terminology 

We found =>

The "zero-value" terminology is purely a spec-internal artefact that was never surfaced in public-facing documentation. The docs already use the user-friendly phrasings ("empty map", "initial value 0", "newly created object", "initial state") that the spec PR is migrating toward, so the public surface is already coherent with the new wording.

[ZERO_VALUE_LIVEOBJECTS_REVIEW.md](https://github.com/user-attachments/files/28016092/ZERO_VALUE_LIVEOBJECTS_REVIEW.md)

sacOO7:
Updated PR as per discussion here -> https://github.com/ably/specification/pull/478/commits/286024eac8beeee6feb772f593fb870ee93cd373

specifications/objects-features.md
```diff
@@ -140,14 +140,14 @@ Objects feature enables clients to store shared data as "objects" on a channel.
...
   - `(RTO4a)` If the `HAS_OBJECTS` flag is 1, the server will shortly perform an `OBJECT_SYNC` sequence as described in [RTO5](#RTO5). Note that this does not imply that objects are definitely present on the channel, only that there may be; the `OBJECT_SYNC` message may be empty
   - `(RTO4b)` If the `HAS_OBJECTS` flag is 0 or there is no `flags` field, the sync sequence must be considered complete immediately, and the client library must perform the following actions in order:
     - `(RTO4b1)` All objects except the one with id `root` must be removed from the internal `ObjectsPool`
-    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be cleared by setting it to a zero-value per [RTLM4](#RTLM4). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
+    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be cleared by setting it to an empty map per [RTLM4c](#RTLM4c). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
```
lawrence-forooghian:
```suggestion
    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be cleared by setting it to the value described in [RTLM4c](#RTLM4c). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
```

specifications/objects-features.md
```diff
@@ -350,7 +350,7 @@ Objects feature enables clients to store shared data as "objects" on a channel.
       - `(RTLO4e3a)` This clause has been replaced by [RTLO6a](#RTLO6a)
       - `(RTLO4e3b)` This clause has been replaced by [RTLO6b](#RTLO6b)
         - `(RTLO4e3b1)` This clause has been replaced by [RTLO6b1](#RTLO6b1)
-    - `(RTLO4e4)` Set the data for the `LiveObject` to a zero-value, as described in [RTLC4](#RTLC4) or [RTLM4](#RTLM4) depending on the object type
+    - `(RTLO4e4)` Clear the internal data of the `LiveObject` as described in [RTLC4b](#RTLC4b) or [RTLM4c](#RTLM4c), depending on the object type
```
lawrence-forooghian:
```suggestion
    - `(RTLO4e4)` Clear the internal data of the `LiveObject` by setting it to the value described in [RTLC4b](#RTLC4b) or [RTLM4c](#RTLM4c), depending on the object type
```

lawrence-forooghian:
@sacOO7 just a couple of minor comments, and please could you squash commits before merge?

specifications/objects-features.md
sacOO7: We don't really set any value for either `LiveMap` or `LiveCounter`, don't you think it's a bit misleading to say `setting it to the value` wdyt?

specifications/objects-features.md
sacOO7: We don't really set any value as per `RTLM4c`. Being explicit, we clear/get rid of all values from the map.
Do you think wording gets misleading here?

specifications/objects-features.md
lawrence-forooghian: What do you mean "we don't really set any value" — isn't this spec point saying that you should set the LiveObject's `data` property to a specific value?

lawrence-forooghian: It could even just lose the word "clear" and say

```
    - `(RTLO4e4)` Set the `data` attribute of the `LiveObject` to the value described in [RTLC4b](#RTLC4b) or [RTLM4c](#RTLM4c), depending on the object type
```

specifications/objects-features.md
lawrence-forooghian: RTLM4c describes a particular value for `data` (i.e. an empty map); we're saying to set it to that value

lawrence-forooghian: again as mentioned in the other comment we can even lose the word "clear"

sacOO7: Addressed, removed `cleared` keyword https://github.com/ably/specification/pull/478/changes/d0de4cab91601e3210be9b8a484ec904494a4ea5

specifications/objects-features.md
sacOO7: Updated https://github.com/ably/specification/pull/478/changes/d0de4cab91601e3210be9b8a484ec904494a4ea5

sacOO7:
> @sacOO7 just a couple of minor comments, and please could you squash commits before merge?

Fixed https://github.com/ably/specification/pull/478/changes/d0de4cab91601e3210be9b8a484ec904494a4ea5

specifications/objects-features.md
```diff
@@ -140,14 +140,14 @@ Objects feature enables clients to store shared data as "objects" on a channel.
...
   - `(RTO4a)` If the `HAS_OBJECTS` flag is 1, the server will shortly perform an `OBJECT_SYNC` sequence as described in [RTO5](#RTO5). Note that this does not imply that objects are definitely present on the channel, only that there may be; the `OBJECT_SYNC` message may be empty
   - `(RTO4b)` If the `HAS_OBJECTS` flag is 0 or there is no `flags` field, the sync sequence must be considered complete immediately, and the client library must perform the following actions in order:
     - `(RTO4b1)` All objects except the one with id `root` must be removed from the internal `ObjectsPool`
-    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be cleared by setting it to a zero-value per [RTLM4](#RTLM4). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
+    - `(RTO4b2)` The data for the `LiveMap` with id `root` must set to an empty map per [RTLM4c](#RTLM4c). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
```
lawrence-forooghian:
```suggestion
    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be set to a new map created per [RTLM4c](#RTLM4c). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
```

lawrence-forooghian: Sorry, that first suggestion was wrong, should be 
```suggestion
    - `(RTO4b2)` The data for the `LiveMap` with id `root` must be set to the value described in [RTLM4c](#RTLM4c). Note that the client SDK must not create a new `LiveMap` instance with id `root`; it must only clear the internal data of the existing `LiveMap` with id `root`
```

lawrence-forooghian (approved):

PR #486, opened 2026-06-01
Clarify `clearTimeserial` must be reset when `LiveMap` is reset

lawrence-forooghian:
When resetting an existing `LiveMap`'s data you are meant to also reset its `clearTimeserial` (as JS / Swift / Kotlin implementations all do). This was never very clear in the spec and 4cec279 made it even less clear. So, define an explicit "reset data for an existing `LiveObject`" procedure and defer to it.


specifications/objects-features.md
```diff
@@ -418,6 +418,8 @@ Objects feature enables clients to store shared data as "objects" on a channel.
 - `(RTLC4)` A new empty `LiveCounter` can be created with the following values:
   - `(RTLC4a)` `objectId` is passed into the constructor and set upon creation
   - `(RTLC4b)` `data` is set to 0
+- `(RTLC17)` The data of a `LiveCounter` can be reset to that of a new empty `LiveCounter` as follows:
```
paddybyers, line 421: I don't think reset of a LiveCounter is a thing

lawrence-forooghian: The spec already (i.e. pre-this-PR) resets a LiveCounter's data as part of the RTLO4e tombstone procedure

paddybyers: ok, why? Unlike a map, there's no space saving if you reset a LiveCounter. Is it just for consistency then?

lawrence-forooghian: I'm not sure to be honest — it's what ably-js does. Possible reasons would be consistency or a design decision that if a user has a reference to a deleted `LiveCounter` then when they call `value()` on it it should return 0, but I'm just speculating.

paddybyers: @VeskeR ?

PR #488, opened 2026-06-08
Move resumability decisions from the client to the server

SimonWoolf:
Implements https://ably.atlassian.net/wiki/spaces/CHN/pages/5078777859/PSRFC-32+move+resumability+decisions+to+the+server+not+the+sdk

Motivation, in short:
- we can get rid of ATTACH_RESUME, which is weird and confusing
- it gives us the freedom to experiment with providing longer resume windows, streaming resumes from new-model channel history, with purely serverside logic
- it allows us to start decoupling “how long we can resume subscriptions for” from “how long we can keep a message queued for publishing”, there's no fundamental reason those two should be tied together

(also tweaked docstrings for connection states, which were just the full wording from the docs copied and pasted, and were way too verbose for docstrings)


specifications/api-docstrings.md
```diff
@@ -474,7 +474,7 @@ Describes the possible states of a [`RestChannel`]{@link RestChannel} or [`Realt
 | ATTACHED || The attach has succeeded. In the `ATTACHED` state a client may publish and subscribe to messages, or be present on the channel. |
 | DETACHING || A detach has been initiated on an `ATTACHED` channel by sending a request to Ably. This is a transient state, followed either by a transition to `DETACHED` or `FAILED`. |
 | DETACHED || The channel, having previously been `ATTACHED`, has been detached by the user. |
-| SUSPENDED || The channel, having previously been `ATTACHED`, has lost continuity, usually due to the client being disconnected from Ably for longer than two minutes. It will automatically attempt to reattach as soon as connectivity is restored. |
+| SUSPENDED || The channel, having previously been `ATTACHED`, is waiting until the client to reconnect to Ably after an extended period of disconnection. It will automatically reattach once connectivity is restored. Message continuity may have been lost, but the authoritative source for this is the [`resumed`]{@link ChannelStateChange#resumed}) flag on the subsequent `attached` state change. |
```
lmars:
```suggestion
| SUSPENDED || The channel, having previously been `ATTACHED`, is waiting for the client to reconnect to Ably after an extended period of disconnection. It will automatically reattach once connectivity is restored. Message continuity may have been lost, but the authoritative source for this is the [`resumed`]{@link ChannelStateChange#resumed}) flag on the subsequent `attached` state change. |
```

specifications/api-docstrings.md
```diff
@@ -977,8 +977,8 @@ Describes the realtime [`Connection`]{@link Connection} object states.
...
+| DISCONNECTED || A temporary failure condition. The client has been disconnect from Ably, usually due to network connectivity. It will periodically attempt to open a new connection (about every 15 seconds). In this state, you can continue to publish messages, which will be queued to be sent once a connection is reestablished. Messages published by other clients while this client is disconnected will be delivered to it upon reconnection if the server is able to preserve continuity. |
```
lmars:
```suggestion
| DISCONNECTED || A temporary failure condition. The client has been disconnected from Ably, usually due to network connectivity. It will periodically attempt to open a new connection (about every 15 seconds). In this state, you can continue to publish messages, which will be queued to be sent once a connection is reestablished. Messages published by other clients while this client is disconnected will be delivered to it upon reconnection if the server is able to preserve continuity. |
```

specifications/api-docstrings.md
```diff
@@ -977,8 +977,8 @@ Describes the realtime [`Connection`]{@link Connection} object states.
...
+| SUSPENDED || A long term failure condition. A client moves to this state once it has been in the `disconnected` state for over two minutes. Developers are unable to publish messages in this state, and any queued messages are discarded. The library will attempt to reconnect every 30 seconds; an attempt can also be triggered by calling [`connect()`]{@link Connection#connect}. |
```
lmars, line 982: The above edit says "The client has been disconnected..." and "you can continue to publish messages", so should this here say "The client has been in the `disconnected` state for over two minutes" and "you are unable to publish messages in this state"?

(I'll add that I'm not sure who "you" is here, we're talking about the client being disconnected, so is it not also the client that can or cannot publish messages?)

lmars (approved):
Some wording related comments, but otherwise LGTM.

specifications/api-docstrings.md
SimonWoolf: 'developers are' was there in the previous version, I just missed updating it to better terminology ('you'), thanks.

> (I'll add that I'm not sure who "you" is here, we're talking about the client being disconnected, so is it not also the client that can or cannot publish messages?)

'you' as in is the person using the client library. So when in the `disconnected` state, the client library can't publish messages, but the person using the client library still can (in the sense that they can call publish() and the messages will be queued). when in `suspended`, not so

PR #491, opened 2026-06-09
Add typed-SDK LiveObjects API spec section (RTTS1-RTTS10)

sacOO7:
- Updated spec for cross SDK strongly typed languages.
- Fixes https://ably.atlassian.net/browse/AIT-875

## Summary

- Introduce a new `### Typed-SDK public API (statically-typed languages)` section in `objects-features.md`             
(RTTS1–RTTS10) that normatively partitions `PathObject` and `Instance` across a class hierarchy for SDKs whose host
language is strongly typed (Java, Swift, Kotlin). The dynamic-SDK form (JavaScript) is unchanged; the new section is 
purely a refinement of the existing `RTPO1a` / `RTINS1a` allowance for type-specific sub-classes.                  
- Add 6 minimal inbound cross-refs (`RTPO1b`, `RTINS1b`, `RTPO13g`, `RTINS10c`, `RTO23f`, `RTINS11c`) so readers of
any existing `RTPO*` / `RTINS*` clause find the typed-SDK partition. Tighten `RTINS11c` so the non-null `compactJson`
 invariant on `Instance` applies universally (Instance is always bound — the existing JS behaviour already satisfies
this).                                                                                                               
- Append a typed-SDK IDL block to the `Interface Definition` section, alongside (not replacing) the dynamic-SDK form,
 with a banner that explains the relationship.
- Add a `# Typed-SDK LiveObjects API` section to `api-docstrings.md` with language-agnostic docstrings for every new
type, each citing its `RTTS*` clause.                                                                                
 
## Key normative decisions                                                                                           
                                                                                                                   
- **`subscribe()` restructure (RTTS7b).** `subscribe` is removed from the base typed `Instance` and lives only on    
`LiveMapInstance` / `LiveCounterInstance`. This turns the existing runtime throw of `RTINS16c` on primitives into a
compile-time error and matches the official docs which state that `subscribe()` is "only available on `LiveMap` and  
`LiveCounter` instances".                                                                                          
- **`SubscriptionEvent` payload (RTTS3d, RTTS10a, RTTS10b).** Typed SDKs MUST deliver both `object` and `message`
fields per `RTPO19e` / `RTINS16e` — a subscription-event type that omits `message` is non-conformant.                
- **`getType()` semantics (RTTS4b, RTTS8a).** `PathObject#getType()` returns `UNKNOWN` on unresolved paths (silent,
consistent with `RTPO3c1`); `Instance#getType` never returns `UNKNOWN` in normal operation (Instance is bound).      
- **`compact` deferred (RTTS3f / RTTS7d).** Typed SDKs are not required to implement `compact`; `compactJson` is   
sufficient.                                                                                                          
- **`channel.objects.get()` returns `LiveMapPathObject` (RTTS6d / RTO23f).**                                       
- **Universal `Instance#compactJson` non-null invariant (`RTINS11c`).** One-line tightening of the dynamic spec;     
reflects a property that already held in practice.


specifications/objects-features.md
```diff
@@ -1078,6 +1085,82 @@ An `Instance` holds a direct reference to a specific resolved `LiveObject` or pr
...
+  - `(RTTS5a)` `asLiveMap()` returns a `LiveMapPathObject` ([RTTS6a](#RTTS6a))
+  - `(RTTS5b)` `asLiveCounter()` returns a `LiveCounterPathObject` ([RTTS6b](#RTTS6b))
+  - `(RTTS5c)` `asNumber()`, `asString()`, `asBoolean()`, `asBinary()`, `asJsonObject()`, `asJsonArray()` each return the corresponding primitive sub-class per [RTTS6c](#RTTS6c)
+  - `(RTTS5d)` These helpers must not throw based on the current resolved type; they only re-wrap. Operations on the returned wrapper carry their own failure semantics:
+    - `(RTTS5d1)` Read operations (`value`, `instance`, `entries`, `keys`, `values`, `size`, `compactJson`) follow [RTPO3c1](#RTPO3c1) - return null/empty if the resolved type does not match the wrapper's expectation
+    - `(RTTS5d2)` Write and terminal operations (`set`, `remove`, `increment`, `decrement`) follow [RTPO3c2](#RTPO3c2) - throw an `ErrorInfo` with `statusCode` 400 and `code` 92007 if the resolved type does not match
```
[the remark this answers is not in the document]
sacOO7: Good catch — fixed in https://github.com/ably/specification/commit/7592b8e13fb6be1199a591bfbcb6a9b8cd710319. RTTS5d2 now distinguishes the two failure modes: `92005` when the path does not resolve (per RTPO3c2) and `92007` when it resolves to the wrong type (per RTPO15e/RTPO16e for set/remove and RTPO17e/RTPO18e for increment/decrement), instead of citing RTPO3c2 for the 92007 case.

specifications/objects-features.md
```diff
@@ -1078,6 +1085,82 @@ An `Instance` holds a direct reference to a specific resolved `LiveObject` or pr
...
+  - `(RTTS6b)` `LiveCounterPathObject` - adds `increment` per [RTPO17](#RTPO17), `decrement` per [RTPO18](#RTPO18), and a `value()` method delegating to [RTPO7](#RTPO7) with the return type narrowed to `Number?` (returns null per [RTPO3c1](#RTPO3c1) / [RTPO7e](#RTPO7e) if the resolved value is not an `InternalLiveCounter`). SDKs may narrow this further to the host's idiomatic floating-point type (e.g. `Double?`), since a counter's value is always a finite floating-point number per [RTLCV4a](#RTLCV4a)
+  - `(RTTS6c)` Primitive `PathObject` sub-classes - one per primitive: `NumberPathObject`, `StringPathObject`, `BooleanPathObject`, `BinaryPathObject`, `JsonObjectPathObject`, `JsonArrayPathObject`. Each adds only a `value()` method delegating to [RTPO7](#RTPO7) with the return type narrowed to its corresponding primitive (returning null if the resolved value is not of that primitive type, per [RTPO3c1](#RTPO3c1)). They do not expose `get`, `at`, or any write / iteration methods
```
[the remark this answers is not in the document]
sacOO7: Agreed — fixed in https://github.com/ably/specification/commit/7592b8e13fb6be1199a591bfbcb6a9b8cd710319. RTTS6b and RTTS6c now describe filtered `value()` semantics (return the value only when the resolved value is of the expected category, `null` otherwise — e.g. `NumberPathObject#value()` never returns an `InternalLiveCounter`'s value) rather than plain delegation to RTPO7, which also returns counter/primitive values.

specifications/objects-features.md
```diff
@@ -1078,6 +1085,82 @@ An `Instance` holds a direct reference to a specific resolved `LiveObject` or pr
   - `(RTINS16g)` The subscription is identity-based: it follows the specific `LiveObject` instance, regardless of where it sits in the graph
   - `(RTINS16h)` This operation must not have any side effects on `RealtimeObject`, the underlying channel, or their status
 
+### Typed-SDK public API (statically-typed languages)
```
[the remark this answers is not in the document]
sacOO7: Thanks — the spec heading ("statically-typed languages") is the intended wording and is consistent throughout RTTS1/RTTS3; the PR description was the outlier. I've updated the PR description to match, so no spec change was needed here.

specifications/objects-features.md
```diff
@@ -1261,9 +1348,119 @@ Types and their properties/methods are public and exposed to users by default. A
...
       increment(Number amount?) => io // RTINS14
       decrement(Number amount?) => io // RTINS15
       subscribe((InstanceSubscriptionEvent) -> listener) -> Subscription // RTINS16
+
+    // === Typed-SDK variant (RTTS1-RTTS10) =====================================
+    // Statically-typed SDKs (Java/Swift/Kotlin) expose the hierarchy below in
+    // place of the monomorphic `class PathObject` / `class Instance` above.
+    // ==========================================================================
+
+    enum ValueType: // RTTS2
```
lawrence-forooghian, line 1376: I know we discussed this name plenty already, but I don't remember whether we discussed `TypeOfValue` as an option? The term "value type" already has a meaning and I think especially in Swift it might be a bit confusing.

specifications/objects-features.md
```diff
@@ -1261,9 +1348,119 @@ Types and their properties/methods are public and exposed to users by default. A
...
+      asLiveCounter: LiveCounterInstance // RTTS9b
+      asNumber: NumberInstance // RTTS9c
+      asString: StringInstance // RTTS9c
+      asBoolean: BooleanInstance // RTTS9c
+      asBinary: BinaryInstance // RTTS9c
+      asJsonObject: JsonObjectInstance // RTTS9c
+      asJsonArray: JsonArrayInstance // RTTS9c
+
+    class LiveMapInstance extends Instance: // RTTS10a
+      id: String // RTINS3a (narrowed to non-nullable)
```
lawrence-forooghian: Narrowed when compared to what? `Instance`? `Instance` doesn't have an `id` property in the IDL; if that's intentional then "narrowed" isn't the term here

specifications/objects-features.md
```diff
@@ -1261,9 +1348,119 @@ Types and their properties/methods are public and exposed to users by default. A
...
+      compactJson() -> JsonObject // RTTS7a1 (narrowed)
+      get(String key) -> Instance? // RTINS5
+      entries() -> [String, Instance][] // RTINS6
+      keys() -> String[] // RTINS7
+      values() -> Instance[] // RTINS8
+      size() -> Number // RTINS9 (narrowed to non-nullable)
```
lawrence-forooghian: Narrowed when compared to what? `Instance`? `Instance` doesn't have an `id` property in the IDL; if that's intentional then "narrowed" isn't the term here

specifications/objects-features.md
```diff
@@ -1261,9 +1348,119 @@ Types and their properties/methods are public and exposed to users by default. A
...
+      set(String key, (Boolean | Binary | Number | String | JsonArray | JsonObject | LiveCounter | LiveMap) value) => io // RTINS12
+      remove(String key) => io // RTINS13
+      subscribe((InstanceSubscriptionEvent) -> listener) -> Subscription // RTINS16, RTTS7b
+
+    class LiveCounterInstance extends Instance: // RTTS10b
+      id: String // RTINS3a (narrowed to non-nullable)
```
lawrence-forooghian: Same q re narrowed from what

specifications/objects-features.md
```diff
@@ -1261,9 +1348,119 @@ Types and their properties/methods are public and exposed to users by default. A
...
+      compactJson() -> JsonPrimitive // RTTS7a3 (narrowed)
+      value() -> Number // RTINS4 (narrowed to non-nullable)
```
lawrence-forooghian: Same q re narrowed compared to what

specifications/objects-features.md
```diff
@@ -1078,6 +1085,86 @@ An `Instance` holds a direct reference to a specific resolved `LiveObject` or pr
...
+  - `(RTTS8a)` `getType` - returns the `ValueType` ([RTTS2](#RTTS2)) of the wrapped value. Returns the matching `ValueType` member per [RTTS2a](#RTTS2a). Must not return `UNKNOWN` in normal operation because an `Instance` is always constructed from a resolved value (see e.g. [RTPO8c](#RTPO8c), [RTINS5c](#RTINS5c)); see [RTTS2a9](#RTTS2a9) for the future-compatibility caveat
+  - `(RTTS8b)` There is no `exists` helper on `Instance`, because an `Instance` is only ever constructed when it wraps a resolved value; the existence question is unambiguous
+- `(RTTS9)` Typed `Instance` `as*` cast helpers - type-checked accessors. Unlike the best-effort `PathObject` casts ([RTTS5](#RTTS5)), an `Instance` wraps an already-resolved value of a known, fixed type ([RTTS8a](#RTTS8a)), so a mismatched cast is a programming error and these helpers throw rather than re-wrap ([RTTS9d](#RTTS9d)). Exposed as properties in host languages that distinguish properties from methods, since the wrapped value is already resolved
```
lawrence-forooghian: Not all languages have the concept of a throwing property, I think — Swift didn't until recently. In that case they'll have to use a method

specifications/objects-features.md
```diff
@@ -1078,6 +1085,86 @@ An `Instance` holds a direct reference to a specific resolved `LiveObject` or pr
...
+- `(RTTS3)` Typed `PathObject` base class - the typed SDK's public root of the `PathObject` hierarchy. The base class exposes only those methods of [RTPO1](#RTPO1) whose return type and semantics are identical regardless of the resolved underlying type
+  - `(RTTS3a)` Exposes `path` per [RTPO4](#RTPO4)
+  - `(RTTS3b)` Exposes `instance` per [RTPO8](#RTPO8). On sub-classes other than `LiveMapPathObject` ([RTTS6a](#RTTS6a)) and `LiveCounterPathObject` ([RTTS6b](#RTTS6b)), the behaviour of `instance` is constrained by [RTTS6e](#RTTS6e)
+  - `(RTTS3c)` Exposes `compactJson` per [RTPO14](#RTPO14), with the SDK's generic, nullable JSON element type as its return type. Unlike the typed `Instance` sub-classes ([RTTS7a](#RTTS7a)), the typed `PathObject` sub-classes do not narrow this return type: a `PathObject` cast is best-effort and `compactJson` may resolve to a value of any type, or fail to resolve and return null (per [RTPO3c1](#RTPO3c1)), so no narrower type can be guaranteed for any sub-class
```
lawrence-forooghian: I thought we'd previously said we wouldn't use the word "cast" for anything related to `PathObject` since it's not really doing a cast? And "best-effort" isn't really correct either; who is making a best effort to do what exactly? It's just a view that represents the type of value that a user _expects_ to find at a given path

specifications/objects-features.md
sacOO7: Yeah, will remove `narrowed` from everywhere, for `Instance` it just exists on sub-types

specifications/objects-features.md
sacOO7: Fixed in c5cc6b8a — dropped `narrowed`; the IDL comment now reads `(non-nullable; id exists only on LiveObject sub-types)`, and the RTTS10a prose states the plain return type.

specifications/objects-features.md
sacOO7: Done in c5cc6b8a — `narrowed` removed everywhere it described members added on sub-types (`id`/`size`/`value`). It is kept only for `compactJson`, where it is genuine covariant return-type narrowing relative to the base `Instance#compactJson` (RTTS7a).

specifications/objects-features.md
sacOO7: Fixed in c5cc6b8a — same rewording as the `LiveMapInstance` `id` line.

specifications/objects-features.md
sacOO7: Fixed in c5cc6b8a — now `(non-nullable)`, with the RTTS10b prose stating the plain `Number` return type.

specifications/objects-features.md
sacOO7: Agreed — reworded in c5cc6b8a. The `as*` helpers are now **view helpers** throughout, which also matches the ably-java Javadoc (it consistently calls what they return "a ... view of this path/instance"). `PathObject` views are described as *unchecked* accessors "that return a sub-class view representing the type the caller expects at this path, without performing any validation", and `Instance` views as *checked* accessors that throw on mismatch. `best-effort` is gone from the `as*` prose ("unchecked" carries the meaning), and the `exists`/`getType` path helpers are now described as *point-in-time* instead.

specifications/objects-features.md
sacOO7: Good catch — addressed in c5cc6b8a. RTTS9 now reads: "Exposed as properties in host languages that distinguish properties from methods **and support throwing property accessors**; host languages without throwing property accessors may expose these as methods instead".

specifications/objects-features.md
sacOO7: Fair point — `TypeOfValue` (or similar) would avoid the "value type" collision, especially in Swift. Treating it as a separate naming decision since renaming touches RTTS2, the IDL, `ValueType.java` and the docs; will follow up on it independently rather than bundling it into this change.

PR #492, opened 2026-06-09
ait: add liveobjects passthrough specs

owenpearson:
Needed for [AIT-838](https://ably.atlassian.net/browse/AIT-838)

[AIT-838]: https://ably.atlassian.net/browse/AIT-838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/specifications/ai-transport-features.md
+++ b/specifications/ai-transport-features.md
@@ -134,6 +134,11 @@ The agent (server-side) session manages the run lifecycle over an Ably channel. 
 
 - `(AIT-ST14)` The agent session must expose the underlying Ably channel's presence object via a read-only `presence` accessor. The accessor must return the channel's `RealtimePresence` directly, applying no AI-Transport-specific semantics, and must be usable before `connect()` is called.
 
+### LiveObjects
+
+- `(AIT-ST15)` The agent session must expose the underlying Ably channel's LiveObjects entry point via a read-only `object` accessor. The accessor must return the channel's `RealtimeObject` directly, applying no AI-Transport-specific semantics, and must be usable before `connect()` is called. Operating on it requires the Realtime client to have been constructed with the LiveObjects plugin and the channel to have been attached with the object channel modes (AIT-ST16); when either is absent the underlying SDK throws and the session must not suppress the error.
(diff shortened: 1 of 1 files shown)


specifications/ai-transport-features.md
```diff
@@ -134,6 +134,11 @@ The agent (server-side) session manages the run lifecycle over an Ably channel.
...
 
+### LiveObjects
+
+- `(AIT-ST15)` The agent session must expose the underlying Ably channel's LiveObjects entry point via a read-only `object` accessor. The accessor must return the channel's `RealtimeObject` directly, applying no AI-Transport-specific semantics, and must be usable before `connect()` is called. Operating on it requires the Realtime client to have been constructed with the LiveObjects plugin and the channel to have been attached with the object channel modes (AIT-ST16); when either is absent the underlying SDK throws and the session must not suppress the error.
+- `(AIT-ST16)` The agent session must accept an optional `channelModes` option listing additional Ably channel modes to request on the session's channel (e.g. the object modes required by AIT-ST15). When provided, the session must request the union of these modes with the modes the SDK itself requires, deduplicated and in a deterministic order. The SDK's required base modes must equal the channel's default mode set, so that requesting additional modes does not alter any other channel behaviour. When omitted, the session must not set channel modes, preserving the default attachment behaviour.
```
VeskeR: Which SDK do "the modes the SDK itself requires" and "the SDK's required base modes" refer to here - the AIT SDK or core pub/sub? I read it as the AIT SDK, but it would help to say so explicitly.

Neither the pub/sub SDK nor the AIT SDK sets any channel modes today, so "the channel's default mode set" isn't defined in code anywhere. As I understand it, it's the set granted when no modes are requested (`publish`, `subscribe`, `presence`, `presence_subscribe`). Since the implementation has to introduce this base list from scratch, can we spell it out in the spec?

The last sentence also reads as contradicting the union: "When omitted, the session must not set channel modes" sounds like we set no modes at all, even though we just defined a union of modes.

specifications/ai-transport-features.md
```diff
@@ -247,6 +252,11 @@ The client session manages the client-side conversation lifecycle over an Ably c
...
+- `(AIT-CT23)` The client session must accept an optional `channelModes` option listing additional Ably channel modes to request on the session's channel (e.g. the object modes required by AIT-CT22). When provided, the session must request the union of these modes with the modes the SDK itself requires, deduplicated and in a deterministic order, and must apply the same resolved modes to every resolution of the channel's options for the channel's lifetime — including any auxiliary consumers such as React channel providers — so that the modes cannot be silently reverted. The SDK's required base modes must equal the channel's default mode set, so that requesting additional modes does not alter any other channel behaviour. When omitted, the session must not set channel modes, preserving the default attachment behaviour.
```
VeskeR: ditto channel modes comment from above

specifications/ai-transport-features.md
owenpearson: yeah i'm talking here from the perspective of the client, so setting no modes just means not sending modes on the wire. You're right that this results in those default modes being granted by realtime, and that's what I meant by "the channel's default mode set"

owenpearson: I will go ahead and make it more explicit though 👍

owenpearson: and done

VeskeR: nice 👍

VeskeR (approved):

PR #494, opened 2026-06-17
Clarify that RSH3h isn't meant to re-initialise the device

lawrence-forooghian:
RSH8a says "The LocalDevice is initialised when first required" but it doesn't say "only when first required"; Marat pointed out that RSH3h could arguably be interpreted as meaning "perform this initialisation process again". (I'm on the fence and think that "initialisation" implies a one-off thing, but no harm in clarifying.)


maratal (approved):
> RSH8a says "The LocalDevice is initialised when first required" but it doesn't say "only when first required"; Marat pointed out that RSH3h could arguably be interpreted as meaning "perform this initialisation process again". (I'm on the fence and think that "initialisation" implies a one-off thing, but no harm in clarifying.)

I'm a bit confused by this wording, all I was trying to say is that the device can be (and actually is for identified clients) initialized elsewhere (not upon state machine init) and thus automatically violating RSH3h. So loosing this requirement would make things easier for the current implementations. The change itself is fine.

PR #495, opened 2026-06-18
Add ObjectData json type and UNKNOWN enum sentinels to objects spec

sacOO7:
## Summary

Extends the LiveObjects wire types in `features.md` with a new `json` value type for `ObjectData`, plus `UNKNOWN` forward-compatibility sentinels on two wire enums.

### `ObjectData.json` value field (`OD2g`)
- Adds a `json` field holding a JSON object or array leaf value in the object graph.
- Sent to / received from the server as a JSON-encoded string when using the JSON protocol.
- `OD2c`–`OD2f` updated so `json` is included in the set of mutually-exclusive value fields (`boolean`, `bytes`, `number`, `string`, `json`).
- Added to the IDL as `json: String?`.

### `ObjectData` size calculation (`OD3a`, `OD3g`)
- `OD3a` updated to include `json` in the summed properties.
- New `OD3g`: the size of a `json` property is the byte length of its JSON-encoded string representation.
- Matches both reference implementations (ably-js `_getObjectDataSize` → `dataSizeBytes(data.json)`; ably-java `ObjectData.size()` → `json.toString().byteSize`).

### `UNKNOWN` enum sentinels (`OOP2a`, `OMP2a`)
- Adds a client-side `UNKNOWN` value to both `ObjectOperationAction` and `ObjectsMapSemantics`.
- It has **no wire value** and is **never encoded or sent**.
- When decoding, an `action`/`semantics` value that doesn't map to a known member must decode to `UNKNOWN` rather than failing.
- This gives clients a defined way to detect operations/maps they don't understand (e.g. introduced by a newer protocol version) so an `ObjectOperation` with an `UNKNOWN` action can be safely ignored and not applied.
- Both reflected in the IDL.

## Validation
- Spec linter passes (`npm run lint`) — no duplicate spec IDs.


specifications/features.md
```diff
@@ -1311,6 +1311,7 @@ The core SDK provides an API for wrapper SDKs to supply Ably with analytics info
 
 - `(OOP1)` An `ObjectOperation` describes an operation to be applied to an object on a channel
 - `(OOP2)` `ObjectOperationAction` enum has the following values in order from zero: `MAP_CREATE`, `MAP_SET`, `MAP_REMOVE`, `COUNTER_CREATE`, `COUNTER_INC`, `OBJECT_DELETE`, `MAP_CLEAR`
+  - `(OOP2a)` The enum additionally has an `UNKNOWN` value, which has no wire representation. When decoding, an `action` that does not map to any of the wire-ordered members above must be decoded as `UNKNOWN` rather than failing. The client library must never encode or send `UNKNOWN`, and must ignore (not apply) any `ObjectOperation` whose `action` is `UNKNOWN`
```
ttypic: I think it can be implemented differently (for example, as a nullable field). The main point is that if we receive an unknown value, we shouldn’t fail. Anyway, we agreed that we don’t want to add spec points that contradict the js implementation.

specifications/features.md
```diff
@@ -1512,16 +1514,18 @@ The core SDK provides an API for wrapper SDKs to supply Ably with analytics info
...
   - `(OD2b)` `encoding` string - may be set by the client library to indicate that value in `string` field have an encoding
-  - `(OD2c)` `boolean` boolean - a primitive boolean leaf value in the object graph. Only one of the value fields - `boolean`, `bytes`, `number` or `string` - must be set at a time
-  - `(OD2d)` `bytes` binary \| string - a primitive binary leaf value in the object graph. It is sent to and received from the server as a Base64-encoded string when using the JSON protocol. Only one of the value fields - `boolean`, `bytes`, `number` or `string` - must be set at a time
-  - `(OD2e)` `number` number - a primitive number leaf value in the object graph. Only one of the value fields - `boolean`, `bytes`, `number` or `string` - must be set at a time
-  - `(OD2f)` `string` string - a primitive string leaf value in the object graph. Only one of the value fields - `boolean`, `bytes`, `number` or `string` - must be set at a time
+  - `(OD2c)` `boolean` boolean - a primitive boolean leaf value in the object graph. Only one of the value fields - `boolean`, `bytes`, `number`, `string` or `json` - must be set at a time
```
ttypic: weird, don't see any changes to the (OD2c-f), if it's just trailing spaces could we revert them to keep the git history cleaner?

ttypic (changes requested):

specifications/features.md
sacOO7: Good call — reworded both OOP2a and OMP2a to specify only the behaviour (decode an unrecognised `action`/`semantics` without failing; never encode/send it; ignore ops with an unrecognised action) and left the representation to the implementation (sentinel value, nullable field, etc.). I've also dropped `UNKNOWN` from the wire-ordered enums in the IDL, since adding it there contradicted ably-js, where the numeric wire enum stays pure and `'unknown'` lives only on the decoded type. Done in 1230e2ca.

specifications/features.md
sacOO7: They weren't just trailing spaces — each line was adding `json` to the "only one set at a time" list. Rather than repeat that across all five value fields, I've hoisted the exclusivity rule into the parent OD2 clause and removed the per-line duplication, so OD2c–OD2g are now back to a single descriptive sentence each and future value fields won't churn every line. Done in 1230e2ca.

ttypic (approved):
LGTM

PR #500, opened 2026-07-01
feat: add `createApnsBroadcast` (RSH1d) and `liveActivity` methods in PushAdmin

ttypic:
- Include tests for `start`, `update`, and `end` methods with various scenarios and error propagation.
- Update related specifications and documentation to reflect test coverage and implementation.

### DIFF
--- a/uts/rest/unit/push/push_admin_apns_broadcast.md
+++ b/uts/rest/unit/push/push_admin_apns_broadcast.md
@@ -1,0 +1,204 @@
+# PushAdmin createApnsBroadcast Tests
+
+Spec points: `RSH1`, `RSH1d`
+
+## Test Type
+Unit test with mocked HTTP client
+
+## Mock HTTP Infrastructure
+
+See `uts/test/rest/unit/helpers/mock_http.md` for the full Mock HTTP Infrastructure specification.
+
+---
+
+## RSH1d — createApnsBroadcast sends POST to /push/apnsBroadcastChannels
+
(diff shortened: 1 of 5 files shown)


uts/rest/unit/push/push_admin_apns_broadcast.md
```diff
@@ -0,0 +1,204 @@
...
+```pseudo
+ASSERT captured_requests.length == 1
+
+body = parse_json(captured_requests[0].body)
+ASSERT body["messageStoragePolicy"] == 0
+```
+
+---
+
+## RSH1d — createApnsBroadcast returns id and apnsChannelId
```
paddybyers, line 97: Can we combine several of these tests? They're doing the same thing, just with different assertions

uts/rest/unit/push/push_admin_live_activity.md
```diff
@@ -0,0 +1,359 @@
...
+)
+```
+
+### Assertions
+```pseudo
+ASSERT captured_requests.length == 1
+
+request = captured_requests[0]
+ASSERT request.method == "POST"
+ASSERT request.url.path == "/push/apnsBroadcastChannels/broadcast-1/start"
```
paddybyers, line 56: Do we want this broadcast id to be randomised to prevent test runs from conflicting?

paddybyers (changes requested):
LGTM but a couple of comments. I think we could be grouping more assertions into a single test

Issue #10, opened 2022-10-02
Specify token revocation - `Auth#revokeTokens` method


QuintinWillison:
This spec addition was proposed by @ikbalkaya and then parked in https://github.com/ably/docs/pull/1426.

The features spec moved to this repository before that proposal was further advanced, therefore this issue represents the requirement to create a pull request in this repository for that proposal.

PR #501, opened 2026-07-04
[AIT-1106] fix(uts): correct objects proxy specs and harden proxy rule/timeout guidance

sacOO7:
- Fixed https://ably.atlassian.net/browse/AIT-1106

## Intent

While evaluating the derived ably-java test suite (`ObjectsFaultsTest`) against the real sandbox through `uts-proxy`, every test in `objects/integration/proxy/objects_faults.md` failed. Root-causing those failures surfaced one genuine spec error (RTO20e), several spec/tooling format mismatches that fail silently or with opaque errors, and a recurring translation trap around timeouts. This PR fixes the specs at source and adds the guidance so other deriving SDKs don't rediscover these the hard way.

Every behavioural claim below was cross-validated against the features spec (`objects-features.md`), ably-js source (the reference SDK), and a passing ably-java derived test run against the sandbox.

### 1. `objects/integration/proxy/objects_faults.md` — RTO20e spec error (the substantive fix)

The test `objects/proxy/RTO20e/publish-fails-on-channel-failed-0` mutated **after** the channel was already FAILED, yet asserted error **92008**:

- **RTO26b** requires a mutation on an already-DETACHED/FAILED/SUSPENDED channel to fail with **90001** "before doing anything else" — it never reaches `publishAndApply`.
- **RTO20e1**'s 92008 applies only to an operation **already waiting** for SYNCED when the channel transitions.
- ably-js confirms: `throwIfInvalidWriteApiConfiguration` → `invalidStateError()` = 90001/400; the only 92008 is the sync-wait rejection in `publishAndApply()`. Its own test ("publishAndApply rejects when channel state changes during sync wait") never mutates after FAILED. ably-java behaves identically.
- The test's own title ("enters FAILED **during SYNCING**") and the sibling unit test (`objects/unit/RTO20e1`) already describe the correct sequence — the proxy steps were the outlier.

The steps are restructured to the RTO20e1 sequence: force SYNCING via an injected ATTACHED carrying HAS_OBJECTS (`flags: 128`, RTO4c), mutate while SYNCING, wait for the publish ACK in the proxy log plus a brief real-time yield, inject the channel ERROR, and assert the pending operation fails with 92008 / statusCode 400 / cause 90000. A note explains the RTO26b-vs-RTO20e1 distinction so the inversion doesn't creep back in. The requirement table is split into RTO20e (wait) and RTO20e1 (fail) rows, mirroring the unit spec.

### 2. Proxy rule format fixes (silent/opaque failure modes)

- `objects_faults.md`: two rules wrote `"match": { "action": 20 }` as a JSON **number** — uts-proxy rejects this at session creation with HTTP 400 (`cannot unmarshal number into Go struct field MatchConfig.match.action of type string`). Now `"20"` (string).
- `connection_resume.md` (RTN15g): rule action type `"refuse"` does not exist (`refuse_connection` does). Unknown action types hit a silent passthrough in the proxy, so the second `ws_connect` was never refused and the test premise broke with no error. Corrected.
- `helpers/proxy.md`: the Match Conditions section now documents the `action` string format explicitly, plus the v0.3.0 limitation that action *names* resolve only up to `AUTH` (17) — unresolvable names (e.g. `"OBJECT_SYNC"`) silently never match, so objects actions need numeric strings until the proxy's name table is extended.
- A full regex sweep of `uts/objects`, `uts/realtime` and `uts/rest` (all match types, action types, `trigger_action` types and match fields validated against the Go proxy's structs) found no further instances.

### 3. Wall-clock timeout guidance (recurring translation trap)

All `WITH timeout` / `poll_until` / `WAIT` durations in integration specs are wall-clock time, but frameworks that virtualise time by default (e.g. kotlinx-coroutines `runTest`) fast-forward the clock while the test idles — a naively translated 15s timeout around a real network await fires instantly with a misleading error. Added:

- `docs/writing-derived-tests.md`: new section *Integration timeouts are wall-clock (beware virtual-time frameworks)* — the counterpart to the existing *No real timers in unit tests* note — with the failure mode and the shared-helper remedy.
- `docs/integration-testing.md` (Timeout Strategy) and `docs/writing-test-specs.md` (State Transitions): short pointers to it.
- `helpers/proxy.md`: new convention (#10).
- `objects_faults.md`: the two previously unbounded `object.get()` awaits now carry `WITH timeout: 15 seconds` — every await in the file is bounded, and the Common Cleanup covers multi-client tests explicitly.

### 4. `objects/PLAN.md`

Summary row updated to the corrected RTO20e scenario ("in-flight publish failing when channel enters FAILED during sync wait").

## Validation

- The corrected `objects_faults.md` is fully implemented by the ably-java derived suite: **all 5 tests pass** against the sandbox through uts-proxy, including the restructured RTO20e (92008 with cause 90000 observed end-to-end).
- A deterministic spec↔test audit reports 5/5 Test ID coverage with no missing assertions.
- Test IDs are unchanged, so existing derived-test `UTS:` tags remain valid.

## Follow-ups (not in this PR)

- `ably/uts-proxy`: extend `actionNames` with `OBJECT`/`OBJECT_SYNC`/`ANNOTATION` so name-based matching works for objects actions.
- Consider moving `realtime/integration/helpers/proxy.md` to `uts/docs/` — it is cross-module infrastructure consumed by realtime, rest and objects; best done as a dedicated rename commit with a reference sweep.
- The RTO20e proxy Test ID (`publish-fails-on-channel-failed-0`) now under-describes the scenario; kept stable for traceability, but a coordinated rename could accompany review feedback.

### DIFF
--- a/uts/README.md
+++ b/uts/README.md
@@ -28,14 +28,13 @@ uts/
 │   │   ├── connection/                # RTN — connection management
 │   │   └── presence/                  # RTP — realtime presence
 │   └── integration/                   # Realtime integration tests
-│       ├── helpers/
-│       │   └── proxy.md               # Proxy infrastructure spec
 │       ├── proxy/                     # Proxy-based fault injection tests
 │       └── *.md                       # Direct sandbox tests
 ├── docs/                              # Guides and reference
 │   ├── writing-test-specs.md          # How to write UTS specs
 │   ├── writing-derived-tests.md       # How to translate specs into SDK tests
 │   ├── integration-testing.md         # Integration testing policy
+│   ├── proxy.md                       # Proxy infrastructure spec (cross-module)
 │   └── completion-status.md           # Spec coverage matrix
 └── README.md                          # This file
 ```
@@ -101,4 +100,4 @@ See [docs/writing-test-specs.md](docs/writing-test-specs.md) for the full pseudo
 
 The programmable proxy for integration testing lives in a separate repository: [ably/uts-proxy](https://github.com/ably/uts-proxy). It sits between the SDK and the Ably sandbox, transparently forwarding traffic while allowing rule-based fault injection.
 
-See `realtime/integration/helpers/proxy.md` for the proxy infrastructure specification used by test specs in this repository.
+See `docs/proxy.md` for the proxy infrastructure specification used by test specs in this repository.
--- a/uts/README.md
+++ b/uts/README.md
@@ -100,4 +100,4 @@ See [docs/writing-test-specs.md](docs/writing-test-specs.md) for the full pseudo
 
 The programmable proxy for integration testing lives in a separate repository: [ably/uts-proxy](https://github.com/ably/uts-proxy). It sits between the SDK and the Ably sandbox, transparently forwarding traffic while allowing rule-based fault injection.
 
-See `docs/proxy.md` for the proxy infrastructure specification used by test specs in this repository.
+See `uts/docs/proxy.md` for the proxy infrastructure specification used by test specs in this repository.
--- a/uts/docs/integration-testing.md
+++ b/uts/docs/integration-testing.md
@@ -149,7 +149,7 @@ AFTER ALL TESTS:
 
 ### Proxy Setup (integration-proxy only)
 
-Proxy tests additionally set up a proxy session per test or group of tests. See `docs/proxy.md` for the proxy infrastructure API.
+Proxy tests additionally set up a proxy session per test or group of tests. See `uts/docs/proxy.md` for the proxy infrastructure API.
 
 ```pseudo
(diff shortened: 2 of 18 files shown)


paddybyers:
@sacOO7 can you also check please whether or not these changes require a change in https://github.com/ably/ably-js/pull/2219 ?

sacOO7:
> @sacOO7 can you also check please whether or not these changes require a change in [ably/ably-js#2219](https://github.com/ably/ably-js/pull/2219) ?

There was exact same recorded deviation for proxy tests. Also, integration tests were only running in `json` mode.
Fixed at -> https://github.com/ably/ably-js/pull/2256

ttypic (approved):
LGTM

PR #503, opened 2026-07-09
only update attachSerial for non-resumed attaches

SimonWoolf:
Previously the spec said to update `attachSerial` whenever you get an ATTACHED.

This doesn't make sense. attachSerial is used for `untilAttach` history, which is used to get history that is contiguous to the stream of realtime messages. If you get an attached with resumed=true because of a brief disconnection, that didn't reset your period of message continuity, so it shouldn't change the point up to which untilAttach gives you history!


paddybyers:
@ttypic does this need a UTS update?

ttypic (approved):
Changes looks good! I created PR with UTS specs changes: https://github.com/ably/specification/pull/510 (need to work on the process for the next spec improvements)

PR #505, opened 2026-07-14
[DX-1211] Add strictMode client option and granted-mode checks on subscribe and presence.get

umair-ably:
### Author's Note
Extends spec to include new strictMode behaviour as implemented in ably/ably-js#2236. Spec changes highlight whether the error is thrown or logged for previously silent subscribe and presence errors caused by channel mode mismatches. 

---

**What**: Specifies the behaviour when a client performs an operation whose effect depends on a channel mode the server did not grant: `RealtimeChannel#subscribe` without `SUBSCRIBE` (new `RTL7i`) and `RealtimePresence#get` without `PRESENCE_SUBSCRIBE` (new `RTP11f`). Adds a `strictMode` client option (`TO3r`) that selects between failing the operation (`true`) and the legacy silent result plus an `ERROR`-level log (`false`, the default), allowing SDKs to ship the fix additively before a future major version flips the default.

**Why**: Today these calls fail silently — `presence.get()` resolves with `[]` and `subscribe()` registers a listener that never fires. This is a significant DX trap for both humans and coding agents (see ably/ably-js#2236 for experiments quantifying agent failure rates: 50% unrecoverable → 0% with strict failures). The checks use the granted mode set (`RTL4m`), not the requested modes, so they also catch the case where a capability restriction caused the server to silently drop a requested mode.

**Error codes**: 90009 (`subscribe_mode_not_enabled`) and 91008 (`presence_subscribe_mode_not_enabled`) are already registered in ably-common.

**Deliberately out of scope**: presence enter/update/leave without the `presence` mode (the server NACKs those, so they are not silent); the `ErrorInfo` remediation/hint field ably-js attaches to these errors (naming under separate discussion); and the annotations mode check (`RTAN4e`), where ably-js already fails the call today — deviating from other SDKs, which log per the current spec text — so `strictMode` gating is not necessary there to preserve existing behaviour.

**Reference implementation**: ably/ably-js#2236.


specifications/api-docstrings.md
```diff
@@ -125,6 +125,7 @@ Passes additional client-specific properties to the REST [`constructor()`]{@link
...
 | addRequestIds: Bool default false ||| TO3p | When `true`, every REST request to Ably should include a random string in the `request_id` query string parameter. The random string should be a url-safe base64-encoding sequence of at least 9 bytes, obtained from a source of randomness. This request ID must remain the same if a request is retried to a fallback host. Any log messages associated with the request should include the request ID. If the request fails, the request ID must be included in the [`ErrorInfo`]{@link ErrorInfo} returned to the user. The default is `false`. |
+| strictMode: Bool default false ||| TO3r | When `true`, operations that would otherwise fail silently, such as those gated on a granted channel mode, fail with an [`ErrorInfo`]{@link ErrorInfo} describing the cause. When `false`, the same operations log a warning and return their legacy result. The default is `false`. A future major version will change the default to `true`. |
```
AndyTWF:
```suggestion
| strictMode: Bool default false ||| TO3r | When `true`, operations that would otherwise fail silently, such as those gated on a granted channel mode, will instead throw an [`ErrorInfo`]{@link ErrorInfo} describing the cause. When `false`, the same operations log a warning and return their legacy result. The default is `false`. A future major version will change the default to `true`. |
```

specifications/features.md
```diff
@@ -1950,6 +1957,9 @@ The core SDK provides an API for wrapper SDKs to supply Ably with analytics info
   - `(TO3o)` `plugins` `Dict<PluginType:Plugin>` A map between a `PluginType` and a `Plugin` object. The client library might downcast a `Plugin` to particular plugin type.
   - `(TO3p)` `addRequestIds` boolean - defaults to false. If true, `RSC7c` applies
   - `(TO3q)` `transportParams` \[String: Stringifiable\]? - defaults to null. Described in [RTC1f](#RTC1f)
+  - `(TO3r)` `strictMode` boolean - defaults to `false`. Controls the behaviour of certain operations that, for historical reasons, complete successfully even though the client is in a state in which the operation cannot have its intended effect, such as operations gated on a granted channel mode (see [`RTL7i`](#RTL7i), [`RTP11f`](#RTP11f)). It is intended that a future major version of the specification will change the default to `true`
```
AndyTWF: Perhaps a minor wording consideration - would we make StrictMode true, or would we just fix the underlying issues in a major release?

umair-ably: good spot, reworded to highlight the strictMode behaviour becomes the default

AndyTWF (approved):

PR #507, opened 2026-07-19
UTS: fix spec errors — wildcard-clientId presence setups, corrupt RSP5g fixture, RTN15h1 error code

paddybyers:
Fixes three classes of UTS spec error found while deriving tests in two SDKs (the ably-js `test/uts` audit and the ably-rust UTS suite). Per the spec-first policy in `uts/docs/writing-derived-tests.md` (#1da26476), these are fixed at source rather than adapted around in each SDK.

## 1. `clientId: "*"` in ClientOptions is invalid per RSA4c/RSA7c — presence setups were unrunnable

**RSA7c**: a `clientId` in `ClientOptions` "cannot contain only a wildcard `'*'` string value as that client ID value is reserved" — construction must reject it. Every `ClientOptions(clientId: "*")` setup therefore cannot run on a compliant SDK. Both ably-js (which rejects `"*"` at construction, per spec) and ably-rust had independently adapted these tests with local workarounds — exactly the silent divergence UTS exists to prevent.

- **enterClient-only tests** (`realtime_presence_enter.md` ×5, incl. both RTP4 bulk tests): now use an *unidentified* client (key auth, no clientId), which is the legitimate way to act on behalf of arbitrary clientIds.
- **RTP8j wildcard test**: the wildcard state RTP8j describes arises from token auth with a `clientId: "*"` token (cf. RSA7a), not from ClientOptions — setup changed accordingly. The obsolete "some SDKs may reject at construction" note is gone (all compliant SDKs must).
- **RTP15c** (`enterclient-no-side-effects-0`): with the wildcard gone this test was doubly contradictory — per **RTP8j** a wildcard/null clientId means plain `enter()` errors immediately, and per **RTP15f** an identified client's `enterClient()` with a different clientId indicates an error. No single clientId permits both operations, so the test now uses two clients: an identified member doing the normal `enter()`, and an unidentified key-auth client doing `enterClient`/`leaveClient` — which is also the real-world shape (end user + server-side process). ably-js's current adaptation (identified `"admin"` + `enterClient("other-user")`) only passes because ably-js does not implement the client-side RTP15f check; SDKs that do (e.g. ably-rust) can never pass it.
- **`realtime_presence_reentry.md`**: same RTP15f contradiction (identified `"admin"` calling `enterClient("alice")`) — now unidentified.
- The file's top-level wildcard note is rewritten to state the RSA7c/RTP8j/RTP15f rules instead of sanctioning per-SDK adaptations.

## 2. RSP5g cipher fixture is a corrupt truncation (`rest_presence.md`)

The `encrypted_data` string `HO4…4U0=` (32 bytes) is a truncation of the canonical ably-common fixture (`test-resources/crypto-data-128.json`) value `HO4…LQFt` (48 bytes) — it cuts off mid-ciphertext, and the plaintext comment (`{"secret":"data"}`) didn't match the fixture either (`{"example":{"json":"Object"}}`). Any SDK implementing this test decrypts garbage: ably-rust hit exactly this and had to source the canonical string; ably-js currently has the test skipped (`TODO: implement when cipher infrastructure is available`) and would hit it on implementation. Restored the canonical string, corrected the comment, and added a value-level assertion on the decrypted JSON so future corruption is detectable.

## 3. RTN15h1 asserts the wrong errorReason code (`connection_failures_test.md`)

The unit spec asserted `errorReason.code == 40142` (pass-through of the server's token error). Per **RSA4a2**, when the SDK has no means to renew it "should indicate an error with error code 40171 … and transition the connection to the FAILED state". This also removes the last internal inconsistency: the proxy integration spec (`connection_resume.md`, RTN15h1) already asserts 40171 with an explanatory note, and `error_reason_test.md` (RTN25) was already corrected to 40171 following the ably-js audit — this file was missed in that sweep. (ably-js's unit test currently omits the code assertion entirely; ably-rust asserts 40171.)

### DIFF
--- a/uts/realtime/unit/connection/connection_failures_test.md
+++ b/uts/realtime/unit/connection/connection_failures_test.md
@@ -77,9 +77,12 @@ AWAIT_STATE client.connection.state == ConnectionState.failed
 # Connection transitioned to FAILED
 ASSERT client.connection.state == ConnectionState.failed
 
-# Error reason is set
+# Error reason is set. Per RSA4a2, the SDK substitutes 40171 ("no means to
+# renew the token") for the server's 40142, having detected it has no way to
+# renew (no key, authCallback or authUrl). Matches the proxy integration spec
+# (connection_resume.md, RTN15h1 note) and error_reason_test.md (RTN25).
 ASSERT client.connection.errorReason IS NOT null
-ASSERT client.connection.errorReason.code == 40142
+ASSERT client.connection.errorReason.code == 40171
 ASSERT client.connection.errorReason.statusCode == 401
 CLOSE_CLIENT(client)
 ```
--- a/uts/realtime/unit/presence/realtime_presence_enter.md
+++ b/uts/realtime/unit/presence/realtime_presence_enter.md
@@ -12,12 +12,15 @@ and `leaveClient` functions. These methods send PRESENCE ProtocolMessages to the
 and handle ACK/NACK responses. Tests cover protocol message format, implicit channel
 attach, connection state conditions, and error cases.
 
-**Note on wildcard clientId:** Several tests use `clientId: "*"` (wildcard) which is
-the Ably convention for clients permitted to act on behalf of any clientId via
-`enterClient`/`updateClient`/`leaveClient`. Some SDKs may reject `"*"` at the
-`ClientOptions` construction level. In such cases, adapt these tests to use a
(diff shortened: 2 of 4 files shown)


paddybyers:
Downstream re-translation work for ably-js once this merges is tracked in ably/ably-js#2265 (RTN15h1 40171 assertion, RTP15c two-client shape, RTP8j wildcard-token setup, comment refreshes).

PR #511, opened 2026-07-23
Sync UTS specs with 6.1.0 server-side resumability

SimonWoolf:
Spec PR #488 ("Move resumability decisions from the client to the server") updated features.md and api-docstrings.md but did not update the UTS test-spec sources or address createRecoveryKey's behaviour in SUSPENDED. This brings both in line.

features.md:
- RTN16g2 is replaced by RTN16g3: createRecoveryKey() no longer returns Null in SUSPENDED. Since the client now retains its connectionKey and always attempts to resume (RTN14h), the connection remains recoverable while suspended. (RTN16g2 as written still listed SUSPENDED, which contradicts RTN8d/RTN9d retaining the key.)

UTS sources:
- connection_id_key: RTN8c/RTN9c "id/key null in SUSPENDED" -> RTN8d/RTN9d "id/key retained in SUSPENDED" (connect first, then suspend).
- connection_failures: RTN15g "no resume after connectionStateTtl" -> RTN14h "still attempts resume after the TTL", asserting every post-suspension reconnect carries resume=<connectionKey>.
- connection_recovery: RTN16g2 -> RTN16g3; the SUSPENDED case now expects a non-null recovery key.
- channel_properties: RTL15b1 "channelSerial cleared on suspended" -> RTL15b2 "retained in suspended"; detach/failed cases relabelled.
- channel_attach: RTL4j "ATTACH_RESUME set on reattach" -> "not set".
- proxy/connection_resume: Test 22 RTN15g/g2 -> RTN14h; the post-TTL reconnect now asserts the resume param is present.
- completion-status: matrix updated for RTN8d/RTN9d, RTN14h, RTN16g3, RTL15b2.


ttypic (approved):
LGTM

paddybyers:
@ttypic how are we tracking now the spec version that each library implements (especially those that have UTS-based test suites)?

AndyTWF:
> [(RTN27d)](https://sdk.ably.com/builds/ably/specification/main/features/#RTN27d) SUSPENDED - a state where the library is not connected to the server and is not actively attempting to become connected, but where a reconnect attempt is scheduled in the future after the TO3l2 suspendedRetryTimeout elapses, and where next connect attempt is a clean connection (not a resume attempt).

Feels like RTN27d could be in the scope of this PR as I believe it's a relic of the pre-RTN14h days, as theoretically an SDK could become "suspended" but the server may well still have the information it needs to resume when it comes back? That is also the basis on which ably-js implemented it.

SimonWoolf:
oh, I forgot to merge this, sorry

> Feels like RTN27d could be in the scope of this PR

Good point thx, I'll trim off the last clause. (it's one of those "spec points" that shouldn't really be a spec point because it's kindof flavour-text for an enum member rather than specifying behaviour, so I don't think it needs a break)

PR #512, opened 2026-07-29
[AIT-1245] objects & uts: RTO27 channel-state data lifecycle, objects/unit test-spec corrections, and UTS authoring conventions

sacOO7:
- Fixes https://ably.atlassian.net/browse/AIT-1245

## Summary

This started as small corrections to the objects `objects/unit` UTS specs — surfaced while translating them to ably-java and cross-checking against ably-js and ably-cocoa — and, through review, grew into three related pieces of work:

1. **A new normative spec point — `RTO27`** — specifying the objects-data lifecycle across channel state transitions, plus its UTS test.
2. **Corrections to the `objects/unit` UTS test specs** — so several tests are portable and spec-faithful.
3. **New UTS authoring conventions** — documenting identifier naming and how to record SDK deviations, so future derived-test work stays consistent.

Net effect vs `main`: **7 files**. No change to existing normative behaviour beyond specifying `RTO27` (which documents behaviour both reference SDKs already implement).

---

## 1. New normative point: `RTO27` — objects-data lifecycle on channel state transitions

**`specifications/objects-features.md`** gains **`RTO27`**: on a channel transition to `DETACHED`/`FAILED` the client clears every pooled object's data **without emitting update events** (and clears the `SyncObjectsPool`); on `SUSPENDED` it **retains** the data. This closes a previously-unspecified gap that both reference SDKs already implement (ably-js `RealtimeObject.actOnChannelState`, ably-java `DefaultRealtimeObject.handleStateChange`). It sits in the channel-lifecycle cluster (after RTO5) and cross-references RTO20e1.

**`realtime_object.md`** gains the matching **white-box UTS test** (`RTO27/channel-state-data-lifecycle-0`): it drives the internal channel-state handler directly and inspects the `ObjectsPool`, because the behaviour is not reachable black-box (access on `DETACHED` throws per RTO25b, and `SUSPENDED` is a connection-level state).

> Resolves review comment #3 (objects_pool detach-clear — *"shouldn't SDKs be consistent?"*): promoted from an SDK-internal note to a normative requirement.

## 2. `objects/unit` test-spec corrections

- **RTO18d — duplicate-listener test kept, with an OPTIONAL note** (`realtime_object.md`). Asserts the RTE4 reference ("the same listener registered twice fires twice") and adds an OPTIONAL note sanctioning SDKs that de-duplicate listeners by instance (e.g. ably-java) to assert `call_count == 1` or skip. *Resolves review comment #1 — the shared test is kept, not deleted; SDK divergence is handled per-SDK.*
- **RTO17/RTO18 — removed the "re-attach after detach" sync-event scenario** (`realtime_object.md`), with a short guard note: it is redundant with "re-sync on new ATTACHED" (same `onAttached → new-sync` path, RTO4c) and not portably expressible (an unsolicited server `DETACHED` becomes `SUSPENDED` in some SDKs).
- **Quiescence barriers before positive read-after-send** (`path_object.md`, RTPO13c5 / RTPO14): clarified these guard an *inbound* server OBJECT message (no apply-on-ack point to await), not a local publish. *Resolves review comment #4.*
- **null-means-omitted increment contract** (`internal_live_counter_api.md`): assert the equivalence directly where `null` is indistinguishable from an omitted argument.
- **camelCase internal symbols** in the RTO27 test (`objectsPool`, `processChannelState`), per the identifier-naming convention below.

> Review comment #2 (op-path update-return portability): rather than documenting a divergence in the shared spec, **ably-java was made to conform** to the `RTLC9g`/`RTLM7f` return contract — its op path now returns the `LiveObjectUpdate`. The portability notes added earlier in this PR are therefore removed and net out of the diff; the general principle is captured in the authoring conventions below.

## 3. UTS authoring conventions

Formalised two previously-implicit conventions so derived-test work is consistent:

- **Identifier naming** (`writing-test-specs.md` § *Identifier Naming*, summarised in `README.md`): spec-surface identifiers mirror the specification's own names (`PascalCase` types, `camelCase` fields/methods); behaviours the spec describes but doesn't name get a `camelCase` coinage; `snake_case` is reserved for test-harness constructs; wire/enum values keep protocol casing; and file names are unrelated to symbol casing.
- **Recording deviations** (`writing-derived-tests.md`): the SDK-deviation choice is disposition-based (env-gated skip for a bug you expect to fix; adapted assertion — preferred — for a permanent/intentional divergence); a new *"Idiomatic translation vs genuine deviations"* subsection (idiomatic naming is not a deviation; public-behaviour differences are, at any tier; internal-API-shape differences are unit-tier-only adaptations that must preserve coverage); all four deviation-section headings kept present in every `deviations.md`; `Status`/`Resolution` entry fields; and the `S-1…S-n` shape-deviation vocabulary convention.

---

## Cross-SDK validation

- Divergences verified in source: ably-js (`eventemitter.ts`, `realtimeobject.ts`), ably-cocoa (`ARTEventEmitter.m`), ably-java (`EventEmitter.java`, `DefaultRealtimeObject.kt`).
- ably-java: the affected `objects/unit` suite was regenerated and runs green; `RTO27` has a matching test; the op path now returns the `LiveObjectUpdate` (conforms to RTLC9g/RTLM7f).
- `RTO27` behaviour matches both reference SDKs.

## Non-goals

- No change to existing normative behaviour beyond specifying `RTO27`.
- Per-SDK deviations (e.g. ably-java's `EventEmitter` listener de-duplication) are recorded in each SDK's own `deviations.md`, not in the UTS.

### DIFF
--- a/specifications/objects-features.md
+++ b/specifications/objects-features.md
@@ -192,6 +192,11 @@ Objects feature enables clients to store shared data as "objects" on a channel. 
     - `(RTO5c5)` The `bufferedObjectOperations` list must be cleared
     - `(RTO5c9)` The `appliedOnAckSerials` set ([RTO7b](#RTO7b)) must be cleared. A state sync causes the channel's LiveObjects data to be replaced, so after a state sync the `appliedOnAckSerials` no longer accurately describes which operations have been applied to the channel's LiveObjects data
     - `(RTO5c8)` The [RTO17](#RTO17) sync state must transition to `SYNCED`
+- `(RTO27)` When the channel transitions to a state other than `ATTACHED`, the client library must manage the stored objects data as follows (for the effect of these transitions on an in-progress `publishAndApply`, see [RTO20e1](#RTO20e1)):
+  - `(RTO27a)` When the channel transitions to the `DETACHED` or `FAILED` state, the current state of the objects data can no longer be known, so the client library must:
+    - `(RTO27a1)` For every object in the internal `ObjectsPool`, clear its internal data, resetting it to the zero value for its type (an empty map, or a counter with value `0`), without emitting any `LiveObjectUpdate` events. The objects themselves remain in the `ObjectsPool`; only their data is cleared
+    - `(RTO27a2)` The `SyncObjectsPool` must be cleared
+  - `(RTO27b)` When the channel transitions to the `SUSPENDED` state, the client library must retain the stored objects data unchanged, since the connection may still recover and the retained data remains a valid best-effort local copy
 - `(RTO6)` Certain object operations may require creating a new object if one does not already exist in the internal `ObjectsPool` for the given `objectId`. This can be done as follows:
   - `(RTO6a)` If an object with `objectId` exists in `ObjectsPool`, do not create a new object
   - `(RTO6b)` The expected type of the object can be inferred from the provided `objectId`:
@@ -292,31 +297,6 @@ Objects feature enables clients to store shared data as "objects" on a channel. 
     - `(RTO18b2)` `SYNCED`: Indicates that the [RTO17](#RTO17) sync state has transitioned to `SYNCED`
   - `(RTO18c)` Registers the provided listener for the specified event
   - `(RTO18d)` If `on` is called more than once with the same listener and event, the listener is added multiple times to the listener registry. As such, if the same listener is registered twice and an event is emitted once, the listener will be invoked twice
-    > **EDITOR'S NOTE — SDKs diverge on this by design; it is not universally testable.** RTO18d
-    > restates the general `EventEmitter#on` contract **RTE4** (`features.md`, mirrored in
-    > `api-docstrings.md`). Registering the **same listener reference** twice for the same event is
-    > handled differently across Ably's own SDKs (verified in source):
-    > - **ably-js** (`eventemitter.ts`): `on` does a plain `listeners.push(listener)` with **no
-    >   identity check** → **fires twice** (matches RTE4/RTO18d).
-    > - **ably-cocoa** (`ARTEventEmitter.m`): `on:` mints a *fresh* `ARTEventListener` wrapping the
-    >   block on every call, and the registry's "AsSet" dedup is pointer-identity on that wrapper (the
-    >   block itself is never the key), so two calls never collide → **fires twice**.
-    > - **ably-java** (`EventEmitter.java`): `on(event, Listener)` does `filters.put(listener, …)` into
-    >   a map keyed by the listener *instance* (and the all-events `on(Listener)` is `contains`-guarded)
-    >   → **de-duplicates → fires once**. ably-java's own Javadoc flags this as a deliberate, SDK-wide
-    >   deviation from RTE4 — for a state-change listener it avoids the double-side-effect footgun the
-    >   WHATWG DOM `addEventListener` also avoids by de-duplicating.
-    >
-    > This is a **design choice, not a language limitation**: the test passes the *same reference*, and
-    > reference identity is available in every one of these languages (JS `===`, ObjC pointer, Java
-    > identity) — js/cocoa simply choose to append, java chooses to dedupe. Because the SDKs genuinely
-    > diverge, the assertion cannot hold identically everywhere, so the corresponding UTS test
-    > (`RTO18d/duplicate-listener-0`) has been **removed** from the objects UTS suite. RTE4's "fires
-    > twice" is retained as the reference behaviour; SDKs that de-duplicate document it locally
-    > (ably-java tracks it in its `deviations.md`), and any SDK may pin its own behaviour in an
-    > SDK-local (non-UTS) test.
-    > Any move to make this *uniform* would be an ecosystem-wide RTE4 change (redefining `off`'s
-    > remove-one-vs-all semantics and closure identity) — out of scope here.
   - `(RTO18e)` When an event is emitted, all registered listeners for that event must be called with no arguments
   - `(RTO18f)` The client library may return a subscription object (or the idiomatic equivalent for the language) as a result of this operation:
     - `(RTO18f1)` The subscription object includes an `off` function
--- a/specifications/objects-features.md
+++ b/specifications/objects-features.md
@@ -292,6 +292,31 @@ Objects feature enables clients to store shared data as "objects" on a channel. 
     - `(RTO18b2)` `SYNCED`: Indicates that the [RTO17](#RTO17) sync state has transitioned to `SYNCED`
   - `(RTO18c)` Registers the provided listener for the specified event
   - `(RTO18d)` If `on` is called more than once with the same listener and event, the listener is added multiple times to the listener registry. As such, if the same listener is registered twice and an event is emitted once, the listener will be invoked twice
+    > **EDITOR'S NOTE — SDKs diverge on this by design; it is not universally testable.** RTO18d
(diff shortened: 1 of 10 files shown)


specifications/objects-features.md
```diff
@@ -292,6 +292,31 @@ Objects feature enables clients to store shared data as "objects" on a channel.
     - `(RTO18b2)` `SYNCED`: Indicates that the [RTO17](#RTO17) sync state has transitioned to `SYNCED`
   - `(RTO18c)` Registers the provided listener for the specified event
   - `(RTO18d)` If `on` is called more than once with the same listener and event, the listener is added multiple times to the listener registry. As such, if the same listener is registered twice and an event is emitted once, the listener will be invoked twice
+    > **EDITOR'S NOTE — SDKs diverge on this by design; it is not universally testable.** RTO18d
```
paddybyers: I don't agree with this approach of deleting the test spec. The UTS test specs should test the assertions in the specification. If some SDKs deviate from that specification, that is something we reflect in the tests for those SDKs, not the UTS tests.

uts/objects/unit/internal_live_counter.md
```diff
@@ -11,6 +11,15 @@ Tests the `InternalLiveCounter` CRDT data structure. InternalLiveCounter holds a
 
 Tests operate directly on InternalLiveCounter by calling `applyOperation()` and `replaceData()` with constructed messages. No channel or connection infrastructure is needed.
 
+> **SDK portability note — op-path update return value.** Several tests below assert on the
```
paddybyers: We should try to design the UTS test assertions so that they are as portable as possible. Can this be done here, or are you saying that the tests in the SDKs are necessarily different?

uts/objects/unit/objects_pool.md
```diff
@@ -119,6 +119,24 @@ ASSERT updates[0].objectMessage IS null
...
 
+> **NOTE — channel DETACHED/FAILED clears objects data (SDK-internal; intentionally not a UTS test).**
+> When the channel transitions to DETACHED or FAILED, the objects data is cleared **without emitting
+> update events** (the current data is unknown in those states); a SUSPENDED channel **retains** its
+> data. This is implemented in both reference SDKs (ably-js `RealtimeObject.actOnChannelState` →
+> `objectsPool.clearObjectsData(false)` + clear SyncObjectsPool; ably-java
+> `DefaultRealtimeObject.handleStateChange`), but it is **deliberately not a shared UTS requirement**:
+> it has no normative `RTO` point, and it is not observable through the public API — access on a
+> DETACHED channel throws (RTO25b), and on re-attach the RTO5c sync **replaces** the pool regardless of
+> whether detach cleared it, so an SDK that skipped the clear would still be observably correct. It is a
```
paddybyers: Shouldn't SDKs nonetheless be implemented consistently?

uts/objects/unit/path_object.md
```diff
@@ -574,6 +574,11 @@ ASSERT result["profile"]["prefs"]["theme"] == "dark"
 mock_ws.send_to_client(build_object_message("test", [
   build_map_set("map:prefs@1000", "back_ref", { objectId: "map:profile@1000" }, "99", "remote")
 ]))
+
+# Quiescence barrier for a POSITIVE read-after-send: SDKs may apply inbound OBJECT messages
```
paddybyers: I though that with apply-on-ack you should always be able to observe the effect of an operation once the publish completes?

paddybyers (changes requested):
Pls review the comments.

(Also fyi @mschristensen @VeskeR )

specifications/objects-features.md
sacOO7: Yeah, this is actually supposed to be in draft state, claude directly raised the PR.
I am doing changes locally 👍

sacOO7: I agree the UTS test spec should be kept.

sacOO7:
This PR is in progress locally, will update and mark it as ready for review

uts/objects/unit/internal_live_counter.md
sacOO7: Yes, removed the note from here, kept in adaptations instead

uts/objects/unit/objects_pool.md
sacOO7: Updated

uts/objects/unit/path_object.md
sacOO7: Updated note

paddybyers (approved):
LGTM, thanks

specifications/objects-features.md
```diff
@@ -192,6 +192,11 @@ Objects feature enables clients to store shared data as "objects" on a channel.
     - `(RTO5c5)` The `bufferedObjectOperations` list must be cleared
     - `(RTO5c9)` The `appliedOnAckSerials` set ([RTO7b](#RTO7b)) must be cleared. A state sync causes the channel's LiveObjects data to be replaced, so after a state sync the `appliedOnAckSerials` no longer accurately describes which operations have been applied to the channel's LiveObjects data
     - `(RTO5c8)` The [RTO17](#RTO17) sync state must transition to `SYNCED`
+- `(RTO27)` When the channel transitions to a state other than `ATTACHED`, the client library must manage the stored objects data as follows (for the effect of these transitions on an in-progress `publishAndApply`, see [RTO20e1](#RTO20e1)):
```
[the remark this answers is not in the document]
sacOO7: Good catch — fixed in 0aba67fd. Rather than add a third clause, I generalized `RTO27b` into a catch-all: it now covers **any** non-`ATTACHED`, non-{`DETACHED`,`FAILED`} state (`SUSPENDED`, `INITIALIZED`, `ATTACHING`, `DETACHING`), all of which retain the stored objects data unchanged, with `SUSPENDED` kept as the named rationale. So `RTO27a` (clear) + `RTO27b` (retain) are now total over every non-`ATTACHED` transition and the "manage … as follows" enumeration is complete.

This matches what the SDKs already do — ably-js `actOnChannelState` and ably-java `handleStateChange` only act on `DETACHED`/`FAILED` and fall through (retain) for all other states — so it is a wording clarification with no behaviour change.

I also added reciprocal cross-references between `RTO4` (the `ATTACHED` transition) and `RTO27` (all other transitions), since they are complementary halves of the channel-state lifecycle and neither pointed at the other.

uts/objects/unit/internal_live_counter_api.md
```diff
@@ -237,6 +237,9 @@ The `null` row applies only where `null` is passable and distinguishable from an
 argument (e.g. a boxed `Integer` in Java). In SDKs whose signature makes null equivalent to
 "omitted" (e.g. `increment(amount?: number)` in JavaScript, where a nullish amount takes the
 default of 1), the row is not applicable — see the pseudocode conventions in `uts/README.md`.
+Such SDKs should instead assert the equivalence directly — `increment(null)` succeeds and
+increments by the default of 1 — pinning the null-means-omitted contract so a later signature
+or coalescing change surfaces as a conscious decision.
```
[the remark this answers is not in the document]
sacOO7: Agreed — reworded in 0aba67fd. The positive assertion is now framed around the omitted-argument call `increment()` (unambiguously the default-of-1 case), with `increment(null)` demoted to a scoped aside that only applies "where the language treats `null` and the absent value as equivalently nullish". The note also now states explicitly that this does not imply `null` is a valid amount, so it no longer reads as contradicting the `null`-row throw requirement.

